PR #3, opened 2017-06-13
Expose Config.LoadAndValidate func and Config.OsClient type field.

fatmcgav:
Rename existing uses of `osClient` to `OsClient`.
This is a rework of hashicorp/terraform@f3c52416a56d7772bd633f08cab48c10696bcf3d

Required for https://github.com/hashicorp/terraform/pull/13871

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -33,10 +33,10 @@ type Config struct {
 	Username         string
 	UserID           string
 
-	osClient *gophercloud.ProviderClient
+	OsClient *gophercloud.ProviderClient
 }
 
-func (c *Config) loadAndValidate() error {
+func (c *Config) LoadAndValidate() error {
 	validEndpoint := false
 	validEndpoints := []string{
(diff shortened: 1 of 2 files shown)


fatmcgav:
Looks like this one needs a rebase now #25 has been merged. Will try and get that done tomorrow... 😃

fatmcgav:
@jtopjian All rebased, should be good to merge now :)

jtopjian:
@fatmcgav I could have swore I commented on this before. Sorry about that.

Is this change to assist with the Swift remote support back in core? If so, let's hang on with this until we know what's going on with that. Otherwise, this looks good to me :)

fatmcgav:
@jtopjian Heh... Yeh, it is to support the Swift remote migration...

jtopjian:
Per https://github.com/hashicorp/terraform/pull/13871, looks like S3 is already doing this and importing the provider into core is the solution for now.

Merge at will :)

PR #25, opened 2017-06-14
Add Region to Provider

### LINKED ISSUE #17: Openstack provider should have a region argument
_This issue was originally opened by @mcanevet as hashicorp/terraform#13625. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Currently we have to either use the `OS_REGION_NAME` environment variable or pass `region` to every OpenStack resource.
It would be nive to have a `region` parameter to the `openstack` provider that defaults to `OS_REGION_NAME`.

jtopjian:
This commit adds the region argument to the OpenStack provider.
This enables a user to specify the region once at the provider-level
and have all resources within that provider be built in that region.

If a resource-level region was specified, it is still recognized
and will override the provider-level region.

Fixes #17

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 
@@ -24,6 +25,7 @@ type Config struct {
 	IdentityEndpoint string
 	Insecure         bool
 	Password         string
+	Region           string
 	Swauth           bool
 	TenantID         string
 	TenantName       string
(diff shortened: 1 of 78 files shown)


jtopjian:
I've finally had a chance to test this out in an environment with a multi-region catalog and I'm happy with how this works. It's quite safe and I was unable to accidentally destroy resources.

It's still possible there are edge cases here, so there will be a warning in the CHANGELOG to do a `plan` before applying.

@fatmcgav Let me know if you see any problems with this.

@mcanevet I found some resources that weren't setting the region in their state, notably the `openstack_compute_instance_v2` resource. If you have a moment, could you test this with the Terraform 0.10.0-dev build?

If either of you don't see any issues with this, I'll go ahead and merge it. It's a big change, so I want to get this one in first in order to avoid a lot of rebasing / merge conflicts :)

mcanevet:
@jtopjian I'll not have time to port all my custom providers to 0.10.0-dev in the nexts few days, but IIRC I add to add `region` in lifecycle's `ignore_changes` with your code because of the lack of `region` in the state. But I think it's a bug with the way it was done before, not the way you're doing it now so I can personally deal with it.
Thanks for your great work!

openstack/util.go
```diff
@@ -36,17 +35,14 @@ func CheckDeleted(d *schema.ResourceData, err error, msg string) error {
 	return fmt.Errorf("%s: %s", msg, err)
 }
 
-// GetRegion returns the region from either d.Get("region") or OS_REGION_NAME
-func GetRegion(d *schema.ResourceData) string {
+// GetRegion returns the region from either d.Get("region") or config.Region
+// which is either specified in the config or pulled from OS_REGION_NAME.
```
fatmcgav: @jtopjian Is the reference to `OS_REGION_NAME` still accurate here? 

Is `which is either specified in the resource config or pulled from provider config.` more accurate?

openstack/config.go
```diff
@@ -135,44 +137,55 @@ func (c *Config) loadAndValidate() error {
 	return nil
 }
 
+func (c *Config) determineRegion(region string) string {
+	// If a resource-level region was not specified, and a provider-level region was set,
```
fatmcgav: Is it possible to get an 'empty' region in here? 

Do we need to check for that condition?

fatmcgav:
@jtopjian Had a quick run through the code and a couple of comments... 

I'll try and get a 0.10-dev build done later to day and test in anger :)

openstack/util.go
jtopjian: Yeah, I'll clean this up.

openstack/config.go
jtopjian: It's still possible to not set a region at all. I've tested this and it works as expected. I'll update the docs to mention no region is possible for single-region catalogs.

jtopjian:
No worries - take your time.

Building the new release isn't difficult, just a few new steps. Just do a `make dev` on `hashicorp/terraform` and `make build` on this branch of. Then you'll have to do a `terraform init` before running plan/apply.

fatmcgav:
@jtopjian I'm not sure if I'm missing something, but I've tried to run the 0.10.0-dev, and I can't get it to pick up the provider... :( 

I can see the `terraform-provider-openstack` bin in `$GOPATH/bin`, but `terraform init` doesn't seem to be looking there... 
I've also tried putting it in my `.terraformrc` and that doesn't help either... 

Any pointers?

Cheers

jtopjian:
@fatmcgav are you sure you're running `~/go/bin/terraform init` and not `terraform` located in another part of path? (`which terraform`)

fatmcgav:
@jtopjian Ah, good shout... I was running from `~/golang/src/github.com/hashicorp/terraform/bin/terraform` rather than `$GOROOT/bin/terraform`.

So I'm guessing it's looking for provider plugins in the same dir as the `terraform` bin, which kinda makes sense now :)

fatmcgav:
So I've had a quick test with various combinations, and they all seem to work for me :) 

I've tried:

- Region specified on resource
- Region specified on provider
- Region specified using OS_REGION_NAME

Haven't yet had chance to try a multi-region plan, but that'll take me a bit of time to get some test configs written up... 

Happy for this to be merged though... :)

jtopjian:
👍 

Awesome.

I've also done a good amount of testing with pre-0.10.0 state files and things worked well.

Merging now.

@mcanevet @fatmcgav Thank you for your help reviewing this :)

Issue #4, opened 2017-06-13
improving control flow for lb_v2 resources


jrperritt:
The control flow of waiting for LB v2 resources to become ready to create/update could be improved. Currently, for example, it just immediately retries if it gets a `409`, which can lead to what appear to be race conditions in OpenStack. It seems like the problem should be fixed in OpenStack, but Terraform should handle this better by, for example, waiting until the LB is `ACTIVE` before continuing to try to create the LB pool after receiving a `409`.

### Terraform Version
0.9.8.

### Affected Resource(s)
- openstack_lb_*_v2

### Terraform Configuration Files

You'll need to provide `image_id` and `flavor_id`:

```hcl
variable "subnet_id" {
  type = "string"
  default = "4409be68-f1d6-44a5-a54a-bc22e790b237"
}

variable "image_id" {
  type = "string"
}

variable "flavor_id" {
  type = "string"
}

resource "openstack_lb_loadbalancer_v2" "lb" {
  vip_subnet_id = "${var.subnet_id}"
  name = "lb"
}

resource "openstack_lb_listener_v2" "lb_listener" {
  loadbalancer_id = "${openstack_lb_loadbalancer_v2.lb.id}"
  protocol = "HTTPS"
  protocol_port = 443
}

resource "openstack_lb_pool_v2" "lb_pool" {
  lb_method = "ROUND_ROBIN"
  protocol = "HTTPS"
  listener_id = "${openstack_lb_listener_v2.lb_listener.id}"
}

resource "openstack_lb_member_v2" "lb_member" {
  address = "${openstack_compute_instance_v2.server.access_ip_v4}"
  pool_id = "${openstack_lb_pool_v2.lb_pool.id}"
  protocol_port = 443
  subnet_id = "${var.subnet_id}"
}

resource "openstack_lb_monitor_v2" "lb_monitor" {
  delay = 30
  max_retries = 3
  pool_id = "${openstack_lb_pool_v2.lb_pool.id}"
  timeout = 5
  type = "PING"
}

resource "openstack_compute_instance_v2" "server" {
  name      = "server"
  image_id  = "${var.image_id}"
  flavor_id = "${var.flavor_id}"
  config_drive = false
}
```

### Expected Behavior
1 LB pool gets created

### Actual Behavior
4 LB pools get created

### Steps to Reproduce
1. `terraform apply`

jtopjian:
[...]

PR #26, opened 2017-06-15
Added retry logic to retry when OpenStack returns 409

slinn0:
OpenStack will return 409 when creating multiple listeners at the same time. This patch will allow it to retry for 10 minutes when 409 or 500 is being returned.

**Testing**
Tested with the test case in https://github.com/hashicorp/terraform/issues/15207
The listeners are now created as expected.

### DIFF
--- a/openstack/resource_openstack_lb_listener_v2.go
+++ b/openstack/resource_openstack_lb_listener_v2.go
@@ -139,7 +139,22 @@ func resourceListenerV2Create(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	log.Printf("[DEBUG] Create Options: %#v", createOpts)
-	listener, err := listeners.Create(networkingClient, createOpts).Extract()
+	var listener *listeners.Listener
+	err = resource.Retry(10*time.Minute, func() *resource.RetryError {
+		var err error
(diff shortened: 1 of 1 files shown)


fatmcgav:
Potential fix for #4 and #23.

jtopjian:
This looks to be the same solution that was implemented in https://github.com/hashicorp/terraform/pull/15225, to which @jrperritt said he found to be incomplete.

jrperritt:
@slinn0 @fatmcgav it's unclear to me why y'all would press on with this after I said it was insufficient. Nevertheless, if there's interest, I have the proper fix already prepared and waiting on #25 to merge.

fatmcgav:
@jrperritt Apologies, must have missed that comment...

slinn0:
@jrperritt It was not clear what was not sufficient. I know that the first PR was not complete because it was not tested or functioning. I had already made a patch against the external providers and tested that and verified that it worked for our test case so I made a PR that fixed our test case the patch that we tested

jtopjian:
All, the fault lies with me for delaying any work on this. I'm trying to get #25 merged in as it touches every file in the provider and merging it now will prevent any conflicts for subsequent merges.

I apologize for any frustration or confusion here.

PR #29, opened 2017-06-16
wait for lb active state before trying to create pool or monitor

jrperritt:
For #4 

This PR waits until the load balancer is in an "ACTIVE" state before trying to create a pool or monitor that depends on the load balancer. Note that if the load balancer ID is not used in the resource's definition in the Terraform configuration file, this check will not happen.

Also of note is that the underlying issue is [seemingly] in OpenStack, where it will return a `409`, yet still create the resource (possibly due to a race condition of getting hammered several times in a few milliseconds with create requests). Even the fix in this PR cannot completely prevent that race condition:
Consider: 
1. a resource is waiting on a load balancer to become "ACTIVE".
2. a different resource is simultaneously waiting on that load balancer to become "ACTIVE".
3. the load balancer goes to an "ACTIVE" state.
4. the two resources will race to OpenStack to update the load balancer.

In addition, it makes 2 other small changes:
1. creates a `checkForRetryableError` function, and
2. renames the `waitFor*Active` and `waitFor*Delete` functions to `checkFor*Active` and `checkFor*Delete`, respectively. Naming those functions that way (I think) was a legacy thing and is a misnomer: the functions don't actually do any waiting. I created a function named `waitForLoadBalancerActive` that actually does wait, so the name changes were also to prevent ambiguity with that new function.

### DIFF
--- a/openstack/resource_openstack_lb_listener_v2.go
+++ b/openstack/resource_openstack_lb_listener_v2.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/loadbalancers"
 )
 
 func resourceListenerV2() *schema.Resource {
@@ -139,10 +140,36 @@ func resourceListenerV2Create(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	log.Printf("[DEBUG] Create Options: %#v", createOpts)
-	listener, err := listeners.Create(networkingClient, createOpts).Extract()
+
+	var listener *listeners.Listener
+	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+		var err error
+		lbID := createOpts.LoadbalancerID
+		if lbID != "" {
+			lb, err := loadbalancers.Get(networkingClient, lbID).Extract()
+			if err != nil {
+				fmt.Printf("[DEBUG] Error attempting to get loadbalancer before creating listener: %s. Continuing...", err)
+			} else {
+				if lb.ProvisioningStatus != "ACTIVE" {
+					err := waitForLoadBalancerActive(d, networkingClient, lbID)
+					if err != nil {
+						fmt.Printf("[DEBUG] Error waiting for loadbalancer before creating listener: %s. Continuing...", err)
+					}
+				}
+			}
+		}
+		log.Printf("[DEBUG] Attempting to create LBaaSV2 listener")
+		listener, err = listeners.Create(networkingClient, createOpts).Extract()
+		if err != nil {
+			return checkForRetryableError(err)
+		}
+		return nil
+	})
+
 	if err != nil {
 		return fmt.Errorf("Error creating OpenStack LBaaSV2 listener: %s", err)
 	}
+
 	log.Printf("[INFO] Listener ID: %s", listener.ID)
 
 	log.Printf("[DEBUG] Waiting for Openstack LBaaSV2 listener (%s) to become available.", listener.ID)
@@ -150,7 +177,7 @@ func resourceListenerV2Create(d *schema.ResourceData, meta interface{}) error {
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"PENDING_CREATE"},
 		Target:     []string{"ACTIVE"},
-		Refresh:    waitForListenerActive(networkingClient, listener.ID),
+		Refresh:    checkForListenerActive(networkingClient, listener.ID),
 		Timeout:    2 * time.Minute,
 		Delay:      5 * time.Second,
 		MinTimeout: 3 * time.Second,
@@ -252,7 +279,7 @@ func resourceListenerV2Delete(d *schema.ResourceData, meta interface{}) error {
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"ACTIVE", "PENDING_DELETE"},
 		Target:     []string{"DELETED"},
-		Refresh:    waitForListenerDelete(networkingClient, d.Id()),
+		Refresh:    checkForListenerDelete(networkingClient, d.Id()),
 		Timeout:    2 * time.Minute,
 		Delay:      5 * time.Second,
 		MinTimeout: 3 * time.Second,
@@ -267,7 +294,7 @@ func resourceListenerV2Delete(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
-func waitForListenerActive(networkingClient *gophercloud.ServiceClient, listenerID string) resource.StateRefreshFunc {
+func checkForListenerActive(networkingClient *gophercloud.ServiceClient, listenerID string) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
 		listener, err := listeners.Get(networkingClient, listenerID).Extract()
 		if err != nil {
@@ -280,7 +307,7 @@ func waitForListenerActive(networkingClient *gophercloud.ServiceClient, listener
 	}
 }
 
-func waitForListenerDelete(networkingClient *gophercloud.ServiceClient, listenerID string) resource.StateRefreshFunc {
+func checkForListenerDelete(networkingClient *gophercloud.ServiceClient, listenerID string) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
 		log.Printf("[DEBUG] Attempting to delete OpenStack LBaaSV2 listener %s", listenerID)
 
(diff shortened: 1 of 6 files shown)


openstack/resource_openstack_lb_loadbalancer_v2.go
```diff
@@ -336,3 +337,19 @@ func waitForLoadBalancerDelete(networkingClient *gophercloud.ServiceClient, lbID
 		return lb, "ACTIVE", nil
 	}
 }
+
+func checkForRetryableError(err error) *resource.RetryError {
```
fatmcgav: @jrperritt Is it worth moving this function into [util.go](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/util.go), as can see it being useful on multiple resources...

openstack/resource_openstack_lb_listener_v2.go
```diff
@@ -139,18 +140,44 @@ func resourceListenerV2Create(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	log.Printf("[DEBUG] Create Options: %#v", createOpts)
-	listener, err := listeners.Create(networkingClient, createOpts).Extract()
+
+	var listener *listeners.Listener
+	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+		var err error
+		lbID := createOpts.LoadbalancerID
+		if lbID != "" {
```
fatmcgav: Is there any merit in extracting this to a function, as it's duplicated in both `resource_openstack_lb_listener_v2` and `resource_openstack_lb_pool_v2`.

fatmcgav:
Should also fix #23.

jtopjian:
@jrperritt Thanks for this.

My understanding is that the `StateChangeConf` is just a more flexible version of `resource.Retry`. The first time I added `resource.Retry` was for a quick fix, and then another quick fix, etc :)

Some of the `waitFor*Delete` functions embed a `Delete` call:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_networking_subnet_v2.go#L393-L423

I wonder if doing the same for `waitFor*Create` is worthwhile. This could move the create logic entirely into a single function that's automatically retried rather than having two separate areas that are doing that.

@jrperritt I can look into this over the weekend unless you want to or unless you see a reason not to investigate that?

> Naming those functions that way (I think) was a legacy thing and is a misnomer: the functions don't actually do any waiting. 

I can't remember how the naming started, either. I'm totally fine renaming them to whatever, but I'll want to follow up with a PR to rename all `waitFor` functions to the same pattern shortly afterwards for consistency.

openstack/resource_openstack_lb_listener_v2.go
jrperritt: Originally I had it like that, but we'd want to distinguish between resource types, so the function would have to accept a `string` for that, and it just looked awkward.

openstack/resource_openstack_lb_loadbalancer_v2.go
jrperritt: That's reasonable. I'd left it here because different resources may have different codes for retrying an operation. I'll move it to `util.go` for now.

jrperritt:
> I can look into this over the weekend unless you want to or unless you see a reason not to investigate that?

I think it'd be worthwhile to look into, though I doubt I'd have time to do it. I'd expect those changes to be in a follow-on PR similar to renaming the `wait` functions.

jtopjian:
This looks good. I'll look at alternative implementations + consistency renaming.

```
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./openstack -v -run=TestAccLBV2Listener -timeout 120m
=== RUN   TestAccLBV2Listener_basic
--- PASS: TestAccLBV2Listener_basic (108.24s)
PASS
ok      github.com/terraform-providers/terraform-provider-openstack/openstack   108.277s

==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./openstack -v -run=TestAccLBV2LoadBalancer -timeout 120m
=== RUN   TestAccLBV2LoadBalancer_basic
--- PASS: TestAccLBV2LoadBalancer_basic (83.27s)
=== RUN   TestAccLBV2LoadBalancer_secGroup
--- PASS: TestAccLBV2LoadBalancer_secGroup (105.74s)
=== RUN   TestAccLBV2LoadBalancer_timeout
--- PASS: TestAccLBV2LoadBalancer_timeout (69.51s)
PASS
ok      github.com/terraform-providers/terraform-provider-openstack/openstack   258.557s

==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./openstack -v -run=TestAccLBV2Member -timeout 120m
=== RUN   TestAccLBV2Member_basic
--- PASS: TestAccLBV2Member_basic (152.77s)
=== RUN   TestAccLBV2Member_timeout
--- PASS: TestAccLBV2Member_timeout (137.32s)
PASS
ok      github.com/terraform-providers/terraform-provider-openstack/openstack   290.124s

==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./openstack -v -run=TestAccLBV2Monitor -timeout 120m
=== RUN   TestAccLBV2Monitor_basic
--- PASS: TestAccLBV2Monitor_basic (156.48s)
=== RUN   TestAccLBV2Monitor_timeout
--- PASS: TestAccLBV2Monitor_timeout (137.70s)
PASS
ok      github.com/terraform-providers/terraform-provider-openstack/openstack   294.222s

==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./openstack -v -run=TestAccLBV2Pool -timeout 120m
=== RUN   TestAccLBV2Pool_basic
--- PASS: TestAccLBV2Pool_basic (132.50s)
=== RUN   TestAccLBV2Pool_timeout
--- PASS: TestAccLBV2Pool_timeout (115.30s)
PASS
ok      github.com/terraform-providers/terraform-provider-openstack/openstack   247.837s
```

jrperritt:
@fatmcgav @jtopjian thanks for reviewing

openstack/util.go
```diff
@@ -95,3 +96,19 @@ func FormatHeaders(headers http.Header, seperator string) string {
...
 	return strings.Join(redactedHeaders, seperator)
 }
+
+func checkForRetryableError(err error) *resource.RetryError {
+	switch errCode := err.(type) {
+	case gophercloud.ErrDefault500:
+		return resource.RetryableError(err)
+	case gophercloud.ErrUnexpectedResponseCode:
+		switch errCode.Actual {
+		case 409, 503:
```
slinn0: Does it make any sense to retry on 500 as well?

slinn0:
Thanks @jrperritt this is working for our test case better than the original.  It failed once because of a 500 from neutron:
```
2017/06/17 10:38:25 [DEBUG] plugin: terraform-provider-openstack: User-Agent: Terraform 0.9.8 (go1.8.3) gophercloud/2.0.0
2017/06/17 10:38:25 [DEBUG] plugin: terraform-provider-openstack: X-Auth-Token: ***
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: 2017/06/17 10:38:28 [DEBUG] Openstack Response Code: 500
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: 2017/06/17 10:38:28 [DEBUG] Openstack Response Headers:
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: Content-Length: 101
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: Content-Type: application/json; charset=UTF-8
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: Date: Sat, 17 Jun 2017 17:38:27 GMT
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: X-Openstack-Request-Id: req-XXX
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: 2017/06/17 10:38:28 [DEBUG] OpenStack Response Body: {
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:   "NeutronError": {
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:     "detail": "",
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:     "message": "An error happened in the driver",
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:     "type": "DriverError"
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:   }
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: }
```
In our environment it looks like we need to retry when getting a 500 as a response code in addition to the 409s and 503s.

jrperritt:
Hmm, that's a tough one: a 500 could be considered as a candidate for retrying actions, but because a 500 is intentionally unspecific, there's no way of knowing if retrying is reasonable or not.

PR #31, opened 2017-06-17
Make checkForRetryableError retry on 500 in addition to 409 and 503

slinn0:
**Summary of Changes**

When creating multiple listeners and load balancers in our environment we see 500s on occasion.
Here is an example:
```
2017/06/17 10:38:25 [DEBUG] plugin: terraform-provider-openstack: User-Agent: Terraform 0.9.8 (go1.8.3) gophercloud/2.0.0
2017/06/17 10:38:25 [DEBUG] plugin: terraform-provider-openstack: X-Auth-Token: ***
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: 2017/06/17 10:38:28 [DEBUG] Openstack Response Code: 500
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: 2017/06/17 10:38:28 [DEBUG] Openstack Response Headers:
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: Content-Length: 101
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: Content-Type: application/json; charset=UTF-8
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: Date: Sat, 17 Jun 2017 17:38:27 GMT
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: X-Openstack-Request-Id: req-***
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: 2017/06/17 10:38:28 [DEBUG] OpenStack Response Body: {
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:   "NeutronError": {
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:     "detail": "",
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:     "message": "An error happened in the driver",
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:     "type": "DriverError"
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack:   }
2017/06/17 10:38:28 [DEBUG] plugin: terraform-provider-openstack: }

```

### DIFF
--- a/openstack/util.go
+++ b/openstack/util.go
@@ -103,7 +103,7 @@ func checkForRetryableError(err error) *resource.RetryError {
 		return resource.RetryableError(err)
 	case gophercloud.ErrUnexpectedResponseCode:
 		switch errCode.Actual {
-		case 409, 503:
+		case 409, 500, 503:
 			return resource.RetryableError(err)
 		default:
 			return resource.NonRetryableError(err)


jtopjian:
@slinn0 Thanks for submitting this.

Just to confirm, and I apologize for asking (I just don't want to assume): adding 500 as a retryable error fixed the issue for you?

If so, I'm curious about what loadbalancer driver you're using.

The reason I ask is to learn if it's a vendor/proprietary LB driver. These kinds of drivers usually respond with errors that don't follow the norm of other drivers. @fatmcgav and I have been kicking around the idea of adding some type of way to flag vendor-specific behavior that deviates from otherwise normal behavior.

I'm OK with merging this, but I have one request: if it is a vendor driver, could you add the 500 in a separate `case` statement with a comment about it being for such-and-such driver? This is for when we do any changes/refactoring, 500 isn't assumed to be an acceptable retry error for _everything_.

slinn0:
@jtopjian,

Thanks for reviewing and the comments.  This PR was based on that I see 500s come from OpenStack/Neutron pretty frequently.  If you look at Neutron API code here https://github.com/openstack/neutron/blob/stable/ocata/neutron/api/api_common.py#L405
you can see that they do not want to return 500s with the comment `# returning 500 is definitely confusing.` .  But you can see at https://github.com/openstack/neutron/blob/stable/ocata/neutron/api/api_common.py#L409-L416 that when a unhandled exception happens that `Everything else is 500` .   I guess the question is if we want to be liberal in what we accept from OpenStack and retry on 500s and put the infrastructure in the expected state when the unexpected happens.

We are using the A10 load balancer.  I do not know if OpenStack will let you see this information from the Neutron API currently. This PR does *not* fix my test case.  I do have a fix for my test case. I will explain what the issue is when I make the PR for the fix. I think we should close this PR.

jtopjian:
@slinn0 Thanks for the info.

IMO, if retrying on a 500 will eventually get things working, then we can definitely add it as retryable code. But from the sound of it, unknown errors have been lumped into 500, so I'm not sure if it makes sense to retry on it.

> This PR does not fix my test case. I do have a fix for my test case. I will explain what the issue is when I make the PR for the fix. I think we should close this PR.

Sounds good. Thanks!

PR #32, opened 2017-06-18
LBaaS v2 Cleanup

jtopjian:
This commit does a large cleanup of the LBaaS v2 resources:

* Renames all "waitFor" / "checkFor" functions to a standard naming
scheme that matches other providers.
* Removes delete actions out of the "waitFor" functions and into
a resource.Retry loop.
* Tries to ensure that required parent resources are in an ACTIVE
state before proceeding with any action.
* Ensures all resources have proper timeout settings.
* General cleanup of logging and other formatting.
* It combines the "timeout" tests into the standard "basic" tests.
* Adds an extra "member" to the member test to try to better
trigger update conflicts.


jtopjian:
A few non-commit notes on this PR:

1. It's large, but that was unavoidable due to how the LBaaS v2 resources are all interconnected. Some of the formatting changes and stuff could have been put off into another PR or a separate commit, but they're relatively minor when compared to the "waitFor" / "checkFor" refactoring.

2. I surveyed other providers and saw that `WaitForState` and `resource.Retry` are used almost the same amount of times. However, `WaitForState` is exclusively used to poll a resource's state and never to make modifications. So I yanked all of the "Delete" calls out of the existing ones.

3. The standard function name that a `WaitForState` will call is `something-something-RefreshFunc`. So all functions have renamed to that naming scheme.

The same will eventually be done for all other resources in the this provider.

jtopjian:
@jrperritt @slinn0 if either of you are able to test this sometime in the next week, I'd very much appreciate it. 

AFAICT, the changes made in #29 have carried over, they've just been placed into dedicated blocks of code rather than combined in a single `resource.Retry` loop. This all passes in my testing environment, but if you have environments that can easily trigger pending update situations, it'd be great to test this out there.

The one piece of logic I removed from #29 was code like this:

```go
lbID := createOpts.LoadbalancerID
if lbID != "" {
```

in situations when the `LoadbalancerID` is a required field in the schema. There's no way a `""` value would have made it to that point, so it's safe to remove.

jrperritt:
I'll try to test today

jtopjian:
@jrperritt Thanks! There's no rush. Anytime within the next week would be great. This code is all contained in these resources, so it's fine to have this PR sit here for a few days.

jrperritt:
I didn't review the code, but I was able to checkout this branch and run 3 times successfully (make and destroy).

jtopjian:
> I didn't review the code, but I was able to checkout this branch and run 3 times successfully (make and destroy).

I'll take it! :)

Thank you for testing this out.

slinn0:
I ran 31 tests today with this code and there were 2 failures. I am pretty new to terraform and golang but I am pretty confident that I understand the issue and can make a PR to fix it.

jtopjian:
> I ran 31 tests today with this code and there were 2 failures. I am pretty new to terraform and golang but I am pretty confident that I understand the issue and can make a PR to fix it.

Sounds good. I'm going to go ahead and merge this, but please definitely submit patches for any errors you find.

Thanks everyone for looking at this.

PR #33, opened 2017-06-18
BlockStorage v2: Remove instance_id from volume attach

jtopjian:
Removing a deprecated argument


jtopjian:
There's no need for a CHANGELOG update on this one. `instance_id` was never used in the first place -- it was just mistakenly added into the schema.

Issue #28, opened 2017-06-16
Error associating multiple floating IPs to VM with more than one NIC


dihedron:
## Description
When creating a VM with two NICs and trying to associate two floating IP addresses, one to each NIC, the operation fails and the VM is left with one NIC bound to a floating IP, the other unbound.

We can consistently reproduce the problem applying the following steps:
1. create a VM with two NICs
2. get two floating IPs from a pool
3. assign each floating IP to a different NIC

### Terraform Version
`Terraform v0.9.8` against an OpenStack "Newton" (Canonical distribution)

### Affected Resource
- openstack_compute_floatingip_associate_v2

### Terraform Configuration Files
```hcl
########################################################
# 1. Create the VM instance ***with 2 NICs***
#
resource "openstack_compute_instance_v2" "terraform-test" {
  name            = "floating_ip_test"
  image_name      = "${var.image_name}"
  flavor_name     = "${var.instance_flavor}"  
  #  
  # NOTE: this VM has two NICs!!
  #
  network  {
    uuid = "${openstack_networking_network_v2.network_1.id}"
  }
  network  {
	uuid = "${openstack_networking_network_v2.network_1.id}"
  }
}

########################################################
# 2. Retrieve two floating IPs from pool
#
resource "openstack_compute_floatingip_v2" "fips" {
  count			  = "2"
  pool = "${var.floatingIpPool}"
}

########################################################
# 3. Attempt to assign a different floating IP to each of the two NICs
#
resource "openstack_compute_floatingip_associate_v2" "fip_ass0" { 
  floating_ip = "${openstack_compute_floatingip_v2.fips.0.address}"
  instance_id = "${openstack_compute_instance_v2.terraform-test.id}"
  fixed_ip    = "${ openstack_compute_instance_v2.terraform-test.network.0.fixed_ip_v4 }"
}

# NOTE: uncomment this to reproduce the failure
/*
resource "openstack_compute_floatingip_associate_v2" "fip_ass1" {
  floating_ip = "${openstack_compute_floatingip_v2.fips.1.address}"
  instance_id = "${openstack_compute_instance_v2.terraform-test.id}"
  fixed_ip    = "${ openstack_compute_instance_v2.terraform-test.network.1.fixed_ip_v4 }"
}
*/

########################################################
# TESTBED SETUP:
# 1. create a network
# 2. create a subnet with it
# 3. create a router with external gateway
# 4. attach the network to the router

########################################################
# 1. create a network
#
[...]

PR #36, opened 2017-06-18
fix(builtin/provider): OpenStack

gappaa:
In some providers, OpenStack can't found the right network uuid. In this case, the content of old values of networks are stored and throw an error at the creation. For example, if we create an instance with a private and a public network. OpenStack doesn't provide the right uuid of private network. In this case, Terraform keep the old value, the public uuid and try to assign a private address to a public network, what is not possible to happen.

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -937,13 +937,16 @@ func getInstanceNetworks(computeClient *gophercloud.ServiceClient, d *schema.Res
 
 		networkID := ""
 		networkName := ""
+		networkFound := false
 		if tenantNetworkExt {
 			for _, network := range networkList {
 				if network.Name == rawMap["name"] {
 					tenantnet = network
+					networkFound = true
 				}
 				if network.ID == rawMap["uuid"] {
 					tenantnet = network
+					networkFound = true
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


jtopjian:
(for reference, previous discussion here: https://github.com/hashicorp/terraform/pull/15329)

@gappaa Thanks! I'll run this though the acceptance tests later today and let you know if there are any problems.

gappaa:
@jtopjian All right !

jtopjian:
This looks good to me. I'm going to go ahead and merge it.

Thanks, @gappaa 

Note for the future: I think a possible solution for the future would be to give an error when there is more than one network with the same name.

fatmcgav:
@jtopjian You mention 2 networks with the same... That's exactly the scenario in #28, which is currently failing...

jtopjian:
@fatmcgav Ah, cool. This PR will end up choosing the first network, which should provide some type of consistency. 

In a way, some of this is the user's responsibility to not get into this situation. But it should be Terraform's job to trigger an error if a situation does occur. This PR is a good holdover until I get time to look clean up instance network configuration in general. Now that Floating IPs have been removed, it'll be a lot easier to do that.

jtopjian:
er, actually, hold on. This PR isn't preventing multiple networks with the same name.

It's preventing adding a default network when no networks were found.

gappaa:
@jtopjian You are welcome :wink:

jtopjian:
@fatmcgav OK, nevermind again. The problem in #28 isn't that there are multiple networks with the same name. It's an instance that is consuming multiple ports on the same network. So completely ignore my statement about user responsibility. :)

PR #49, opened 2017-07-14
Adding support for 'TERMINATED_HTTPS' for lb_listeners backed by A10 networks LBs

rcreasey:
The existing `HTTPS` protocol on a `openstack_lb_listener_v2` resource does not function properly if it's passing to an A10 backed loadbalancer.  This LB needs to have the protocol defined in neutron as "TERMINATED_HTTPS".

### DIFF
--- a/openstack/resource_openstack_lb_listener_v2_test.go
+++ b/openstack/resource_openstack_lb_listener_v2_test.go
@@ -89,6 +89,26 @@ func testAccCheckLBV2ListenerExists(n string, listener *listeners.Listener) reso
 	}
 }
 
+func TestAccLBV2Listener_terminated_https_basic(t *testing.T) {
+	var listener listeners.Listener
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLBV2ListenerDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: TestAccLBV2ListenerConfig_terminated_https_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLBV2ListenerExists("openstack_lb_listener_v2.listener_1", &listener),
+					resource.TestCheckResourceAttr(
+						"openstack_lb_listener_v2.listener_1", "protocol", "TERMINATED_HTTPS"),
+				),
+			},
+		},
+	})
+}
+
 const TestAccLBV2ListenerConfig_basic = `
 resource "openstack_networking_network_v2" "network_1" {
   name = "network_1"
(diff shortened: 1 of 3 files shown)


openstack/resource_openstack_lb_listener_v2_test.go
```diff
@@ -89,6 +89,26 @@ func testAccCheckLBV2ListenerExists(n string, listener *listeners.Listener) reso
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLBV2ListenerExists("openstack_lb_listener_v2.listener_1", &listener),
+					resource.TestCheckResourceAttr(
+						"openstack_lb_listener_v2.listener_1", "protocol", "TERMINATED_HTTPS"),
+				),
+			},
+		},
+	})
+}
+
```
jtopjian: I'm getting an error when running this test:

```
2017/07/15 11:43:26 [DEBUG] OpenStack Response Body: {   
  "NeutronError": {                                      
    "detail": "",                                        
    "message": "Default TLS container was not specified",
    "type": "TLSDefaultContainerNotSpecified"            
  }                                                      
}                                                        
```

It could very well be that the default haproxy load balancer driver requires extra configuration.

Can you confirm that this works on your A10? If so, I am OK with adding support for `TERMINATED_HTTPS` (as it is a valid Protocol in the Neutron API), but let's remove this test for the time being until full TLS support is worked out.

rcreasey:
Hey Joe,

Yeah, I thought I did include the update to the [markdown documentation](https://github.com/terraform-providers/terraform-provider-openstack/pull/49/files#diff-9a5ea366dfea1de34ac98fc6419df4c9).  Is this not accurate?

As for the failed test, our ops guys are telling me that this configuration is indeed needed to work with our A10's and it does work properly (in fact, it doesn't with just `HTTPS` protocol set).  Does this require any corresponding changes to the underlying [gophercloud libs](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/listeners/requests.go#L11-L15)?  I'm not sure if that's still a dependency anymore.

Either way, I'll update the PR by removing the test; I just didn't want to send one that didn't have a test and be that kind of contributor. ;)

jtopjian:
> Yeah, I thought I did include the update to the markdown documentation. Is this not accurate?

Duh - my mistake. 

> As for the failed test, our ops guys are telling me that this configuration is indeed needed to work with our A10's and it does work properly (in fact, it doesn't with just HTTPS protocol set).

:+1: 

That works for me. The failed test I saw is just a matter of configuring the default haproxy lb driver to work correctly with this. The A10 you're using probably requires no further configuration, and that's why it's passing for you.

> Does this require any corresponding changes to the underlying gophercloud libs? I'm not sure if that's still a dependency anymore.

It probably should be added just for completeness, but it looks like `TERMINATED_HTTPS` is being cast to `listeners.Protocol` correctly:

```
2017/07/15 16:49:43 [DEBUG] OpenStack Request Body: {         
  "listener": {                                               
    "admin_state_up": true,                                   
    "connection_limit": 100,                                  
    "loadbalancer_id": "ce1062bb-9291-4587-a279-51a36d96ebc0",
    "name": "listener_1_updated",                             
    "protocol": "TERMINATED_HTTPS",                           
    "protocol_port": 8080                                     
  }                                                           
}                                                             
```

This looks good to me. Thank you for submitting this!

PR #50, opened 2017-07-15
New Resource: openstack_identity_project_v3

jtopjian:
This commit adds initial support for OpenStack Identity resources.
It also adds the openstack_identity_project_v3 resource.

PR #52, opened 2017-07-15
New Resource: openstack_identity_user_v3

jtopjian:
Will rebase after #50

### DIFF
--- a/openstack/resource_openstack_identity_user_v3.go
+++ b/openstack/resource_openstack_identity_user_v3.go
@@ -1,0 +1,311 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/users"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+var userOptions = map[users.Option]string{
+	users.IgnoreChangePasswordUponFirstUse: "ignore_change_password_upon_first_use",
+	users.IgnorePasswordExpiry:             "ignore_password_expiry",
+	users.IgnoreLockoutFailureAttempts:     "ignore_lockout_failure_attempts",
+	users.MultiFactorAuthEnabled:           "multi_factor_auth_enabled",
+}
+
+func resourceIdentityUserV3() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceIdentityUserV3Create,
+		Read:   resourceIdentityUserV3Read,
+		Update: resourceIdentityUserV3Update,
+		Delete: resourceIdentityUserV3Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"default_project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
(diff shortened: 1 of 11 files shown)


fatmcgav:
@jtopjian Once you've rebased this one I'm happy to approve :)

jtopjian:
Hold on one moment. I see a few small changes that need made.

jtopjian:
Alright, that should do it. I pushed a separate commit so you can see the changes made.

I also need to follow-up with a small PR for the Project Import test (to restrict to Admin only). I'll go ahead and merge that myself, though.

openstack/resource_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,311 @@
...
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			// The following are all specific options that must
+			// be bundled into user.Options
```
jtopjian: @fatmcgav For some background on this, I spent a good two hours trying to decide if these options should look like this:

```hcl
resource "openstack_identity_user_v3" "foo" {
  ...
  options {
    ignore_change_password_upon_first_use = true
  }
}
```

or this:

```hcl
resource "openstack_identity_user_v3" "foo" {
  ...
  ignore_change_password_upon_first_use = true
}
```

housing them under `options` makes sense from the API point of view, but since the user won't be able to specify multiple `options` blocks, I opted to just make these all top-level parameters.

fatmcgav: Yeh, I think the model looks quite good..

In-fact it gave me an idea on how to handle the 'vendor-specific' modifications we've discussed a few times...

fatmcgav:
@jtopjian Latest commit looks good to me. 

Did you want to squash and I'll approve?

openstack/resource_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,311 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"multi_factor_auth_enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"multi_factor_auth_rule": &schema.Schema{
```
jtopjian: @fatmcgav This was a fun one to figure out.

You can read more about MFA rules [here](https://docs.openstack.org/releasenotes/keystone/ocata.html). They're defined as "lists of lists" which is hard to implement in HCL. I'm not sure if it's a bug or feature, but HCL will flatten a list of lists as just a list.

Therefore, to get the JSON to render properly, you have to specify a `multi_factor_auth_rule` for each MFA rule. In a way, that's cleaner to the user, though:

```hcl
multi_factor_auth_rule {
  rule = ["password"]
}

multi_factor_auth_rule {
  rule = ["password", "totp"]
}
```

which will render as:

```python
user["options"]["multi_factor_auth_rules"] = [["password"], ["password", "totp"]]
```

jtopjian:
Sure. I will squash the last two commits.

Just left you some FYI notes about this one.

openstack/resource_openstack_identity_user_v3.go
fatmcgav: Hah, always fun when HCL doesn't do quite what you want it to :)

jtopjian:
Thank you!

PR #55, opened 2017-07-23
Update attribute description style

r24y:
Modify the attribute descriptions for `protocol` and `protocol_port` to match the description style used on the terraform.io website.


jtopjian:
Looks good to me. Thanks!

PR #59, opened 2017-08-02
Lbaa s v2 timeouts

xgerman:
In most cases it is is sufficient to just wait for the LB to become
ACTIVE again. In areas where the lbID wasn't readilt available
retrive it through the pool object and wait on LB.

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -212,3 +212,19 @@ func resourceLBV2PoolRefreshFunc(networkingClient *gophercloud.ServiceClient, po
 		return pool, "ACTIVE", nil
 	}
 }
+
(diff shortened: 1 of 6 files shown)


jtopjian:
@xgerman Thank you for this - this is a really nice update!

All changes make sense to me. I have two small nits:

1. Can you fix the commit history?
2. The `waitForLBV2viaPool` function: can this either be renamed to `waitForLBV2viaPoolActive` or add the ability specify states other than `ACTIVE`? 

Also pinging @jrperritt since I don't want this to cause any regression with issues he saw before. But given you two are both at Rackspace, maybe he's already aware of this?

jrperritt:
@jtopjian Thanks. Yes, we are seeing/ have seen several timing issues that _seem_ to be either in Terraform (here) or OpenStack. They all seem to be in Neutron, where resources can have several dependencies.

jtopjian:
@xgerman This looks good to me. Thank you for submitting it!

PR #65, opened 2017-08-10
r/lb_loadbalancer_v2: Remove 'provider' field

radeksimko:
Deprecation was added in https://github.com/hashicorp/terraform/pull/12239 (in February 2017).

That said the user wouldn't ever see the deprecation message because the whole field (along with some other reserved fields) is removed before CRUD even gets chance to process it.

https://github.com/hashicorp/terraform/blob/5bcc1bae5925f44208a83279b6d4d250da01597b/config/loader_hcl.go#L673-L676

This is first step towards bumping `hashicorp/terraform` to `v0.10.0` which has validation for these kinds of things (see https://github.com/hashicorp/terraform/pull/15562).

### DIFF
--- a/openstack/resource_openstack_lb_loadbalancer_v2.go
+++ b/openstack/resource_openstack_lb_loadbalancer_v2.go
@@ -81,14 +81,6 @@ func resourceLoadBalancerV2() *schema.Resource {
 				ForceNew: true,
 			},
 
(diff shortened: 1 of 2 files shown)


jtopjian:
@radeksimko I totally forgot about this -- thank you!

PR #67, opened 2017-08-10
vendor: github.com/hashicorp/terraform/...@v0.10.0

radeksimko:
```
govendor fetch github.com/hashicorp/terraform/...@v0.10.0
```


jtopjian:
@radeksimko Thanks for this. I think this should be quite safe to merge? Acceptance tests are passing for me.

Feel free to merge if so :)

PR #81, opened 2017-08-22
Added reread of security group before deleting default rules

edevreede:
Creating a security group with the attribute delete_default_rules set did not remove the default rules.

This addition makes the provider reread the security group before looping over the rules inside.


fatmcgav:
@edevreede Changes look sensible enough to me. 

Any objections to me merging @jtopjian?

jtopjian:
Looks good to me.

As an aside: we have an acceptance test to verify `delete_default_rules` is working, and the test has always been passing. The test still passes with this PR, which is what we want, but I'm curious to know if the Mirantis cloud is doing something unique or if it's simply a timing issue that is causing this patch to be required.

fatmcgav:
Yeh, I wondered the same, as haven't hit the same issue. 

However the change is simple enough so should mitigate any timing issues :)

PR #83, opened 2017-08-22
Add openstack_compute_flavor_v2 resource

mcanevet:
- [x] Create
- [X] Read
- [X] Delete
- [X] Importer
- [x] Documentation

### DIFF
--- a/openstack/resource_openstack_compute_flavor_v2.go
+++ b/openstack/resource_openstack_compute_flavor_v2.go
@@ -1,0 +1,143 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceComputeFlavorV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeFlavorV2Create,
+		Read:   resourceComputeFlavorV2Read,
+		Delete: resourceComputeFlavorV2Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ram": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: true,
+				ForceNew: true,
+			},
+			"vcpus": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: true,
+				ForceNew: true,
+			},
+			"disk": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: true,
+				ForceNew: true,
+			},
+			"swap": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+			"rx_tx_factor": &schema.Schema{
+				Type:     schema.TypeFloat,
+				Optional: true,
+				ForceNew: true,
+				Default:  1,
+			},
+			"is_public": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+			},
+			"ephemeral": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceComputeFlavorV2Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
+	disk := d.Get("disk").(int)
+	swap := d.Get("swap").(int)
+	isPublic := d.Get("is_public").(bool)
+	ephemeral := d.Get("ephemeral").(int)
+	createOpts := flavors.CreateOpts{
+		Name:       d.Get("name").(string),
+		RAM:        d.Get("ram").(int),
+		VCPUs:      d.Get("vcpus").(int),
+		Disk:       &disk,
+		Swap:       &swap,
+		RxTxFactor: d.Get("rx_tx_factor").(float64),
+		IsPublic:   &isPublic,
+		Ephemeral:  &ephemeral,
+	}
+
+	log.Printf("[DEBUG] Create Options: %#v", createOpts)
+	fl, err := flavors.Create(computeClient, &createOpts).Extract()
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack flavor: %s", err)
+	}
+
+	d.SetId(fl.ID)
+
+	return resourceComputeFlavorV2Read(d, meta)
+}
+
+func resourceComputeFlavorV2Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
+	fl, err := flavors.Get(computeClient, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "flavor")
+	}
+
+	d.Set("name", fl.Name)
+	d.Set("ram", fl.RAM)
+	d.Set("vcpus", fl.VCPUs)
+	d.Set("disk", fl.Disk)
+	d.Set("swap", fl.Swap)
+	d.Set("rx_tx_factor", fl.RxTxFactor)
+	d.Set("is_public", fl.IsPublic)
+	// d.Set("ephemeral", fl.Ephemeral) TODO: Implement this in gophercloud
+	d.Set("region", GetRegion(d, config))
+
+	return nil
+}
+
+func resourceComputeFlavorV2Delete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
+	err = flavors.Delete(computeClient, d.Id()).ExtractErr()
+	if err != nil {
+		return fmt.Errorf("Error deleting OpenStack flavor: %s", err)
+	}
+	d.SetId("")
+	return nil
+}
--- a/openstack/resource_openstack_compute_flavor_v2_test.go
+++ b/openstack/resource_openstack_compute_flavor_v2_test.go
@@ -1,0 +1,98 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
+)
+
+func TestAccComputeV2Flavor_basic(t *testing.T) {
+	var flavor flavors.Flavor
+	var flavorName = fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckAdminOnly(t)
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeV2FlavorDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeV2Flavor_basic(flavorName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeV2FlavorExists("openstack_compute_flavor_v2.flavor_1", &flavor),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckComputeV2FlavorDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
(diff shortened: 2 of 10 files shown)


jtopjian:
@mcanevet This is a really good start. 

I was able to get this working with the following [fixes](https://github.com/jtopjian/terraform-provider-openstack/commit/a4d44c31e2b2a76a853800cf69dffed5dee1b2d6). Some of them are a little tricky, so they deserve some explanation:

* The custom `FlavorCreateOpts` is only needed if you plan on having a Terraform parameter called `value_specs`. `value_specs` is where we have been allowing arbitrary key/value parameters into resources. This is usually required for vendor-specific OpenStack extensions. For example, let's say you had an OpenStack cloud managed by Big Vendor who requires `tenant_id` to be added to the request in order for the flavor to be created. You would then do:

```hcl
resource "openstack_compute_flavor_v2" "my_flavor" {
  name = "my_flavor"
  ...
  value_specs {
    tenant_id = "abcd1234"
  }
}
```

Since the WIP flavor resource you created doesn't have `value_specs`, that was one cause of an error. 

Unless you need `value_specs`, it is safe to not add it until someone specifically requests it. The reason for this is because for each custom type we make in `types.go`, we need to monitor possible upstream changes to Gophercloud.

* `createOpts` needed to be passed as `&createOpts`. I suspect this is a bug in Gophercloud as no other `Create` function requires a pointer (it used to under the old Rackspace version). I will open an issue at Gophercloud and ask about this.

* The only place the `id` field needs to be set or read is with `d.SetId()` and `d.Id()`.

* Passing `disk` and `swap` as pointers is a two-step process rather than directly casting to `*int`.

If you'd like, I can submit a PR of my fixes to your branch, or you can copy them, or implement them yourself with any modifications you need.

One thing I recommend is also including parameters for `is_public` and `ephemeral`. It looks like `ephemeral` is not listed in the Gophercloud `Flavor` result type. I suspect this is also a bug.

With regard to the `_test.go` file I added, you can run this test by doing the following:

```shell
source ~/openrc
cd ~/go/src/github.com/terraform-providers/terraform-provider-openstack              
TF_LOG=DEBUG make testacc TEST=./openstack TESTARGS="-run=TestAccComputeV2Flavor_basic" 2>&1 | tee ~/openstack.log
```

I have a local bash function to quickly run tests:

```shell
ttest() {                                                                                   
  if [[ -n $1 ]]; then                                                                      
    pushd ~/go/src/github.com/terraform-providers/terraform-provider-openstack              
    TF_LOG=DEBUG make testacc TEST=./openstack TESTARGS="-run=$1" 2>&1 | tee ~/openstack.log
    popd                                                                                    
  fi                                                                                        
}                                                                                           

ttest TestAccComputeV2Flavor_basic
```

Let me know if you have any questions. :)

mcanevet:
@jtopjian thanks a lot for your explanations.

Creation works with your fix. I'll see if I've time to look at gophercloud to add the update and delete support.

I first added parameters for `is_public` and `ephemeral`, but I noticed that they are not supported in the pinned version of gophercloud https://github.com/terraform-providers/terraform-provider-openstack/blob/v0.2.0/vendor/github.com/gophercloud/gophercloud/openstack/compute/v2/flavors/results.go#L34-L49 (they are in upstream) so I removed them from my PR for now.

fatmcgav:
@mcanevet Feel free to update the vendor dep if necessary... 

`govendor update [package]` should do the job... 
Have a look at the commit history for example commit messages...

mcanevet:
@fatmcgav I'm trying to update gophercloud with this command:
```
$ govendor fetch github.com/gophercloud/gophercloud@81a3d5b517a645123c0659518ac468bb14588619
```
But somehow it looks like it fails somewhere because I don't have the full diff I can see here : https://github.com/gophercloud/gophercloud/compare/c5dfe1bc7b150a98b1578bb32f7c138ff9568f1d...81a3d5b517a645123c0659518ac468bb14588619
I only have changes in `vendor/github.com/gophercloud/gophercloud/STYLEGUIDE.md` and `vendor/vendor.json`.
Any idea?

fatmcgav:
@mcanevet I suspect `govendor` is just fetching that specific commit.

In the past, I've run `govendor update github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas/routerinsertion/...`, which updates the entire package.

So maybe try that on `github.com/gophercloud/gophercloud/openstack/compute/v2/flavors/...` and see what it does...

mcanevet:
nothing happens...

fatmcgav:
@mcanevet Hmm, did you revert your changes to `vendor.json`?

As I've just run `govendor update github.com/gophercloud/gophercloud/openstack/compute/v2/flavors/...` against master, and have the following changes reported:
```
# git status --short
 M vendor/github.com/gophercloud/gophercloud/openstack/compute/v2/flavors/requests.go
 M vendor/github.com/gophercloud/gophercloud/openstack/compute/v2/flavors/results.go
 M vendor/vendor.json
```

mcanevet:
@fatmcgav after checkouting master it works on by branch...
I don't understand what happened, but whatever.
Thanks

fatmcgav:
@mcanevet Are you intending to update Gophercloud to support Update/Delete?

Just wondering whether it's worth merging this one as is and then add support for Update & Delete operations once Gophercloud has been updated... Rather than leave it sitting open for a while :) 

@jtopjian Thoughts?

fatmcgav:
Also, I guess documentation needs to go on the 'TODO' list :)

mcanevet:
@fatmcgav I'm not sure I'll have time to add support of update/delete to gophercloud.
I added documentation and squashed my commits.

website/docs/r/compute_flavor_v2.html.markdown
```diff
@@ -0,0 +1,69 @@
...
+layout: "openstack"
+page_title: "OpenStack: openstack_compute_flavor_v2"
+sidebar_current: "docs-openstack-resource-compute-flavor-v2"
+description: |-
+  Manages a V2 flavor resource within OpenStack.
+---
+
+# openstack\_compute\_flavor_v2
+
+Manages a V2 flavor resource within OpenStack.
```
fatmcgav: Is it worth adding a note that currently only 'Create' operation is supported?

@jtopjian Is there any prior-art for this?

openstack/resource_openstack_compute_flavor_v2.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	disk := d.Get("disk").(int)
+	swap := d.Get("swap").(int)
+	is_public := d.Get("is_public").(bool)
+	ephemeral := d.Get("ephemeral").(int)
+	createOpts := flavors.CreateOpts{
+		Name:       d.Get("name").(string),
+		RAM:        d.Get("ram").(int),
+		VCPUs:      d.Get("vcpus").(int),
```
fatmcgav: For consistency, should `RAM` and `VCPUs` be assigned to variables as well? 
Granted they'll just be passed straight through rather than pointered... 

Appreciate that gophercloud expects them in different formats currently, but probably should standardise :)

openstack/resource_openstack_compute_flavor_v2.go
```diff
@@ -0,0 +1,137 @@
...
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ram": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: true,
+				ForceNew: false,
```
fatmcgav: As Gophercloud doesn't currently support 'Update', is it worth setting these to `ForceNew: true`? 

@jtopjian Thoughts?

mcanevet: @fatmcgav as Gophercloud does not support delete either, `ForceNew: true` will not help...

openstack/resource_openstack_compute_flavor_v2.go
mcanevet: @jtopjian Thoughts?

jtopjian:
This will have to wait until Update and Delete are implemented in Gophercloud. It doesn't make sense to have a Terraform resource that only creates.

I can look at adding Update and Delete in the next week.

openstack/resource_openstack_compute_flavor_v2.go
fatmcgav: Ah, true... 😃

openstack/resource_openstack_compute_flavor_v2.go
jtopjian: I think it's fine to leave this as-is. The pattern throughout the rest of the OpenStack resources is more or less:

```go
passAsInt := d.Get("foo").(string)
needsWorkDoneFirst := resourceComputeV2SomethingBar(d.Get("bar"))
createOpts := something.CreateOpts{
  Foo: &passAsInt,
  Bar: needsWorkDoneFirst,
  Baz: d.Get("baz").(string),
}
```

jtopjian: Put another way, if you wanted to assign ram and vcpus as a variable as well, that opens the door to other contributors doing something like:

```go
var1 := d.Get("var1").(string)
var2 := d.Get("var2").(string)
var3 := d.Get("var3").(string)
createOpts := something.createOpts{
  Var1: var1,
  Var2: var2,
  Var3: var3,
}
```

While that's not bad and it technically works, I'd rather stick with something closer to how the rest of the resources look.

fatmcgav: Fine by me... I guess ideally Gophercloud should be updated to use the same pattern throughout [flavors.CreateOpts](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/flavors/requests.go#L97-L124)

jtopjian: The reason `disk` takes a pointer is because it's actually a tertiary type of setting: 0, non-zero, nil. :)

fatmcgav: Ah, Ok... Lots of little nuances then :)

fatmcgav:
@mcanevet I've opened a PR to add support for Delete operation to Gophercloud - https://github.com/gophercloud/gophercloud/pull/501

jtopjian:
> Edit: So according to the Compute API Docs, you can't update flavors anyways, so Create and Delete is sufficient, combined with ForceNew: true on any changes :)

Yes, this is correct. I realized this this evening. While Horizon gives the ability to update flavors, it's really doing a delete/create in the background.

If you guys get around to adding Delete, I'd like to do one final review before it gets merged.

mcanevet:
@fatmcgav @jtopjian , I tried to update `gophercloud` using `govendor update github.com/gophercloud/gophercloud/...` but I get compilation errors unrelated with my changes:
```
 make build
==> Checking that code complies with gofmt requirements...
go install
# github.com/terraform-providers/terraform-provider-openstack/vendor/github.com/gophercloud/gophercloud/openstack/identity/v3/users
vendor/github.com/gophercloud/gophercloud/openstack/identity/v3/users/requests.go:216: cannot use "github.com/terraform-providers/terraform-provider-openstack/vendor/github.com/gophercloud/gophercloud/pagination".LinkedPageBase literal (type "github.com/terraform-providers/terraform-provider-openstack/vendor/github.com/gophercloud/gophercloud/pagination".LinkedPageBase) as type "github.com/gophercloud/gophercloud/pagination".LinkedPageBase in field value
GNUmakefile:7: recipe for target 'build' failed
make: *** [build] Error 2
```
If I update only the flavor using `govendor update github.com/gophercloud/gophercloud/openstack/compute/v2/flavors/...` it still compiles (I still have to implement `Delete`though.

Any idea? Should I update only `openstack/compute/v2/flavors`?

jtopjian:
> Any idea? Should I update only openstack/compute/v2/flavors?

Yes.

I've been meaning to do a full Gophercloud vendor update and see if I can resolve any errors. But for now, definitely just update the flavors package.

fatmcgav:
@mcanevet Yeh, I think the general practise is to only update the package required... 

So just updating `flavors` is fine.

mcanevet:
I just implemented `Delete`

openstack/resource_openstack_compute_flavor_v2.go
```diff
@@ -0,0 +1,143 @@
...
+func resourceComputeFlavorV2Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
+	disk := d.Get("disk").(int)
+	swap := d.Get("swap").(int)
+	is_public := d.Get("is_public").(bool)
```
jtopjian: nit: `isPublic` instead of `is_public`.

openstack/resource_openstack_compute_flavor_v2_test.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+		}
+
+		found, err := flavors.Get(computeClient, rs.Primary.ID).Extract()
+		if err != nil {
+			return err
+		}
+
+		if found.Name != rs.Primary.ID {
```
jtopjian: `found.ID`

The test was failing and this took me a 30 minutes to spot :smile:

website/docs/r/compute_flavor_v2.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+* `ram` - See Argument Reference above.
+* `vcpus` - See Argument Reference above.
+* `disk` - See Argument Reference above.
+* `swap` - See Argument Reference above.
+* `rx_tx_factor` - See Argument Reference above.
+* `is_public` - See Argument Reference above.
+
+## Import
+
+Flavors can be imported using the `name`, e.g.
```
jtopjian: `s/name/ID`

website/docs/r/compute_flavor_v2.html.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+```
+$ terraform import openstack_compute_flavor_v2.my-flavor test-flavor
```
jtopjian: `s/test-flavor/4142e64b-1b35-44a0-9b1e-5affc7af1106/`

jtopjian:
LGTM

@mcanevet Thank you for submitting this and your patience!

@fatmcgav Thanks for getting the flavor Delete functionality in place!

PR #90, opened 2017-08-25
LBaaS v2: Don't send connection_limit unless it has been set

### LINKED ISSUE #87: LBaaSv2: connection_limit always set for listener, resulting in errors for some provider
Hi there,

### Terraform Version
`Terraform v0.10.2`
`provider-openstack_v0.2.0_x4`

### Affected Resource(s)
- `openstack_lb_listener_v2`

```hcl
resource "openstack_lb_listener_v2" "listener" {
  protocol        = "HTTP"
  protocol_port   = 80
  loadbalancer_id = "${openstack_lb_loadbalancer_v2.loadbalancer.id}"
  admin_state_up  = "true"
}
```

The OpenStack provider I am currently using with Terraform doesn't allow to set the `connection_limit` attribute for the LBaaSv2 listener. For some reasons, even if I do not specify this property for my resource, Terraform ist setting this field in the API request:

```
2017/08/18 09:56:48 [DEBUG] OpenStack Request Body: {
  "listener": {
    "admin_state_up": true,
    "connection_limit": 0,
    "loadbalancer_id": "f904502b-a215-40db-9cec-060661d81f95",
    "protocol": "HTTP",
    "protocol_port": 80
  }
}
```

This results in an error:

```
2017/08/18 09:56:49 [DEBUG] OpenStack Response Body: {
  "NeutronError": {
    "detail": "",
    "message": "Rules on create_listener by *** disallowed by policy",
    "type": "PolicyNotAuthorized"
  }
}
```

Sending the request manually without the `connection_limit` attribute works fine. Since this is an optional parameter, I don't see why this is set even if the user doesn't specify it. I also don't know if 0 is a good default value, because the internal OpenStack default is -1, meaning no limit.

jtopjian:
This commit changes the behavior of the `openstack_lb_listener_v2`
resource so that the `connection_limit` argument is not sent
unless it has been explicitly defined in the configuration.

Fixes #87

### DIFF
--- a/openstack/resource_openstack_lb_listener_v2.go
+++ b/openstack/resource_openstack_lb_listener_v2.go
@@ -86,6 +86,7 @@ func resourceListenerV2() *schema.Resource {
 			"connection_limit": &schema.Schema{
 				Type:     schema.TypeInt,
 				Optional: true,
+				Computed: true,
 			},
 
 			"default_tls_container_ref": &schema.Schema{
@@ -122,7 +123,6 @@ func resourceListenerV2Create(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	adminStateUp := d.Get("admin_state_up").(bool)
-	connLimit := d.Get("connection_limit").(int)
 	var sniContainerRefs []string
 	if raw, ok := d.GetOk("sni_container_refs"); ok {
 		for _, v := range raw.([]interface{}) {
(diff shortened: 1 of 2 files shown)


jtopjian:
cc'ing some people I know are working with LBaaS at the moment so they are aware: @jrperritt @xgerman @jamiehannaford

PR #93, opened 2017-08-28
LBaaS v2: Remove ID fields

jtopjian:
Terraform core recently made the `id` field reserved. This commit
removes the redundant id field from the LBaaS v2 resources.


jtopjian:
Not quite ready to merge this.

jtopjian:
@catsby The only hesitation I had on this was confirmation that removing these fields won't have an impact. I'm 99% sure they won't. Can you confirm?

jtopjian:
I just ran a few before/after tests and all seems fine. Going to merge.

PR #105, opened 2017-09-08
make it possible to search for non active networks

Crapworks:
Hi all,

my openstack provider for some reasons does not allow to filter the external network by status. So the default filter "&status=ACTIVE" will not show the external network when searching for it. 

I created a new attribute called **only_active**, which can be set to false to remove the ACTIVE filter. If this attribute is not used, it defaults to the ACTIVE filter, so it shouldn't breaking existing terraform scripts.

I am **NOT** a Go programmer, so if this can be done in a more elegant way, just mention it and I will change it accordingly.

Cheers,
Christian

### DIFF
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -31,6 +31,10 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"status": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 			"matching_subnet_cidr": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
@@ -64,7 +68,10 @@ func dataSourceNetworkingNetworkV2Read(d *schema.ResourceData, meta interface{})
 		ID:       d.Get("network_id").(string),
 		Name:     d.Get("name").(string),
 		TenantID: d.Get("tenant_id").(string),
-		Status:   "ACTIVE",
+	}
+
+	if v, ok := d.GetOk("status"); ok {
+		listOpts.Status = v.(string)
 	}
 
 	pages, err := networks.List(networkingClient, listOpts).AllPages()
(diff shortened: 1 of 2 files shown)


jtopjian:
@Crapworks I think a better way to go is to add a new parameter called `status`. If no status is specified, then the query won't include a status.

This will be semi-breaking for backwards compatibility as some users will have to add `status = "ACTIVE"` for their configurations, but that's not terribly bad for two reasons:

1. It's possible some users won't have to add it at all as the only networks returned will coincidentally be active ones anyway.
2. It allows filtering of networks with other statuses if that use-case happens.

Additionally, hard-coding a status of active was a poor assumption on my part, so it's best to fix it.

Thoughts? @fatmcgav ?

fatmcgav:
@jtopjian Yeh, having a `status` field is probably the most sensible. 

Is there any merit in supporting multiple status'? Or is 'ACTIVE' the only status of any value?

jtopjian:
Off the top of my head, I can't think of a reason to filter by any other status, but that's not to say there isn't a use-case.

One option is to add a `status` parameter with a default value of "ACTIVE". That's probably the most appropriate solution, but we still need some way to turn it off for @Crapworks. 

Maybe a `status_not_supported` boolean field? That sounds like an appropriate vendor-specific type of field.

Crapworks:
@jtopjian I actually like the idea of just adding a status field the most, although it make break backwards compatibility. It's in my oppinion the most intuitive since most openstack resources have parameters that translate directly to API parameters.

I would implement it that way if you both agree @jtopjian @fatmcgav

jtopjian:
@Crapworks go for it :)

I did some quick tests and I don't think there's going to be a big backwards compatibility issue. The only occurrences would be if a user has an active and non-active network with the same search credentials which would now be returned (two results). I think there's a low chance of that.

Crapworks:
@jtopjian That should do the trick and looks less messy than the first solution :)

jtopjian:
This looks good to me. Tests pass both with and without declaring `status`. 

@fatmcgav any comments/thoughts on this?

fatmcgav:
@jtopjian / @Crapworks All looks good to me :)

Issue #24, opened 2017-06-13
neutron port-update updates unnecessary attributes (allowed_address_pairs)


hashibot:
_This issue was originally opened by @michalmedvecky as hashicorp/terraform#15220. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

v0.9.6

### Affected Resource(s)
- openstack_networking_port_v2

I have a network port in OpenStack which needs a new security group to be attached on it. When I run `terraform plan`, there is a correct output (removing one security group, adding another one):

```
openstack_networking_port_v2.otplb_port_backend.0: Modifying... (ID: f8c309c7-0c81-404d-a2df-41d0e5f5dce7)
  security_group_ids.1691341947: "721c89fb-b8a6-46dd-b031-73038443ded8" => "721c89fb-b8a6-46dd-b031-73038443ded8"
  security_group_ids.1894619809: "ae9af23e-a5d7-45be-a561-21220c74493b" => "ae9af23e-a5d7-45be-a561-21220c74493b"
  security_group_ids.4281792489: "a9f112b3-76f5-4d6b-b573-929df4169a63" => ""
  security_group_ids.691798261:  "" => "2e801289-8912-446a-9bbc-030e498e31fa"
```

but `terraform apply` fails, wanting to unnecessarily update `allowed_address_pairs`, which I have no permissions to update:

```
* openstack_networking_port_v2.otplb_port_backend.0: Error updating OpenStack Neutron Network: Expected HTTP response code [200 201] when accessing [PUT http://1.2.3.4:9696/v2.0/ports/f8c309c7-0c81-404d-a2df-41d0e5f5dce7], but got 403 instead
[...]

PR #114, opened 2017-09-14
Port fixes

jtopjian:
@fatmcgav This fixes a bug I recently found in Gophercloud which explains some oddities we've seen with ports and allowed address pairs. Let me know if these changes check out with your cloud.

For #24

### DIFF
--- a/openstack/resource_openstack_networking_port_v2.go
+++ b/openstack/resource_openstack_networking_port_v2.go
@@ -74,9 +74,13 @@ func resourceNetworkingPortV2() *schema.Resource {
 				Type:     schema.TypeSet,
 				Optional: true,
 				ForceNew: false,
-				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
+			},
+			"no_security_groups": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
 			},
 			"device_id": &schema.Schema{
 				Type:     schema.TypeString,
@@ -105,7 +109,6 @@ func resourceNetworkingPortV2() *schema.Resource {
 				Type:     schema.TypeSet,
 				Optional: true,
 				ForceNew: false,
-				Computed: true,
 				Set:      allowedAddressPairsHash,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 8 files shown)


openstack/resource_openstack_networking_port_v2.go
```diff
@@ -142,6 +151,16 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
 		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
 	}
 
+	var securityGroups []string
+	v := d.Get("security_group_ids")
+	securityGroups = resourcePortSecurityGroupsV2(v.(*schema.Set))
+	noSecurityGroups := d.Get("no_security_groups").(bool)
+
+	// Check and make sure an invalid security group configuration wasn't given.
+	if noSecurityGroups && len(securityGroups) > 0 {
```
fatmcgav: Is there a reason that 'ConflictsWith' won't work for this?
e.g. https://github.com/terraform-providers/terraform-provider-openstack/blob/b217557828d76c0ce6a7c207f7525be0cffc98fc/openstack/resource_openstack_fw_firewall_v1.go#L72

jtopjian: Do you know if the following is possible?

```hcl
resource "openstack_fw_firewall" "foo" {
  no_routers = false
  associated_routers = ...
}
```

Would Terraform flag this since both are defined in the schema even though one is false?

fatmcgav: Looks like TF does flag it:
>2 error(s) occurred:
>
>* openstack_fw_firewall_v1.firewall_1a: "associated_routers": conflicts with no_routers (false)
>* openstack_fw_firewall_v1.firewall_1a: "no_routers": conflicts with associated_routers ([]interface {}{"${openstack_networking_router_v2.tf_nc_test_rtr_1a.id}"})

jtopjian: > Whether that's a beneficial behaviour or not, I'm not sure yet ;)

Yeah, that's where I'm at, too. The reasons I didn't use `ConflictsWith` were:

1. Assumption that the above error happened (thanks for looking into that - I was just being lazy)
2. Thinking of the workflow where a user would initially have `no_security_groups = true`, then later add `security_group_ids` and set `no_security_groups` to `false` rather than remove it.

`ConflictsWith` continues to be the thing we want to use the most but never works out :)

fatmcgav: It almost feels like `ConflictsWith` should handle booleans' and only flag if not set to the `Default` value...

jtopjian:
@fatmcgav I rebased with master, retested, and ran the entire CI test suite. All is passing. Unless you have any objections, let's plan to merge this tomorrow and we'll request a new release. We'll most likely have to hold off on 1.0.0 since this is a breaking change, but this is a fix we need to get in place.

lhuard1A:
Hello,

I tested this PR and I can confirm that it fixes the two issues I had:
* Security group assignment to a LBaaS vPort (#126) and;
* FIP association to a LBaaS vPort.

Thks!

jtopjian:
@lhuard1A Thank you very much for testing 😄

PR #117, opened 2017-09-15
Compute v2: Fix for Instance NICs

jtopjian:
A previous patch which fixed detection of IPv4/IPv6 NICs created
a regression bug where instances with multiple NICs on the same
network would not be correctly detected.

This fixes the bug by correlating IPv4 and IPv6 addresses by MAC
address.


jtopjian:
@fatmcgav Once I got CI working again, the multi-nic test was showing failure. Upon investigation I saw this was a pretty serious bug. I've tested all cases and believe this is a good fix, so I'm going to merge, given the severity.

Issue #108, opened 2017-09-10
Import support for openstack_networking_router resources


haxorof:
### Terraform Version

0.10.4

### OpenStack Provider Version

0.2.1

### Affected Resource(s)

Please list the resources as a list, for example:

- openstack_networking_router_interface_v2
- openstack_networking_router_v2
- openstack_networking_router_route_v2

### Terraform Configuration Files

```hcl
resource "openstack_networking_router_v2" "router_1" {
  name           = "router_1"
  admin_state_up = "true"
}

resource "openstack_networking_network_v2" "network_1" {
  name           = "network_1"
  admin_state_up = "true"
}

resource "openstack_networking_subnet_v2" "subnet_1" {
  network_id = "${openstack_networking_network_v2.network_1.id}"
  cidr       = "192.168.199.0/24"
  ip_version = 4
}

resource "openstack_networking_router_interface_v2" "int_1" {
  router_id = "${openstack_networking_router_v2.router_1.id}"
  subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
}

resource "openstack_networking_router_route_v2" "router_route_1" {
  depends_on       = ["openstack_networking_router_interface_v2.int_1"]
  router_id        = "${openstack_networking_router_v2.router_1.id}"
  destination_cidr = "10.0.1.0/24"
  next_hop         = "192.168.199.254"
}
```

OpenStack provider information is only set in terraform, no environment OS prefixed variables are set.

### Expected Behavior

Router related resources imported.

### Actual Behavior

Import of resources fails because it is not supported:

```
resource openstack_networking_router_interface_v2 doesn't support import
resource openstack_networking_router_v2 doesn't support import
resource openstack_networking_router_route_v2 doesn't support import
```

### Steps to Reproduce

1. Execute `terraform apply`
2. Rename `terraform.tfstate` to `terraform.tfstate_org`
3. Import resources according to documentation but router related resources does not support import

### Possible fix

I have implemented a possible fix which I tried out but have not yet implemented any acceptance tests for these new imports:
- https://github.com/haxorof/terraform-provider-openstack/commit/1e52cba66920450850527bc64b305e8f72012384

Then it will be possible to import using the following syntax:

```
terraform import openstack_networking_router_v2.<Name> <RouterID>
terraform import openstack_networking_router_route_v2.<Name> <RouterInterfaceID>/<RouterID>/<SubnetID>/<PortID>
[...]

PR #119, opened 2017-09-19
Networking v2: Import router interface subnet

jtopjian:
This commit amends the router interface import feature
by setting the subnet_id when the subnet_id can be
confidently determined.

For #108


jtopjian:
Indeed it has :)

Thanks!

PR #131, opened 2017-10-23
add property error_on_empty

iaroslavscript:


jtopjian:
@iaroslavscript Thank you for submitting this.

Are there other Terraform data sources (outside of the OpenStack provider) which allow this type of functionality? It's my understanding (though I might be wrong) that a data source should trigger an error if no results were found.

iaroslavscript:
Hi @jtopjian 
Yeah, I have to agree with you that data source is not the best place for querying.
Could you advice me a workaround with resource.openstack_images_image_v2?

jtopjian:
@iaroslavscript Let me think/ask around on that. I understand what you're trying to do in #132, so I'll see if there's a better way to go about it.

iaroslavscript:
Thank you @jtopjian 
I'll be waiting for any info from you.

jtopjian:
@iaroslavscript I spoke to some of the other Terraform devs today. It was agreed that having a data source _not_ error would break precedence. However, the use-case of having a data source not error is valid. The general Terraform-wide solution is not known at this time, though.

For now, one solution is to create an `openstack_images_image_ids_v2` data source similar to the [`aws_ami_ids`](https://www.terraform.io/docs/providers/aws/d/ami_ids.html) data source. It would return a list of all matching results IDs (a list of strings), including an empty list when there are no results, so you should be able to test the length of the list.

I can create this resource when I have some time. If it's something you'd like to work on, you are most welcomed to. It would look like a mix of the `openstack_images_image_v2` resource and `aws_ami_ids` resource.

iaroslavscript:
Hi @jtopjian thank you so much, that's good decision.
Let's closed this pull-request and I'll open WIP one.

PR #136, opened 2017-10-26
Networking v2: ConflictsWith for Subnet Gateway

jtopjian:
This commit adds a schema ConflictsWith for the subnet gateway
options allowing the error to be caught earlier in the
plan/apply phase instead of when resources began to be created.

Saw the opportunity to do this while making the subnet data source.

### DIFF
--- a/openstack/resource_openstack_networking_subnet_v2.go
+++ b/openstack/resource_openstack_networking_subnet_v2.go
@@ -73,15 +73,17 @@ func resourceNetworkingSubnetV2() *schema.Resource {
 				},
 			},
 			"gateway_ip": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: false,
-				Computed: true,
+				Type:          schema.TypeString,
+				ConflictsWith: []string{"no_gateway"},
+				Optional:      true,
+				ForceNew:      false,
+				Computed:      true,
 			},
 			"no_gateway": &schema.Schema{
-				Type:     schema.TypeBool,
-				Optional: true,
-				ForceNew: false,
+				Type:          schema.TypeBool,
+				ConflictsWith: []string{"gateway_ip"},
+				Optional:      true,
+				ForceNew:      false,
 			},
 			"ip_version": &schema.Schema{
 				Type:     schema.TypeInt,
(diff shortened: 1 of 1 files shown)


openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -73,15 +73,17 @@ func resourceNetworkingSubnetV2() *schema.Resource {
...
+				Computed:      true,
 			},
 			"no_gateway": &schema.Schema{
-				Type:     schema.TypeBool,
-				Optional: true,
-				ForceNew: false,
+				Type:          schema.TypeBool,
+				ConflictsWith: []string{"gateway_ip"},
+				Optional:      true,
+				ForceNew:      false,
```
catsby: I believe instead we want to replace `ForceNew: false` on `no_gateway` with a `Default: false` ( I realize the `ForceNew` is not actually part of this change)

openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -149,18 +151,12 @@ func resourceNetworkingSubnetV2Create(d *schema.ResourceData, meta interface{})
...
-	}
-
-	if gatewayIP != "" {
+	if v, ok := d.GetOk("gateway_ip"); ok {
+		gatewayIP := v.(string)
 		createOpts.GatewayIP = &gatewayIP
 	}
 
-	if noGateway {
+	if _, ok := d.GetOk("no_gateway"); ok {
```
catsby: Need to use `d.GetOkExists` here instead; that will tell you if a boolean value has been supplied. `d.GetOk` tells you if a _non-default value_ has been specified, in which case if a user specified `false` this block would not execute. This prevents users from toggling it on and off.

jtopjian: I saw that `GetOkExists` was added to core a few releases ago, but I'm not sure how we can use it -- it's just been a low priority for me to ask.

Should we do a vendor update of core for this provider?

catsby: Ah, well, if you provide a default of `false` then `GetOkExists` may not be necessary at all and you won't need to update the dependency. 

Instead, define a default in the schema of `no_gateway` and then use regular `d.Get()` here, then set the property according to the value so that users can toggle `true` and `false`.

jtopjian: That makes sense -- thank you :)

I'll get this fixed up soon.

As an aside, there are handful of other areas where OpenStack does a `true/false/nil` type of setting, so `GetOkExists` would definitely be useful. Again, low priority and we can discuss some other time on Slack.

jtopjian:
@catsby Updated based on review.

By adding a default value to `no_gateway`, I had to make changes to the `Read` function accordingly.

I also realized I didn't modify the `Update` function to account for the original `ConflictsWith` addition.

Let me know how this looks and if you have any other comments/pointers :)

PR #139, opened 2017-10-26
data source openstack_images_image_ids_v2

iaroslavscript:


jtopjian:
@iaroslavscript Just a quick note: the reason why Travis failed is because `go fmt` needs to be run on the file(s). The makefile supplied with this repo has a shortcut of `make fmt` to help with this.

iaroslavscript:
Sorry I have been on a long vacation and I guess I have a question but  I'll be able to return to it in a few days.

jtopjian:
@iaroslavscript No problem at all - take your time and enjoy your vacation :)

tivvit:
I think this is useful, is it possible to finish this?

iaroslavscript:
Hi, @jtopjian,
this PR was abandoned for a few years (thanks for not closing it). Now it's almost finished and there are some questions, I'd like to discuss:

1. `dataSourceImagesImageV2Read` and `dataSourceImagesImageIDsV2Read` are very similar. So in order to avoid code duplication I moved code which filters by properties to a separate function `imagesFilterByProperties`. I preserved original logic of filtering at `dataSourceImagesImageV2Read` which is applied if there are at least 2 images and I added a remark of it at documentation. [https://github.com/terraform-providers/terraform-provider-openstack/blob/66a77dde7bf5a6343ad36e2020dd6b1d78c3b8f7/openstack/data_source_openstack_images_image_v2.go#L228](https://github.com/terraform-providers/terraform-provider-openstack/blob/66a77dde7bf5a6343ad36e2020dd6b1d78c3b8f7/openstack/data_source_openstack_images_image_v2.go#L228) But if there is no need in the limitation any more I could remove conditional check.  

2. `dataSourceImagesImageIDsV2Read` has a support of a new sort API and search by regex which could be applied to `dataSourceImagesImageV2Read` as well but I prefer to do it in separate PR.

3. I have no clear understanding how Acc test works because it gives me a different results depending on the steps. Could you please see `TestAccOpenStackImagesV2ImageIDsDataSource_basic`.
 - Currently it passes OK but if I when I try to test only one step (commenting other out) if fails. For example `testAccOpenStackImagesV2ImageIDsDataSource_properties` fails when tested alone and passes when tested in group.
 - Search by `size_min`/`size_max` always return 2 images regarding of value. Why?
 - When I run tests with **TF_LOG=DEBUG** I see messages **[DEBUG] Retrieved 0 images in openstack_images_image_ids_v2: []** But the test passes OK and I don't understand it.

openstack/data_source_openstack_images_image_ids_v2.go
```diff
@@ -0,0 +1,215 @@
...
+
+// dataSourceImagesImageIdsV2Read performs the image lookup.
+func dataSourceImagesImageIdsV2Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	imageClient, err := config.ImageV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack image client: %s", err)
+	}
+
+	_, nameOk := d.GetOk("name")
```
ozerovandrei, line 136: It's better to use [ConflictsWith](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/schema/schema.go#L183) in one of the conflicting arguments instead of doing this checks manually.

openstack/data_source_openstack_images_image_ids_v2.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	_, sortOk := d.GetOk("sort")
+	sortKeyValue := d.Get("sort_key")
+	sortDirectionValue := d.Get("sort_direction")
+
+	if sortOk {
+		// Attribute "sort" cannot be used simultaneously with
+		// "sort_key". If both are present, only "sort" will be used.
+		sortKeyValue = ""
```
ozerovandrei, line 150: It's better to use ConflictsWith same as above.
We also have some examples of usage: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_identity_role_assignment_v3.go#L30

openstack/data_source_openstack_images_image_ids_v2.go
```diff
@@ -0,0 +1,215 @@
...
+		allImages = imagesFilterByRegex(allImages, d.Get("name_regex").(string))
+		log.Printf("[DEBUG] Image list filtered by regex: %s", d.Get("name_regex"))
+	}
+
+	log.Printf("[DEBUG] Got %d images after filtering in openstack_images_image_ids_v2: %+v", len(allImages), allImages)
+
+	imageIDs := make([]string, 0)
+	for _, image := range allImages {
+
+		imageIDs = append(imageIDs, image.ID)
```
ozerovandrei, line 208: It's better to preallocate a slice since we already know its length:
```
imageIDs := make([]string, len(allImages))
for i, image := range allImages{
    imageIDs[i] = image.ID
}
```

iaroslavscript:
Hi @ozerovandrei,
thanks for your remarks, I fixed it.

ozerovandrei:
@iaroslavscript could you please try to fix the issues with the commented-out acceptance tests?

Or can you please delete them if they are not working.

Unfortunately I don't know what can cause issues that you described.

iaroslavscript:
@ozerovandrei and @jtopjian thanks!

Issue #134, opened 2017-10-25
openstack_networking_router: enable/disable snat


Crapworks:
Hi there,

I learned today that some OpenStack providers have SNAT disabled by default. Of course, I'd like to enable it, so I can actually access the internet from instances without attached floating IPs...

The router resource doesn't support it, so I tried to add the argument, just to find out that this is not supported by Gophercloud. Do you think it is worth implementing? In this case, I could propose a PR for Gophercloud as well.

Cheers,
Christian

jtopjian:
@Crapworks What's the missing parameter?

The `openstack_networking_router_v2` resource supports something we call `value_specs` which is a funny way of saying "arbitrary key/values". `value_specs` are used for vendor-specific JSON parameters, but it's also a workaround for injecting missing parameters from Gophercloud.

Crapworks:
@jtopjian Yes, I know about that and it is very useful, but in this case, the parameter I need (`enable_snat` ) is inside the `external_gateway_info` structure. From the OpenStack API docs:

```json
{
    "router": {
        "name": "router1",
        "external_gateway_info": {
            "network_id": "ae34051f-aa6c-4c75-abf5-50dc9ac99ef3",
            "enable_snat": true,
            "external_fixed_ips": [
                {
                    "ip_address": "172.24.4.6",
                    "subnet_id": "b930d7f6-ceb7-40a0-8b81-a425dd994ccf"
                }
            ]
        },
        "admin_state_up": true
    }
}
```

And as far as I know I can't inject a parameter in a nested dict?

jtopjian:
Ah, I see.

You're more than welcome to open a Gophercloud PR. Looks like a simple patch. Just remember to find where `enable_snat` is defined in the actual Neutron code. :)

Crapworks:
Using my quite limited GO skills, I guess this is all that is needed to enable this? Looks a bit to easy for me...

https://github.com/gophercloud/gophercloud/pull/574

jtopjian:
@Crapworks You're good to go if you want to add this to Terraform.

You'll have to do a vendor update. To do this, you'll need to install `govendor` and then:

```shell
$ cd ~/go/src/github.com/terraform-providers/terraform-provider-openstack
$ govendor fetch github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers/...
$ git status # should only show the `routers` files being updated
$ git add .
$ git commit -m "vendor: Updating gophercloud
```

[...]

PR #140, opened 2017-10-27
openstack_networking_router: enable/disable snat

Crapworks:
Hi!

This is for #134

This PR adds the `enable_snat` property to the networking_router_v2 resource to enable/disable snat for neutron routers.

I've testet this against my openstack provider and it works fine! But again, I don't really know GO, so if this can be done more elegant, I'm open for suggestions :)

I am also not sure if it is the right thing to raise an error if `external_gateway` is not set, or if I should just ignore the parameter in that case.

Cheers,
Christian

### DIFF
--- a/openstack/resource_openstack_networking_router_v2.go
+++ b/openstack/resource_openstack_networking_router_v2.go
@@ -57,6 +57,12 @@ func resourceNetworkingRouterV2() *schema.Resource {
 				ForceNew: false,
 				Computed: true,
 			},
+			"enable_snat": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+				Computed: true,
+			},
 			"tenant_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
@@ -103,6 +109,14 @@ func resourceNetworkingRouterV2Create(d *schema.ResourceData, meta interface{}) 
 			NetworkID: externalGateway,
 		}
 		createOpts.GatewayInfo = &gatewayInfo
+	}
+
+	if esRaw, ok := d.GetOk("enable_snat"); ok {
+		if externalGateway == "" {
+			return fmt.Errorf("Error setting enable_snat: option requires external_gateway to be set")
+		}
+		es := esRaw.(bool)
+		createOpts.GatewayInfo.EnableSNAT = &es
 	}
 
 	log.Printf("[DEBUG] Create Options: %#v", createOpts)
@@ -153,6 +167,7 @@ func resourceNetworkingRouterV2Read(d *schema.ResourceData, meta interface{}) er
 	d.Set("distributed", n.Distributed)
 	d.Set("tenant_id", n.TenantID)
 	d.Set("external_gateway", n.GatewayInfo.NetworkID)
+	d.Set("enable_snat", n.GatewayInfo.EnableSNAT)
 	d.Set("region", GetRegion(d, config))
 
 	return nil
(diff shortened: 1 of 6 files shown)


openstack/resource_openstack_networking_router_v2.go
```diff
@@ -186,6 +201,14 @@ func resourceNetworkingRouterV2Update(d *schema.ResourceData, meta interface{})
...
 		}
 	}
+	if d.HasChange("enable_snat") {
+		enableSNAT := d.Get("enable_snat").(bool)
+		if updateOpts.GatewayInfo != nil {
+			updateOpts.GatewayInfo.EnableSNAT = &enableSNAT
+		} else {
+			return fmt.Errorf("Error setting enable_snat: option requires external_gateway to be set")
+		}
+	}
```
jtopjian: There's an issue with the logic here.

Let's say a user starts with this configuration:

```hcl
resource "openstack_networking_router_v2" "router_1" {
  name = "router"                                     
  admin_state_up = "true"                             
  distributed = "false"                               
  external_gateway = "asdf"                             
}                                                     
```
Then

```shell
$ terraform apply
```

And then they decide to disable SNAT:

```hcl
resource "openstack_networking_router_v2" "router_1" {
  name = "router"                                     
  admin_state_up = "true"                             
  distributed = "false"                               
  external_gateway = "asdf"
  enable_snat = false
}                                                     
```

`external_gateway` isn't being updated, so the code within `d.HasChange("external_gateway")` is never triggered. When the user runs another `terraform apply`, they'll get an error.

What about something like this?

```go
var gatewayInfoChanged bool
gatewayInfo := routers.GatewayInfo{}

if d.HasChange("external_gateway") {
	externalGateway := d.Get("external_gateway").(string)
	if externalGateway != "" {
                gatewayInfoChanged = true
                gatewayInfo.NetworkID = externalGateway
	}
}
if d.HasChange("enable_snat") {
        externalGateway := d.Get("external_gateway").(string)
        if externalGateway == "" {
		return fmt.Errorf("Error setting enable_snat: option requires external_gateway to be set")
        }
	enableSNAT := d.Get("enable_snat").(bool)
        gatewayInfoChanged = true
        gatewayInfo.EnableSNAT = &enableSNAT
        gatewayInfo.NetworkID = externalGateway
}

if gatewayInfoChanged {
        updateOpts.GatewayInfo = &gatewayInfo
}
```

Note how `gatewayInfo.NetworkID` is being set in the `enable_snat` block of code. The two must be sent together.

Crapworks:
@jtopjian Oh, you're right. I wasn't testing this scenario. I am just wondering, if I do something like this:

```go
    gatewayInfo := routers.GatewayInfo{}
    externalGateway := d.Get("external_gateway").(string)
    if externalGateway != "" { 
        gatewayInfo.NetworkID = externalGateway
    } 
    if d.HasChange("external_gateway") { 
        updateOpts.GatewayInfo = &gatewayInfo
    } 
    if d.HasChange("enable_snat") { 
        enableSNAT := d.Get("enable_snat").(bool)
        gatewayInfo.EnableSNAT = &enableSNAT
        if gatewayInfo.NetworkID != "" { 
            updateOpts.GatewayInfo = &gatewayInfo
        } else { 
            return fmt.Errorf("Error setting enable_snat: option requires external_gateway to be set")
        } 
    } 
```

It should work as well, even without having an additional variable that checks for changes, right? Or am I missing something?

jtopjian:
Yep, that works, too :)

Crapworks:
@jtopjian I've made the changes

jtopjian:
Looks good! Thank you for working on this.

Issue #142, opened 2017-10-31
feature request: designate zone datasource


jrperritt:
This is a bug for creating a data source for Designate zones. I'll submit a PR for this.

jtopjian:
Implemented in #145

PR #145, opened 2017-11-02
dns zone v2 data source

jrperritt:
for #142

### DIFF
--- a/openstack/data_source_openstack_dns_zone_v2.go
+++ b/openstack/data_source_openstack_dns_zone_v2.go
@@ -1,0 +1,199 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud/openstack/dns/v2/zones"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceDNSZoneV2() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDNSZoneV2Read,
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"pool_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"email": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"status": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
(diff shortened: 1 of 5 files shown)


stamak:
cool
@jrperritt does this PR cover a cases when I do not use/have OpenStack DNS?

openstack/data_source_openstack_dns_zone_v2_test.go
```diff
@@ -0,0 +1,64 @@
...
+			return fmt.Errorf("DNS Zone data source ID not set")
+		}
+
+		return nil
+	}
+}
+
+const testAccOpenStackDNSZoneV2DataSource_zone = `
+resource "openstack_dns_zone_v2" "z1" {
+        name = "terraform-dns-zone-v2-test.com."
```
jtopjian: I've only done a quick skim of this so far, but caught this. The name will need to be randomly generated (like is done [here](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_dns_zone_v2_test.go)) to prevent a false test failure if the zone already exists (for whatever reason).

jrperritt: Thanks, I'll make the change

jrperritt:
@stamak This Terraform provider _is_ for OpenStack, so this data source is only able to communicate with OpenStack-compatible environments. However, it just communicates with OpenStack via its API, so I'd guess any DNS provider that is configured to work with the Designate API should work.

openstack/data_source_openstack_dns_zone_v2.go
```diff
@@ -0,0 +1,199 @@
...
+	d.Set("version", zone.Version)
+	d.Set("serial", zone.Serial)
+
+	// time.Times
+	d.Set("created_at", zone.CreatedAt.String())
+	d.Set("updated_at", zone.UpdatedAt.String())
+	d.Set("transferred_at", zone.TransferredAt.String())
+
+	// maps
+	d.Set("attributes", zone.Attributes)
```
jtopjian: I recently learned that all non-scalar types should be checked for errors when being `Set`. Trying to enforce this from this point on.

If it's not essential for this field to be set, the error can simply be logged.

openstack/data_source_openstack_dns_zone_v2.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// ints
+	d.Set("ttl", zone.TTL)
+	d.Set("version", zone.Version)
+	d.Set("serial", zone.Serial)
+
+	// time.Times
+	d.Set("created_at", zone.CreatedAt.String())
+	d.Set("updated_at", zone.UpdatedAt.String())
+	d.Set("transferred_at", zone.TransferredAt.String())
```
jtopjian: I completely overlooked this, but the `image` resource and data source also have timestamp fields, but they're not working because an automatic string conversion isn't being done.

Looking at the AWS provider (which I use as a standard), timestamps are being formatted either as `time.RFC1123` or `time.RFC3339`. I guess depending on the point of displaying this data (readability or parsabiilty) will decide on which one to use. I feel like the latter format is a safe one to go with.

jtopjian:
Looks great! Thank you!

jrperritt:
Thank you for reviewing so quickly

everett-toews:
Thanks @jtopjian. This is a big help for us. GFY!

PR #146, opened 2017-11-02
Add resource openstack_objectstorage_object_v1

### LINKED ISSUE #144: Missing openstack_objectstorage_object_v1 resource
will work on it as the gophercloud lib has been updated with an objectstorage/v1/objects api.

yanndegat:
Closes #144

### DIFF
--- a/openstack/resource_openstack_objectstorage_object_v1.go
+++ b/openstack/resource_openstack_objectstorage_object_v1.go
@@ -116,7 +116,7 @@ func resourceObjectStorageObjectV1() *schema.Resource {
 				ConflictsWith: []string{"content", "copy_from", "object_manifest"},
 			},
 
-			//Read Only
+			// Read Only
 			"content_length": {
 				Type:     schema.TypeInt,
 				Computed: true,
@@ -197,7 +197,7 @@ func resourceObjectStorageObjectV1Put(d *schema.ResourceData, meta interface{}) 
 	}
 
 	if !isValid {
-		return fmt.Errorf("Must specify \"source\", \"content\" or \"copy_from\" field")
+		return fmt.Errorf("Must specify \"source\", \"content\", \"copy_from\" or \"object_manifest\" field")
 	}
 
 	if v, ok := d.GetOk("content_disposition"); ok {
@@ -229,7 +229,7 @@ func resourceObjectStorageObjectV1Put(d *schema.ResourceData, meta interface{}) 
 		createOpts.DetectContentType = "true"
 	}
 
-	//this attribute is used to trigger resource updates if the file content is changed
+	// this attribute is used to trigger resource updates if the file content is changed
 	if v, ok := d.GetOk("etag"); ok {
 		createOpts.ETag = v.(string)
 	}
--- a/openstack/resource_openstack_objectstorage_object_v1.go
+++ b/openstack/resource_openstack_objectstorage_object_v1.go
@@ -1,0 +1,330 @@
+package openstack
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/mitchellh/go-homedir"
+)
+
+func resourceObjectStorageObjectV1() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceObjectStorageObjectV1Put,
+		Read:   resourceObjectStorageObjectV1Read,
+		Update: resourceObjectStorageObjectV1Put,
+		Delete: resourceObjectStorageObjectV1Delete,
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"container_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"content_disposition": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"copy_from"},
+			},
+
+			"content_encoding": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"copy_from"},
+			},
+
+			"content_type": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"copy_from"},
+			},
+
+			"content": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"source", "copy_from", "object_manifest"},
+			},
+
+			"copy_from": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"content", "source", "object_manifest"},
+			},
+
+			"delete_after": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+
+			"delete_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
+			},
+
+			"detect_content_type": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			// this attribute is used to trigger resource updates
+			// if the file content is changed
+			"etag": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"metadata": &schema.Schema{
+				Type:     schema.TypeMap,
+				Optional: true,
+			},
+
+			"object_manifest": {
+				Type:          schema.TypeString,
+				Computed:      true,
+				Optional:      true,
+				ConflictsWith: []string{"copy_from", "source", "content"},
+			},
+
+			"source": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"content", "copy_from", "object_manifest"},
+			},
+
+			//Read Only
+			"content_length": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"date": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"last_modified": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"trans_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceObjectStorageObjectV1Put(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	objectStorageClient, err := config.objectStorageV1Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack object storage client: %s", err)
+	}
+
+	name := d.Get("name").(string)
+	cn := d.Get("container_name").(string)
+
+	createOpts := &objects.CreateOpts{
+		Metadata:         resourceObjectMetadataV1(d),
+		TransferEncoding: "chunked",
+	}
+
+	var isValid bool
+	if v, ok := d.GetOk("source"); ok {
+		isValid = true
+		source := v.(string)
+		path, err := homedir.Expand(source)
+		if err != nil {
+			return fmt.Errorf("Error expanding homedir in source (%s): %s", source, err)
+		}
+		file, err := os.Open(path)
+		defer file.Close()
+		if err != nil {
+			return fmt.Errorf("Error opening openstack swift object source (%s): %s", source, err)
+		}
+		fileinfo, err := file.Stat()
+		if err != nil {
+			return fmt.Errorf("Error opening openstack swift object source (%s): %s", source, err)
+		}
+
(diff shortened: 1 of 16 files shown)


yanndegat:
@jtopjian

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack container object: %s", err)
+	}
+	log.Printf("[INFO] Object %s has been added to container : %s", name, cn)
+
+	d.Set("etag", result.ETag)
+	d.Set("content_length", result.ContentLength)
+	d.Set("date", fmt.Sprintf("%v", result.Date))
+	d.Set("content_type", result.ContentType)
+	d.Set("last_modified", fmt.Sprintf("%v", result.LastModified))
```
jtopjian: For the exported/read-only time fields, would you object to formatting it as RFC3339? This coincidentally came up [yesterday](https://github.com/terraform-providers/terraform-provider-openstack/pull/145#discussion_r148706898). It would be nice if all exported time is a uniform format, but I am totally fine if that's not possible.

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+	d.Set("trans_id", result.TransID)
+	// Store the ID now
+	d.SetId(fmt.Sprintf("%s_%s_%v", cn, name, result.ETag))
```
jtopjian: According to the [Swift API documentation](https://developer.openstack.org/api-ref/object-store/#create-or-replace-object), when an object's content is updated, the etag will change, which will give this resource a new ID. Would it be better to have the ID set to only `cn_name` (or `cn/name`)? This way, when either change, a new resource is already being created.

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+
+	if v, ok := d.GetOk("multipart_manifest"); ok && v.(bool) {
+		deleteOpts.MultipartManifest = "delete"
+	}
+
+	_, err = objects.Delete(objectStorageClient, cn, name, deleteOpts).Extract()
+	if err != nil {
+		return fmt.Errorf("Error getting OpenStack container object: %s", err)
+	}
+	d.SetId("")
```
jtopjian: Minor nit: `d.SetId("")` is no longer required. Terraform does it automatically. This can safely be removed everywhere. Older resources haven't had this cleaned up, and this sometimes slips by during code review.

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+		createOpts.Content = file
+	} else if v, ok := d.GetOk("content"); ok {
+		content := v.(string)
+		createOpts.Content = bytes.NewReader([]byte(content))
+	} else if v, ok := d.GetOk("copy_from"); ok {
+		createOpts.CopyFrom = v.(string)
+		createOpts.Content = bytes.NewReader([]byte(""))
+	} else {
+		return fmt.Errorf("Must specify \"source\", \"content\" or \"copy_from\" field")
+	}
```
jtopjian: This collection of `if/else` statements can be simplified. Since you are using `ConflictsWith`, you can be sure that only one will ever be set. 

As mentioned in another comment, let's get rid of `source`, so that leaves `content` and `copy_from`. You can then do:

```go
var isValid bool
if v, ok := d.GetOk("content"); ok {
  isValid = true
  content := v.(string)
  createOpts.Content ...
}

if v, ok := d.GetOk("copy_from"); ok {
  isValid = true
  ...
}

if !isValid {
  return fmt.Errorf("Must specify ...")
}
```

IMO, this is more readable because `else if` statements usually require the reader to understand all conditionals. But it's only my opinion -- I won't enforce it :)

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"copy_from"},
+			},
+
+			"content_length": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
```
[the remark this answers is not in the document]
yanndegat: great, will do

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+				Computed:      true,
+				Optional:      true,
+				ConflictsWith: []string{"copy_from"},
+			},
+
+			"source": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"content", "copy_from"},
+			},
```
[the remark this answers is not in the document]
yanndegat: well, in fact, that's what i thought first, but for this, i got inspiration from the aws_s3_bucket_object resource. This is how they treat the content of an object.
i've tried to use the "file()" function with content on a large file, and in fact the content a the file is loaded in the state file, which you might not want to for large files.

what do you think ?

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: again, a paste from aws_s3_bucket_object resource. 
but will replace else if by ifs only. makes sense.

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: nice, will do

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: will do, my bad

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: ok

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,335 @@
...
+
+	_, err = objects.Delete(objectStorageClient, cn, name, deleteOpts).Extract()
+	if err != nil {
+		return fmt.Errorf("Error getting OpenStack container object: %s", err)
+	}
+	d.SetId("")
+	return nil
+}
+
+func resourceObjectMetadataV2(d *schema.ResourceData) map[string]string {
```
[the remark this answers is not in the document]
yanndegat: yup, bad copy/paste

openstack/resource_openstack_objectstorage_object_v1.go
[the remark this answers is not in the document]
yanndegat: ok

openstack/resource_openstack_objectstorage_object_v1.go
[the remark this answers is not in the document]
yanndegat: this is the way they treat the content of large files in the aws_s3_bucket_object.
if think it's because we don't want large files to be loaded in the terraform state file.

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: will do

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: will do

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: will do, my bad

openstack/resource_openstack_objectstorage_object_v1.go
yanndegat: will do

openstack/resource_openstack_objectstorage_object_v1.go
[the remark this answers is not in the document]
yanndegat: bad copy/paste ;)

yanndegat:
hi @jtopjian 

thanks a lot for your review. it's always great to get qualified feedback that fast on terraform.

1. the "vendor" files should be in a dedicated commit with a message of:
   "vendor: Updating gophercloud for object support"
ok, will do

2. Given the amount of input arguments, we'll want to include a few more acceptance tests (or add more arguments/attributes to the existing tests).
arf, ;) i somehow knew that one would come up ;)

3. I need to think about the CreateOrUpdate functionality here. I understand why you're doing it, but it's the first I've seen it used in Terraform. Let's start with this first round of changes and I'll revisit. 
i got this from the aws_s3_bucket_object.
they called their function "ObjectPut", maybe i should stick to "Put" ?


I'm not very familiar with the Swift API either but what i can say is that
this resource is mostly useful for "examples" purposes only. Or a very few usecases.

I don't think some would rely on terraform to upload their files in their containers in the long term.

For instance, here's the usecase i'm actually implementing: 
Launch a Spark job to process a large dataset.
- create the swift container
- upload the dataset
- create the spark cluster
- submit the job
- tear down all resources.

but this is an example usecase. Because in a real scenario, you would launch
your spark job on an existing dataset preuploaded.
And the output of the spark job would probably end up in the swift container so you don't want
terraform to destroy the container in the end.

openstack/resource_openstack_objectstorage_object_v1.go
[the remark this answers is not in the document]
jtopjian: You're absolutely correct on both:

1. It's done in `aws_s3_bucket_object`. When in doubt, I always defer to what's being done in AWS.
2. Using `file` would indeed cause large amounts of data to be stored in state.

One other option is to configure the state to only store a checksum of the content. However, I tried this with the [LXD provider](https://github.com/sl1pm4t/terraform-provider-lxd/blob/master/lxd/resource_lxd_container.go#L114) and am running into some _weird_ bugs where the hash (not the content) is randomly being used as the content when there's a high amount of parallel activity.

So yes, let's keep source.

openstack/resource_openstack_objectstorage_object_v1.go
jtopjian: You're more than welcome to keep this as-is. I was only pushing my own style :smile:

jtopjian:
@yanndegat 

> i got this from the aws_s3_bucket_object.
> they called their function "ObjectPut", maybe i should stick to "Put" ?

OK, perfect. If there's precedence for this, then I am good with it. Yes, let's change it to "Put" so it's more visible that this was copied from AWS.

> I'm not very familiar with the Swift API either but what i can say is that
> this resource is mostly useful for "examples" purposes only. Or a very few usecases.

I'm inclined to agree with you. But one reason I enjoy working on this stuff is to provide the capability to do things we only recognize as "examples" and then have others discover other clever cases :)

I think an object data source would also be quite useful, too. I can imagine storing a provisioning script in an object and have the content passed to a provisioner.

yanndegat:
@jtopjian should be updated accordingly

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,330 @@
...
+				ConflictsWith: []string{"copy_from", "source", "content"},
+			},
+
+			"source": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"content", "copy_from", "object_manifest"},
+			},
+
+			//Read Only
```
jtopjian: Space between `//` and `Read`:

```
// Read Only
```

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+
+		createOpts.DeleteAt = int(t.Unix())
+	}
+
+	if v, ok := d.GetOk("detect_content_type"); ok && v.(bool) {
+		createOpts.DetectContentType = "true"
+	}
+
+	//this attribute is used to trigger resource updates if the file content is changed
```
jtopjian: Space between `//` and `this`.

openstack/resource_openstack_objectstorage_object_v1.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if v, ok := d.GetOk("object_manifest"); ok {
+		isValid = true
+		createOpts.Content = bytes.NewReader([]byte(""))
+		createOpts.ObjectManifest = v.(string)
+	}
+
+	if !isValid {
+		return fmt.Errorf("Must specify \"source\", \"content\" or \"copy_from\" field")
```
jtopjian: Should this now include `object_manifest`?

yanndegat:
hi @jtopjian 

well the acc tests are working on my side because:
- i've set an input date string in french TZ
- swift server is in french tz
- GetHeader parsing returns date in french TZ

so what would be the proper way to write acc tests which would not rely on the timezone?
with a custom attr check function?

jtopjian:
@yanndegat Very good point. I think it's very probable that someone would use a time zone different than the time zone of the OpenStack Swift service, so making sure the time is correct regardless of the time zone is important. I think the best way to go about this is to use a `DiffSuppressFunc`. 

Implementing the `DiffSupressFunc` wouldn't have been too difficult, but figuring out how to accurately test for it was. I went ahead and wrote it up and submitted a PR to your local copy. You can simply merge it and it'll be reflected in this PR.

yanndegat:
@jtopjian thanks a lot for the PR
i had no idea of this diffsuppressfunc

i worked on my side and i also  pushed the typos

jtopjian:
@yanndegat I had a feeling this resource was going to be a little unwieldy, so thank you _very_ much for your help and patience with it :smile:

Issue #147, opened 2017-11-03
feature request: add a field to LB operations to allow use of Octavia


jrperritt:
Currently, all the LB operations go through Neutron. This bug is for enabling the ability to optionally provide a value (I was thinking just a `use_octavia` key that took a `bool`) to the `*_openstack_lb_*_v2` resources that would, instead of creating a Neutron service client, create an Octavia service client to make the calls. If that sounds reasonable, I'll submit a PR for this.

cc @jtopjian

jtopjian:
Sorry - I totally missed this one yesterday. At first read, this makes sense. I'll have some time to read this over soon.

jrperritt:
No worries, I'm still fumbling over the implementation, which _should be_ simple.

jtopjian:
Implemented in #149

PR #149, opened 2017-11-03
add option to use octavia service for lb operations

jrperritt:
for #147

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	Token            string
 	Username         string
 	UserID           string
+	useOctavia       bool
 
 	OsClient *gophercloud.ProviderClient
 }
@@ -207,6 +208,13 @@ func (c *Config) objectStorageV1Client(region string) (*gophercloud.ServiceClien
 	}
 
 	return openstack.NewObjectStorageV1(c.OsClient, gophercloud.EndpointOpts{
+		Region:       c.determineRegion(region),
+		Availability: c.getEndpointType(),
+	})
+}
+
+func (c *Config) loadBalancerV2Client(region string) (*gophercloud.ServiceClient, error) {
+	return openstack.NewLoadBalancerV2(c.OsClient, gophercloud.EndpointOpts{
 		Region:       c.determineRegion(region),
 		Availability: c.getEndpointType(),
 	})
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners"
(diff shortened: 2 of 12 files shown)


jrperritt:
OK, this one's ready for a first look.

jrperritt:
I went with one `use_octavia` config field in the provider. Originally I had a field in each LB V2 resource schema but, as was pointed out to me, having even one of those not set to true while another is would fail to deploy correctly.

jtopjian:
> having even one of those not set to true while another is would fail to deploy correctly.

Good call. Provider-level is definitely the way to go.

I like `use_octavia` because it's a decisive boolean, but so this discussion can be done now instead of later (and also so the discussion is documented): is there benefit to instead having `loadbalancer_service` with options of `neutron` and `octavia`, defaulting to `neutron`?

`loadbalancer_service` accounts for a future possibility of yet another load balancing service. That might never happen. I can also see a user accidentally setting it to the actual load balancing driver like "f5".

All in all, this change looks good to me. I'll need to run some compiled tests before and after this patch to note changes (if any) to an existing state file. Just a heads up that this might take me a day or two to get to.

jrperritt:
> is there benefit to instead having loadbalancer_service with options of neutron and octavia, defaulting to neutron?

Great question. This `use_octavia` option is only necessary until they deprecate (or, maybe, remove) LB operations in Neutron, at which point it would be deprecated here, so that's unfortunate. 

When creating an LB in Neutron, there is a `loadbalancer_provider` option. If you set that to "octavia", then Neutron will proxy LB requests to Octavia. From what I've been told (and what we've experienced at $WORK), the Neutron and Octavia databases can get out of sync, and the only way to fix it is to make manual changes to the database.

jrperritt:
Another option I hadn't considered for this is to not have a config field and instead just look for an environment variable. So instead of having `use_octavia` with a default of `$OS_USE_OCTAVIA`, it would just look for `$OS_USE_OCTAVIA` when the config is created. That would prevent creating the specific and temporary config field. Not sure if that would be better.

jtopjian:
I think the way it's configured now is fine.

I'm not concerned about the field being temporary. In fact, I think it'll stick around for quite a while. When Octavia becomes the official load balancer service, there will still be a lot of Neutron-based lb services around. Once Octavia has become the majority deployed, `use_octavia` can be set to a default value of `true`.

If/when the time comes to remove it, Terraform has an excellent process of deprecating fields: the field gets tagged as `deprecated` but still works, then it gets tagged as `removed` which allows an error, but doesn't break schemas.

jrperritt:
> I get an error when setting this to true, which is expected since I don't have the Octavia service, so it's definitely working.

Yeah, I couldn't think of a good way to test this (as in, a good test to write). From the user's perspective, everything should look the same whether using Neutron or Octavia.

jtopjian:
LGTM - thank you for adding this.

jrperritt:
@jtopjian Thank you for reviewing and testing this.

I also wanted to ask inre releases: I took a look at the existing releases, and it looks like maybe they are created as-needed/as-asked-for? Asking because we'd be interested in seeing a release that has this PR as well as the DNS V2 zone data source.

jtopjian:
Indeed - releases are done ad hoc at the moment.

I'll look at getting a release out this week. There seems to be some flaky AWS issues which are causing the nightly CI to fail. Once I can get it to go green, we should be good.

jrperritt:
Great, thanks :)

Issue #141, opened 2017-10-31
trove support


eugenetaranov:
Hello,

I would like to send a few pull requests adding support for trove, but got questions about acceptance tests: 
- is there a quick guid how to run it on devstack?
- would it be possible to run acceptance tests on your side?

Thanks you

jtopjian:
@eugenetaranov Thank you for offering.

Do you have an existing Trove environment you can use? If so, you can run acceptance tests on that environment.

There's some basic documentation about acceptance tests at the bottom of this page:

https://www.terraform.io/docs/providers/openstack/index.html

There's some more information here:

https://github.com/gophercloud/gophercloud/tree/master/acceptance

I wrote the series of blog posts linked in the above page specifically for Gophercloud testing (which Terraform is compatible with).

To run the Terraform acceptance tests, make sure you have the required environment variables set and then run:

```shell
$ cd ~/go/src/github.com/terraform-providers/terraform-provider-openstack/
$ TF_LOG=DEBUG make testacc TEST=./openstack TESTARGS="-run=TestAccComputeV2Instance_basic"
```

Where `TestAccComputeV2Instance_basic` is a test defined in one of the `*_test.go` files.

The above test is safe to run in a production OpenStack cloud if you wanted to try it out.

If you have problems running the tests, I can certainly run them on my side, though the caveat is that development of the resources will take a little longer.

eugenetaranov:
@jtopjian thank you for provided instructions, it helped a lot. I got acceptance tests running on production openstack but there was a minor hiccup: TestAccComputeV2Instance_basic fails with "Multiple possible networks found, use a Network ID to be more specific" error, I fixed by specifying network `network { uuid = "..." }`
But for some reason it worked okay with packstack (without network being specified). Do you think it is something that needs to be fixed? If so, I will send a PR.

jtopjian:
@eugenetaranov There are usually two causes of this:

1. You're running the tests as the "admin" user. The tests should be run by the "demo" user. There are some tests which require the admin user -- they're just skipped for now.
2. Your demo user has multiple networks, possibly ones that existed before the testing or some which were not cleaned up after testing.

Let me know if this helps.

eugenetaranov:
Hi @jtopjian , thanks for explanation, it was #2 in my case: got several networks in my account.

jtopjian:
[...]

PR #155, opened 2017-11-13
openstack_db_instance resource

eugenetaranov:
This PR adds openstack_db_instance resource and addresses https://github.com/terraform-providers/terraform-provider-openstack/issues/141

Acceptance tests require two extra vars to be set:
`
OS_DATASTORE_VERSION=
OS_DATASTORE_TYPE=
`

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -220,6 +220,13 @@ func (c *Config) loadBalancerV2Client(region string) (*gophercloud.ServiceClient
 	})
 }
 
+func (c *Config) databaseInstanceClient(region string) (*gophercloud.ServiceClient, error) {
+	return openstack.NewDBV1(c.OsClient, gophercloud.EndpointOpts{
+		Region:       c.determineRegion(region),
+		Availability: c.getEndpointType(),
+	})
+}
+
 func (c *Config) getEndpointType() gophercloud.Availability {
 	if c.EndpointType == "internal" || c.EndpointType == "internalURL" {
 		return gophercloud.AvailabilityInternal
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -220,7 +220,7 @@ func (c *Config) loadBalancerV2Client(region string) (*gophercloud.ServiceClient
 	})
 }
 
-func (c *Config) databaseInstanceClient(region string) (*gophercloud.ServiceClient, error) {
+func (c *Config) databaseV1Client(region string) (*gophercloud.ServiceClient, error) {
 	return openstack.NewDBV1(c.OsClient, gophercloud.EndpointOpts{
 		Region:       c.determineRegion(region),
 		Availability: c.getEndpointType(),
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -168,6 +168,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_compute_floatingip_v2":           resourceComputeFloatingIPV2(),
 			"openstack_compute_floatingip_associate_v2": resourceComputeFloatingIPAssociateV2(),
 			"openstack_compute_volume_attach_v2":        resourceComputeVolumeAttachV2(),
+			"openstack_db_instance_v1":                  resourceDatabaseInstanceV1(),
 			"openstack_dns_recordset_v2":                resourceDNSRecordSetV2(),
 			"openstack_dns_zone_v2":                     resourceDNSZoneV2(),
 			"openstack_fw_firewall_v1":                  resourceFWFirewallV1(),
@@ -195,7 +196,6 @@ func Provider() terraform.ResourceProvider {
 			"openstack_networking_secgroup_v2":          resourceNetworkingSecGroupV2(),
 			"openstack_networking_secgroup_rule_v2":     resourceNetworkingSecGroupRuleV2(),
 			"openstack_objectstorage_container_v1":      resourceObjectStorageContainerV1(),
-			"openstack_db_instance":                     resourceDatabaseInstance(),
 		},
 
 		ConfigureFunc: configureProvider,
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -195,6 +195,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_networking_secgroup_v2":          resourceNetworkingSecGroupV2(),
 			"openstack_networking_secgroup_rule_v2":     resourceNetworkingSecGroupRuleV2(),
 			"openstack_objectstorage_container_v1":      resourceObjectStorageContainerV1(),
+			"openstack_db_instance":                     resourceDatabaseInstance(),
 		},
 
 		ConfigureFunc: configureProvider,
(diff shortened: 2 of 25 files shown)


ronaldtse:
Thank you @eugenetaranov! We are already running this in pre-production.

openstack/config.go
```diff
@@ -213,6 +213,13 @@ func (c *Config) objectStorageV1Client(region string) (*gophercloud.ServiceClien
 	})
 }
 
+func (c *Config) databaseInstanceClient(region string) (*gophercloud.ServiceClient, error) {
```
jtopjian: This should be `databaseV1Client` since it can be used for all future Trove/DBaaS V1 resources.

openstack/provider.go
```diff
@@ -195,6 +195,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_networking_secgroup_v2":          resourceNetworkingSecGroupV2(),
 			"openstack_networking_secgroup_rule_v2":     resourceNetworkingSecGroupRuleV2(),
 			"openstack_objectstorage_container_v1":      resourceObjectStorageContainerV1(),
+			"openstack_db_instance":                     resourceDatabaseInstance(),
```
jtopjian: The name of the resource should be `openstack_db_instance_v1` to match the Trove/DBaaS API version. It should also be in alphabetical order in this list.

openstack/provider_test.go
```diff
@@ -24,6 +24,8 @@ var (
 	OS_POOL_NAME              = os.Getenv("OS_POOL_NAME")
 	OS_REGION_NAME            = os.Getenv("OS_REGION_NAME")
 	OS_SWIFT_ENVIRONMENT      = os.Getenv("OS_SWIFT_ENVIRONMENT")
+	OS_DATASTORE_VERSION      = os.Getenv("OS_DATASTORE_VERSION")
+	OS_DATASTORE_TYPE         = os.Getenv("OS_DATASTORE_TYPE")
```
jtopjian: minor fix: It's very helpful to have these match Gophercloud's environment variables. So these should be `OS_DB_DATASTORE_VERSION` and `OS_DB_DATASTORE_TYPE`.

openstack/resource_openstack_db_instance.go
```diff
@@ -0,0 +1,253 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+					},
+				},
+			},
+			"network": {
+				Type:     schema.TypeList,
+				Required: true,
```
jtopjian: It's possible to create a database instance without specifying a network. If the tenant only has one network available, that network will automatically be used.

openstack/resource_openstack_db_instance.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf(
+			"Error waiting for instance (%s) to delete: %s",
+			d.Id(), err)
+	}
+
+	d.SetId("")
```
jtopjian: This is no longer needed. It exists in other resources out of laziness :)

openstack/resource_openstack_db_instance.go
```diff
@@ -0,0 +1,253 @@
...
+	return nil
+}
+
+// InstanceStateRefreshFunc returns a resource.StateRefreshFunc that is used to watch
+// an cloud database instance.
+func InstanceStateRefreshFunc(client *gophercloud.ServiceClient, instanceID string) resource.StateRefreshFunc {
```
jtopjian: So this doesn't get confused with a Compute Instance, let's change this to `DatabaseInstanceV1StateRefreshFunc`.

jtopjian:
@ronaldtse Thank you for 7c9b82e, however I have to kindly ask that the copyright headers are removed. This Terraform project is already under the MPL which already implies the disclaimers made in the headers you've attached.

I apologize for any inconvenience with this and we are happy to discuss this further if you'd like.

ronaldtse:
@jtopjian could we keep the copyright header (without the disclaimer)? All our open source contributions contain our copyright signature, e.g., https://github.com/openssl/openssl/pulls/4552.

This does not conflict with your MPL.

catsby:
Hey @ronaldtse I'm sorry to say that in order to accept the pull request we'd need that removed. We've consulted lawyers and have been advised to remove those in favor of maintaining the MPL2 license in the root of the repository. I hope that doesn't cause an issue on your end, but unfortunately I don't believe we can accept this code contribution as-is. 

Thank you for your understanding, I do hope we can resolve this and accept your contributions.

ronaldtse:
@catsby sure. According to our IP lawyer, the copyright header should not affect any rights offered under the MPL. That said, we do understand your concerns, and since that's the only way to proceed, we will drop the line. On the other hand, it would however be appreciated if there could be an acknowledgement somewhere about the contribution, e.g., in the CHANGELOG. 😉 

Thanks!

ronaldtse:
Reverted and rebased.

openstack/provider_test.go
ronaldtse: Addressed.

openstack/provider.go
ronaldtse: Addressed.

openstack/config.go
ronaldtse: Addressed.

openstack/resource_openstack_db_instance.go
ronaldtse: Addressed.

openstack/resource_openstack_db_instance.go
ronaldtse: @jtopjian can you clarify it is line 231 (`d.SetID("")`) that is no longer needed, or line 228 the `d.ID()` part?

openstack/resource_openstack_db_instance.go
ronaldtse: @jtopjian we've actually run into "creation failed" problems when we didn't specify the network on Ocata. @eugenetaranov is this still the case?

openstack/resource_openstack_db_instance.go
jtopjian: Line 231 is no longer needed. It's automatically handled within Terraform.

openstack/resource_openstack_db_instance.go
jtopjian: I apologize, I might have been unclear: if the tenant has multiple networks available to launch on, then they must specify a network.

However, if a tenant only has one network available, it will be implicitly used. Because of this case, the `network` field should be set to optional.

eugenetaranov:
@jtopjian , thank you for your review and please do one more :)

eugenetaranov:
@jtopjian thank you for feedback, updated

jtopjian:
I made a mistake - the state is actually `SHUTDOWN`. That's OK, though, I will update that after this commit.

Thank you both for your patience and work on this :)

ronaldtse:
Thank you and really appreciate it @jtopjian ! 👍

PR #160, opened 2017-11-20
[WIP] Add support for ipv6_ra/address_mode and availability_zone_hints

stone:
Hi!

This PR scratches some itches we got to use terraform fully, so this PR adds support
for:

**Subnets:**
Adds support for Stateless Address Autoconfiguration (SLAAC), Stateful DHCPv6, and Stateless DHCPv6 on a subnet.

 * i[pv6_ra_mode](https://developer.openstack.org/api-ref/network/v2/index.html#create-subnet)
 * [ipv6_address_mode](https://developer.openstack.org/api-ref/network/v2/index.html#create-subnet)

```
resource "openstack_networking_subnet_v2" "subnet_ipv6" {
  name                = "terraform-subnet-ipv6"
  network_id          = "${openstack_networking_network_v2.network.id}"
  cidr                = "ffff:c0:1f0:1002::/64"
  ipv6_address_mode   = "slaac"
  ipv6_ra_mode        = "slaac"
  ip_version          = 6
  enable_dhcp         = "true"
  gateway_ip          = "ffff:c0:1f0:1002::1"
}

```

**Network**
This allows users to associate an availability zone with their resources so that the resources get high availability.

 * [availability_zone_hints](https://developer.openstack.org/api-ref/network/v2/index.html#networks) (missing in docs but supported by openstack cli and is about to be added to [webui](https://git.openstack.org/cgit/openstack/horizon/commit/?id=b9664a1bf139e6714aa4d507a68ce03d551779aa))

```
  name               = "terraform-ipv6"
  admin_state_up = "true"
  shared             = "false"
  availability_zone_hints = ["zone1", "zone2"]
}

```

**Router**
This allows users to associate an availability zone with their resources so that the resources get high availability.
 * [availability_zone_hints](https://developer.openstack.org/api-ref/network/v2/index.html#create-router)

```
 resource "openstack_networking_router_v2" "testrouter" {
  name                      = "testrouter"
  availability_zone_hints   = ["zone1"]
}
```

Would be nice if someone could have a look at the PR and see if/what I need to fix for it to be accepted. Uncertain if the test case that are included are correct.

// Fredrik

### DIFF
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -56,6 +56,12 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"availability_zone_hints": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
 		},
 	}
 }
@@ -111,6 +117,11 @@ func dataSourceNetworkingNetworkV2Read(d *schema.ResourceData, meta interface{})
 	}
 
 	network := refinedNetworks[0]
+
+	err = d.Set("availability_zone_hints", network.AvailabilityZoneHints)
+	if err != nil {
+		log.Printf("[DEBUG] Unable to set availability_zone_hints: %s", err)
+	}
 
 	log.Printf("[DEBUG] Retrieved Network %s: %+v", network.ID, network)
 	d.SetId(network.ID)
--- a/openstack/data_source_openstack_networking_network_v2_test.go
+++ b/openstack/data_source_openstack_networking_network_v2_test.go
@@ -74,6 +74,28 @@ func TestAccOpenStackNetworkingNetworkV2DataSource_networkID(t *testing.T) {
 	})
 }
 
+func TestAccOpenStackNetworkingNetworkV2DataSource_availability_zone_hints(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccOpenStackNetworkingNetworkV2DataSource_network,
+			},
+			resource.TestStep{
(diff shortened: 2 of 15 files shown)


jtopjian:
@stone Thank you for submitting this. I agree these would all be helpful additions.

However, it looks like you have modified the vendor files outside of the actual `github.com/gophercloud/gophercloud` repo. In order for Terraform to take advantage of these, they need merged into the upstream Gophercloud library.

The good news is that I help maintain that library, so I can help you with this. 

Just a heads up that PRs to Gophercloud can be a little strict, but it's for good reason. To begin, you'll first need to open an Issue that describes the feature (one feature per issue) as well as points to the OpenStack source code where the feature is implemented. For example, to implement routers `ListOpts.AvailabilityZones` field, you'll need to search https://github.com/openstack/neutron to find where `availability_zones` is defined ([here](https://github.com/openstack/neutron/blob/04d29e9aedd643cef46caec5b0496badc28e6f73/neutron/db/availability_zone/router.py)). Then open a PR which implements the feature + unit tests. Repeat for each addition to Gophercloud as needed.

Once all of that is done, we can revisit this. 

Please let me know if you have any questions.

stone:
Hi,

Thanks for taking the time to look at the PR, I have created three different issues at gophercloud/gophercloud. I will be away for a week but when I return I will update
the issues with corresponding PR with code and tests.

- gophercloud/gophercloud/issues/620 - ipv6_address_mode and ipv6_ra_mode for a subnet
- gophercloud/gophercloud/issues/621 - availability_zone_hints to a network 
- gophercloud/gophercloud/issues/622 - setting availability_zone_hints to a router

jtopjian:
All sounds good!

stone:
Thanks for helping out getting the changes merged into the upstream Gophercloud library. How do we continue? Should I create new PR's or can we work with what we got here?

jtopjian:
@stone No problem at all.

These are all small enhancements, but 3 separate PRs is still a good idea.

When doing each PR, you'll want to have two commits: one for the vendor update and one for the actual Terraform change. So the work flow will be something like this for the ipv6 changes:

```shell
$ govendor fetch github.com/gophercloud/gophercloud/openstack/networking/v2/subnets/...
$ git status
$ git add vendor
$ git commit -m "vendor: Updating Gophercloud"
$ <make changes to Terraform>
```

Let me know if you have any questions or need any help.

jtopjian:
I'm going to close this PR in favor of the others.

PR #163, opened 2017-11-24
Acc Tests: Update Acceptance Tests

jtopjian:
This commit cleans up some of the acceptance test conditionals,
including adding a conditional for environments which do not support
the OpenStack database service.

### DIFF
--- a/openstack/provider_test.go
+++ b/openstack/provider_test.go
@@ -13,6 +13,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


jtopjian:
I'll do a run of this in the CI environment tomorrow when the automated tests have finished.

jtopjian:
This is working as intended in the CI environment. There was a false-positive test failure which had nothing to do with this change. Going to merge.

PR #166, opened 2017-11-24
Docs: Add note about ports and instances

### LINKED ISSUE #153: openstack_compute_instance_v2 implicitly deletes ports
### Terraform Version
0.10.8

### Affected Resource(s)

there's a race condition when using openstack instances with ports:
when a change requiring a destroy/create on the resource, the api calls are too fast, 
and the neutron port is not in a available state when the create instance step is run

Here is the error that is raised: 
* openstack_compute_instance_v2.myinstance: Error trying to get network information from the Network API: Could not find any matching port for port c0f33b19-79bf-4dc2-a3e1-bf842544ba16

a second apply will raise no error.

i think the destroy func should check the state of the port when a port has been attached before exiting

jtopjian:
Fixes #153

### DIFF
--- a/website/docs/r/compute_instance_v2.html.markdown
+++ b/website/docs/r/compute_instance_v2.html.markdown
@@ -496,6 +496,11 @@ resource "openstack_compute_instance_v2" "foo" {
 Neutron Ports are a great feature and provide a lot of functionality. However,
 there are some notes to be aware of when mixing Instances and Ports:
 
+* In OpenStack environments prior to the Kilo release, deleting or recreating
+an Instance will cause the Instance's Port(s) to be deleted. One way of working
+around this is to taint any Port(s) used in Instances which are to be recreated.
+See [here](https://review.openstack.org/#/c/126309/) for further information.
+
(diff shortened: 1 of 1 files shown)


website/docs/r/compute_instance_v2.html.markdown
```diff
@@ -496,6 +496,11 @@ resource "openstack_compute_instance_v2" "foo" {
 Neutron Ports are a great feature and provide a lot of functionality. However,
 there are some notes to be aware of when mixing Instances and Ports:
 
+* In OpenStack environments prior to the Kilo release, deleting or recreating
```
yanndegat: prior to the Newton release, not kilo

jtopjian: It's listed in the release notes for Kilo: 

http://lists.openstack.org/pipermail/openstack-dev/2015-April/061449.html

https://wiki.openstack.org/wiki/ReleaseNotes/Kilo:

> Neutron ports are no longer deleted after your server is deleted, if you created them outside of Nova: 1153a46738fc3ffff98a1df9d94b5a55fdd58777

PR #167, opened 2017-11-24
Update README.md

niuzhenguo:
This replaces 'make bin' with 'make build' as there is no rule to
make target bin, which will confuse users.


jtopjian:
Thank you!

PR #168, opened 2017-11-24
Make a unique hash by adding the MAC address so that the same IP addr…

olivier-mauras:
…ess can be used multiple times

It may be needed to add the same ip_address in allow_address_pairs with a different mac_address.
The actual code can't as the hash is only generated on the ip_address. This fixes the issue by adding the hash of the mac_address used as well.


jtopjian:
@olivier-mauras Thank you for submitting this. This change makes sense to me.

Just a heads up that I'm going to wait until the next release to merge this. This is going to be a breaking change (since it is going to cause existing state files to change) and our current pending release has no breaking changes.

olivier-mauras:
Sure no problem :)

Issue #171, opened 2017-11-27
Some LB resource requests need a Neutron client


jrperritt:
In the PR to allow the use of Octavia, the LB V2 resources were changed to use either the Octavia or Neutron service client for requests. However, some of the calls use the `ports` package, which should use the Neutron client regardless of the LB client. Example:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_lb_loadbalancer_v2.go#L178

I'll send a fix for this.

jtopjian:
Fixed in #172

PR #172, opened 2017-11-27
use neutron client for port requests in lb v2 loadbalancer calls

jrperritt:
For #171 

There are 3 places in `resource_openstack_lb_loadbalancer_v2.go` where the client was changed to use the Neutron client (corresponding with the `ports` calls).
Additionally, `networkingClient` was renamed to `lbClient` when used for LB calls.


jtopjian:
Looks good. Thank you!

jrperritt:
@jtopjian thank you for reviewing so quickly

jtopjian:
No problem. I'm planning to request a release sometime this week. There's an incoming change to the new `openstack_db_instance_v1` resource that I'm waiting on first.

PR #178, opened 2017-12-01
Networking v2: Updates to openstack_networking_router_v2

jtopjian:
This commit makes two changes to the openstack_networking_router_v2
resource:

1. Deprecates external_gateway in favor of external_network_id. The
new name provides a better indication of what value should be set.

2. Adds support for external_fixed_ips. These can be specified by
the user as well as are set by the API when not specified.


jtopjian:
Thanks! Just like with #168, I'm going to wait until the next release to merge this. Our current release is all features and enhancements and I'd like to keep it that way :)

PR #179, opened 2017-12-03
openstack_db_database_v1 resource

eugenetaranov:
This PR adds openstack_db_database_v1 resource and addresses https://github.com/terraform-providers/terraform-provider-openstack/issues/141

### DIFF
--- a/openstack/resource_openstack_db_database_v1.go
+++ b/openstack/resource_openstack_db_database_v1.go
@@ -42,16 +42,6 @@ func resourceDatabaseDatabaseV1() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-			"charset": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-			},
-			"collate": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-			},
 		},
 	}
 }
@@ -67,9 +57,7 @@ func resourceDatabaseDatabaseV1Create(d *schema.ResourceData, meta interface{}) 
 
 	var dbs databases.BatchCreateOpts
 	dbs = append(dbs, databases.CreateOpts{
-		Name:    dbname,
-		CharSet: d.Get("charset").(string),
-		Collate: d.Get("collate").(string),
+		Name: dbname,
 	})
 
 	instance_id := d.Get("instance").(string)
@@ -117,8 +105,6 @@ func resourceDatabaseDatabaseV1Read(d *schema.ResourceData, meta interface{}) er
 	for _, v := range allDatabases {
 		if v.Name == dbname {
 			d.Set("name", v.Name)
-			d.Set("charset", v.CharSet)
-			d.Set("collate", v.Collate)
 			break
 		}
 	}
@@ -135,7 +121,6 @@ func resourceDatabaseDatabaseV1Delete(d *schema.ResourceData, meta interface{}) 
 	}
 
 	dbname := d.Get("name").(string)
-	// instance_id := d.Get("instance").(string)
 
 	pages, err := databases.List(databaseV1Client, d.Id()).AllPages()
 	allDatabases, err := databases.ExtractDBs(pages)
--- a/openstack/resource_openstack_db_database_v1.go
+++ b/openstack/resource_openstack_db_database_v1.go
@@ -172,7 +172,7 @@ func DatabaseDatabaseV1StateRefreshFunc(client *gophercloud.ServiceClient, insta
 
 		pages, err := databases.List(client, instance_id).AllPages()
 		if err != nil {
-			return nil, "", fmt.Errorf("Unable to retrieve databases, pageszzz: %s", err)
+			return nil, "", fmt.Errorf("Unable to retrieve databases, pages: %s", err)
 		}
 
 		allDatabases, err := databases.ExtractDBs(pages)
--- a/openstack/resource_openstack_db_database_v1.go
+++ b/openstack/resource_openstack_db_database_v1.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


jtopjian:
oops - sorry, commented on the wrong PR earlier :)

@eugenetaranov This looks good, though I noticed an oddity. It might just be my environment, though.

Trove is only returning the database name when querying the databases of an instance:

2017/12/06 04:26:18 [DEBUG] OpenStack Response Body: {
  "databases": [                                      
    {                                                 
      "name": "basic"                                 
    }                                                 
  ]                                                   
}                                                     

Because of this, collate and charset are being set to empty strings which is causing a conflict with the values given in the test fixture.

Can you confirm if these values are returning for you?

You can view extra log information by setting the OS_DEBUG environment variable to 1.

eugenetaranov:
@jtopjian checked, looks like trove does not send anything besides the db names indeed:

`
2017/12/06 14:43:28 [DEBUG] Openstack Response Code: 200
2017/12/06 14:43:28 [DEBUG] Openstack Response Headers:
Content-Length: 51
Date: Wed, 06 Dec 2017 11:43:28 GMT
Server: nginx/1.10.3 (Ubuntu)
2017/12/06 14:43:28 [DEBUG] OpenStack Response Body: {
  "databases": [
    {
      "name": "basic"
    },
    {
      "name": "sys"
    }
  ]
}
2017/12/06 14:43:28 Retrieved databases [{basic  } {sys  }]
`
Confirmed also with mitmproxy.

jtopjian:
@eugenetaranov OK, maybe that was a change in recent versions of Trove? Or it's an error in Gophercloud.

For now, I recommend not setting collate and charset via the API. During the Read call, just confirm the database name was returned.

eugenetaranov:
@jtopjian thank you, removed charset and collate for now

openstack/resource_openstack_db_database_v1.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf(
+			"Error waiting for database to become ready: %s", err)
+	}
+
+	// Store the ID now
+	d.SetId(instance_id)
```
jtopjian: The ID of the resource should be `instance_id/db_name`. If it's just `instance_id`, then this will limit a user from only be able to create one `openstack_db_database_v1` resource per database instance.

openstack/resource_openstack_db_database_v1.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func resourceDatabaseDatabaseV1Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	databaseV1Client, err := config.databaseV1Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating cloud database client: %s", err)
+	}
+
+	dbname := d.Get("name").(string)
```
jtopjian: If the ID is now `instance_id/db_name`, you can now do:

```go
pieces := strings.Split(db.Id(), "/", 2)
instanceID := pieces[0]
dbname := pieces[1]
```

openstack/resource_openstack_db_database_v1.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to retrieve databases, extract: %s", err)
+	}
+
+	for _, v := range allDatabases {
+		if v.Name == dbname {
+			d.Set("name", v.Name)
+			break
+		}
+	}
```
jtopjian: There should be some type of error check here if the dbname could not be found. Similar to what you're doing in `Delete` with `dbExists`. You should return an error if the database could not be found.

openstack/resource_openstack_db_database_v1.go
```diff
@@ -0,0 +1,176 @@
...
+
+	for _, v := range allDatabases {
+		if v.Name == dbname {
+			dbExists = true
+			break
+		}
+	}
+
+	if !dbExists {
+		log.Printf("Database %s was not found on instance %s", dbname, d.Id())
```
jtopjian: In addition to logging, also just `return nil` here. The database doesn't exist and we shouldn't try to do a `databases.Delete` call here.

openstack/resource_openstack_db_database_v1_test.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return fmt.Errorf("Unable to retrieve databases, extract: %s", err)
+		}
+
+		for _, v := range allDatabases {
+			if v.Name == db.Name {
+				return nil
+			}
+		}
+		return nil
```
jtopjian: I think this should return an error that the database could not be found.

eugenetaranov:
@jtopjian thank you, addressed all points, the only thing that I did differently is database ID: chose dot "." as a separator since "/" might cause issues when sent in http request.
In addition, found a bug: terraform.tfstate does not include openstack_db_database_v1 resourcem it is not generated. My test config is below, could you please advise what may cause such behavior.

```resource "openstack_db_instance_v1" "basic" {
  region        = "region"
  name          = "test"
  flavor_id     = "31792d21-c355-4587-9290-56c1ed0ca386"
  size          = 10
  network {
    uuid = "8ece1c9a-0ae9-4ffa-819e-cc5523bacc53"
  }

  datastore {
    version = "mysql-5.7"
    type    = "mysql"
  }
}

resource "openstack_db_database_v1" "test" {
  name     = "testdb18"
  instance = "${openstack_db_instance_v1.basic.id}"
}
```

animationzl:
recheck

jtopjian:
This looks good to me. @eugenetaranov Thank you for your work on this.

@animationzl let me know when I can merge this. Feel free to use this to test out the CI environment for however long you need.

jtopjian:
recheck trove

jtopjian:
LGTM - thank you @eugenetaranov  :smile:

PR #186, opened 2017-12-06
Docs: Adding database to sidebar

jtopjian:


jtopjian:
I could have swore I added this at one point, but maybe I intended to and forgot.

PR #189, opened 2017-12-10
LBaaSv2: Make subnet_id for members optional

### LINKED ISSUE #188: openstack_lb_member_v2 optional subnet_id
Hi,

I would like to propose for the subnet_id in the openstack_lb_member_v2 resource to be optional:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_lb_member_v2.go
```
"subnet_id": &schema.Schema{
  Type:     schema.TypeString,
  Required: true,
  ForceNew: true,
}
```
### Terraform Version
Terraform v0.11.1
+ provider.openstack v1.0.0

### Affected Resource(s)
- openstack_lb_member_v2

### Expected Behavior
subnet_id  in the  openstack_lb_member_v2 resource should be optional.

### Actual Behavior
subnet_id  in the openstack_lb_member_v2 resources is required but not saved by Opencontrail and therefor the resource is destroyed and created on every apply of the Terraform config. This will drop traffic to the members of the loadbalancer.

### Steps to Reproduce
1. `create a loadbalancer setup`
2. `terraform apply`
3. `terraform apply` 

### Important Factoids
Tested on Openstack Mikata with OpenContrail.

### References
Contrail not supporting subnet_id:
http://www.opencontrail.org/documentation/api/r3.2/contrail_openapi.html#loadbalancermembertype-label

Neutron models spec subnet_id for loadbalancer pool members should NULL-able: 
https://github.com/openstack/neutron-lbaas/blob/master/neutron_lbaas/db/loadbalancer/models.py#L88

jtopjian:
Fixes #188 

All signs point to this being correct. It's not required in Gophercloud. In Terraform, we're even checking to see if it's set as if it was an optional parameter:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_lb_member_v2.go#L111-L114

### DIFF
--- a/openstack/resource_openstack_lb_member_v2.go
+++ b/openstack/resource_openstack_lb_member_v2.go
@@ -72,7 +72,7 @@ func resourceMemberV2() *schema.Resource {
 
 			"subnet_id": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 
--- a/website/docs/r/lb_member_v2.html.markdown
+++ b/website/docs/r/lb_member_v2.html.markdown
@@ -31,7 +31,7 @@ The following arguments are supported:
 * `pool_id` - (Required) The id of the pool that this member will be
     assigned to.
 
-* `subnet_id` - (Required) The subnet in which to access the member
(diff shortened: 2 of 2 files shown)


jtopjian:
Thanks!

PR #193, opened 2017-12-12
Support for Stateless Address Autoconfiguration (SLAAC), Stateful DHCPv6, and Stateless DHCPv6 on a subnet.

stone:
Adds support for Stateless Address Autoconfiguration (SLAAC), Stateful DHCPv6, and Stateless DHCPv6 on a subnet.

 * i[pv6_ra_mode](https://developer.openstack.org/api-ref/network/v2/index.html#create-subnet)
 * [ipv6_address_mode](https://developer.openstack.org/api-ref/network/v2/index.html#create-subnet)

```
resource "openstack_networking_subnet_v2" "subnet_ipv6" {
  name                = "terraform-subnet-ipv6"
  network_id          = "${openstack_networking_network_v2.network.id}"
  cidr                = "ffff:c0:1f0:1002::/64"
  ipv6_address_mode   = "slaac"
  ipv6_ra_mode        = "slaac"
  ip_version          = 6
  enable_dhcp         = "true"
  gateway_ip          = "ffff:c0:1f0:1002::1"
}

### CODE REVIEW

File: openstack/data_source_openstack_networking_subnet_v2.go [hunk 1]
```diff
@@ -235,3 +259,12 @@ func dataSourceNetworkingSubnetV2Read(d *schema.ResourceData, meta interface{})
 
 	return nil
 }
+
+func dataSourceSubnetV2IPv6AddessMode(v interface{}, k string) (ws []string, errors []error) {
```
jtopjian: How about renaming this to `validateSubnetV2IPv6Mode` and move it to the `utils.go` file? That way both the data source and the resource can use it.

stone: Good catch, thanks! Forgot about the validating function in resource.. New commit pushed to PR.

jtopjian:
Looks good! Thank you :)

PR #196, opened 2017-12-13
 Support for availability_zone_hints for networks

stone:
This allows users to associate an availability zone with their network 

 * [availability_zone_hints](https://developer.openstack.org/api-ref/network/v2/index.html#networks) (missing in docs but supported by openstack cli and is about to be added to [webui](https://git.openstack.org/cgit/openstack/horizon/commit/?id=b9664a1bf139e6714aa4d507a68ce03d551779aa))

```
  name               = "terraform-ipv6"
  admin_state_up = "true"
  shared             = "false"
  availability_zone_hints = ["zone1", "zone2"]
}

```

### DIFF
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -57,7 +57,7 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Computed: true,
 			},
 			"availability_zone_hints": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -56,6 +56,12 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"availability_zone_hints": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


openstack/data_source_openstack_networking_network_v2.go
```diff
@@ -56,6 +56,12 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"availability_zone_hints": &schema.Schema{
+				Type:     schema.TypeSet,
```
jtopjian: I want to make sure `TypeSet` is the correct choice here.

If you specified `["zone2", "zone1"]` in Terraform and Neutron replied with `["zone1", "zone2"]`, then `TypeSet` is the correct choice. However, if Neutron respects the ordering you gave, then we can use a `TypeList` instead.

The benefit of using `TypeList` is that the user can then use numerical indexing to access the elements.

Unfortunately I don't have access to an environment that has more than one AZ to test with. Are you able to confirm?

openstack/resource_openstack_networking_network_v2.go
```diff
@@ -183,6 +192,7 @@ func resourceNetworkingNetworkV2Read(d *schema.ResourceData, meta interface{}) e
 	d.Set("shared", strconv.FormatBool(n.Shared))
 	d.Set("tenant_id", n.TenantID)
 	d.Set("region", GetRegion(d, config))
+	d.Set("availability_zone_hints", n.AvailabilityZoneHints)
```
jtopjian: Small change here: when setting a non-scalar type (Set or List), we need to check for an error. We've only recently begun to do this so there aren't many cases of it being done, so it's no fault this was missed.

```go
if err := d.Set("availability_zone_hints", n.AvailabilityZoneHints); err != nil {
  log.Printf("[DEBUG] unable to set availability_zone_hints: %s", err)
}
```

openstack/data_source_openstack_networking_network_v2.go
stone: I have done some empirical testing, and it looks like the ordering is kept, tried to look into the source for the [field](https://github.com/openstack/neutron/blob/4a128494ae4d7d9d0223dad136e858b7ead83122/neutron/objects/network.py#L190) and it looks like a [Python list](https://git.openstack.org/cgit/openstack/oslo.versionedobjects/tree/oslo_versionedobjects/fields.py#n1117) (and the [List "type"](https://git.openstack.org/cgit/openstack/oslo.versionedobjects/tree/oslo_versionedobjects/fields.py#n634)) which is ordered.

What I did to test was to set different ordering of zones, and then check with "openstack" cli output.

I will do some more testing, and if I don't find anything more I think TypeList is the way to go.

jtopjian: Sounds good - thank you!

openstack/resource_openstack_networking_network_v2.go
stone: Ok, np, added in latest push to this PR.

stone:
Hmm.. something didn't work as expected, looking into it..

jtopjian:
This looks good to me. Thank you for your work on this :)

PR #203, opened 2017-12-15
Allow setting availability_zone_hints on router.

stone:
This allows users to associate an availability zone with their
resources.

```
resource "openstack_networking_router_v2" "testrouter" {
    name                      = "testrouter"
    availability_zone_hints   = ["zone1", "zone3"]
}
```

When/if #196 is accepted, one improvement would be to break out  resourceNetworkingRouterkAvailabilityZoneHintsV2_ into util.go and 
use it from borth networks and router.

### DIFF
--- a/openstack/resource_openstack_networking_network_v2.go
+++ b/openstack/resource_openstack_networking_network_v2.go
@@ -110,7 +110,7 @@ func resourceNetworkingNetworkV2Create(d *schema.ResourceData, meta interface{})
 		networks.CreateOpts{
 			Name:                  d.Get("name").(string),
 			TenantID:              d.Get("tenant_id").(string),
-			AvailabilityZoneHints: resourceNetworkingNetworkAvailabilityZoneHintsV2(d),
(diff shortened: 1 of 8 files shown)


jtopjian:
> When/if #196 is accepted, one improvement would be to break out resourceNetworkingRouterkAvailabilityZoneHintsV2_ into util.go and use it from borth networks and router.

This sounds like a good idea to me. Let's name it `resourceNetworkingAvailabilityZoneHintsV2`.

jtopjian:
@stone This looks good.

Can you add a doc entry for the router `availability_zone_hints`? Since it has `ForceNew` set, the doc should mention  "Changing this creates a new router". If you could also amend the docs for the `openstack_networking_network_v2` resource (not data source) to say the same, that would save me a follow-up patch :)

Thank you for your help and work on these - it's really appreciated.

jtopjian:
Perfect. Thank you so much for your help and work on these!

PR #206, opened 2017-12-19
Handle errors of waiting for floating ip ready

niuzhenguo:
This PR adds error handler when waiting for floating ip ready.


jtopjian:
Looks good to me. Thank you!

PR #216, opened 2018-01-02
Set region if defined in clouds.yaml

### LINKED ISSUE #215: Region is not set correctly when cloud is only specified in provider
I was planning to only use the recently added `cloud` support in the provider. However it seems it crashes if `region` is not specified in the provider even if `region_name` is specified in `clouds.yaml`

### Terraform Version
```
Terraform v0.11.1
+ provider.openstack v1.1.0
```

### Affected Resource(s)
I have only tested with a few resources but I can guess it affects all resources which require region in some way.

```hcl
provider "openstack" {
  cloud  = "test"
}

data "openstack_networking_network_v2" "external" {
  name = "ext-net"
}
```

### Panic Output
https://gist.github.com/haxorof/42aefd817de74b51dfa4649bedf7d442

### Problem area
[config.go](terraform-provider-openstack/openstack/config.go) uses `gophercloud/utils` which handles `cloud.yaml`. However the in the case when `cloud` is used region will not be set in the Config struct which leads to the crash I see.  Currently the `region_name` is not returned by `clientconfig` so it seems some small thing is missing here so the Config struct gets the `region_name` from the cloud configuration.

### Expected Behavior
What should have happened?

```
> terraform refresh
Empty or non-existent state file.

Refresh will do nothing. Refresh does not error or return an erroneous
exit status because many automation scripts use refresh, plan, then apply
and may not have a state file yet for the first run.

data.openstack_networking_network_v2.external: Refreshing state...
```

### Actual Behavior
What actually happened?

```
> terraform refresh
Empty or non-existent state file.

Refresh will do nothing. Refresh does not error or return an erroneous
exit status because many automation scripts use refresh, plan, then apply
and may not have a state file yet for the first run.

data.openstack_networking_network_v2.external: Refreshing state...

Error: Error refreshing state: 1 error(s) occurred:

* data.openstack_networking_network_v2.external: 1 error(s) occurred:

* data.openstack_networking_network_v2.external: data.openstack_networking_network_v2.external: unexpected EOF


panic: runtime error: invalid memory address or nil pointer dereference
...

!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

Terraform crashed! This is always indicative of a bug within Terraform.
A crash log has been placed at "crash.log" relative to your current
working directory. It would be immensely helpful if you could please
report the crash with Terraform[1] so that we can fix this.

When reporting bugs, please include your terraform version. That
information is available on the first line of crash.log. You can also
get it by running 'terraform --version' on the command line.

[1]: https://github.com/hashicorp/terraform/issues

!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

### Steps to Reproduce
1. `terraform refresh`
2. CRASH!

### Workaround
Set `region` in provider.

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

* Services like keystone have several endpoints (one endpoint per region)

jtopjian:
This commit will set the region name if one is provided
in the clouds.yaml entry and the user did not specify
one in the provider configuration. If the user set one
in the provider, it will override what is set in
clouds.yaml.

Fixes #215

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -76,6 +76,15 @@ func (c *Config) LoadAndValidate() error {
 		if err != nil {
 			return err
 		}
+
+		cloud, err := clientconfig.GetCloudFromYAML(clientOpts)
+		if err != nil {
+			return err
+		}
+
+		if c.Region == "" && cloud.RegionName != "" {
+			c.Region = cloud.RegionName
+		}
 	} else {
 		ao = &gophercloud.AuthOptions{
 			DomainID:         c.DomainID,


jtopjian:
Thanks!

PR #219, opened 2018-01-03
usable examlpe of compute_flavor_v2

rremer:
Flavors are properly documented as configuring RAM in megabytes, but the example is not really useful (unless you have an image that can boot with only 8 MB).

I think the intent was to show 8096... but some value of at least 2048 MB should be copy/pasteable for the average user's image.


jtopjian:
@rremer Good catch, thank you :)

PR #221, opened 2018-01-04
Openstack db configuration instance

eugenetaranov:
This PR adds configuration to db instance addressing https://github.com/terraform-providers/terraform-provider-openstack/issues/141

### DIFF
--- a/openstack/resource_openstack_db_instance_v1.go
+++ b/openstack/resource_openstack_db_instance_v1.go
@@ -274,7 +274,6 @@ func resourceDatabaseInstanceV1Create(d *schema.ResourceData, meta interface{}) 
 	if configuration, ok := d.GetOk("configuration_id"); ok {
 		instances.AttachConfigurationGroup(databaseV1Client, instance.ID, configuration.(string))
 		log.Printf("Attaching configuration %v to the instance %v", configuration, instance.ID)
-		log.Printf("Restarting instance instance %v", instance.ID)
 	}
 
 	// Store the ID now
@@ -322,9 +321,6 @@ func resourceDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{}) er
 			instances.AttachConfigurationGroup(databaseV1Client, d.Id(), new.(string))
 			log.Printf("Attaching configuration %v to the instance %v", new, d.Id())
 		}
-
-		instances.Restart(databaseV1Client, d.Id())
-		log.Printf("Restarting instace %v", d.Id())
 	}
 
 	return resourceDatabaseInstanceV1Read(d, meta)
--- a/openstack/resource_openstack_db_instance_v1.go
+++ b/openstack/resource_openstack_db_instance_v1.go
@@ -272,7 +272,10 @@ func resourceDatabaseInstanceV1Create(d *schema.ResourceData, meta interface{}) 
 	}
 
 	if configuration, ok := d.GetOk("configuration_id"); ok {
-		instances.AttachConfigurationGroup(databaseV1Client, instance.ID, configuration.(string))
+		err := instances.AttachConfigurationGroup(databaseV1Client, instance.ID, configuration.(string)).ExtractErr()
+		if err != nil {
+			return err
+		}
 		log.Printf("Attaching configuration %v to the instance %v", configuration, instance.ID)
 	}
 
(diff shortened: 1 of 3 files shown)


openstack/resource_openstack_db_instance_v1.go
```diff
@@ -151,6 +152,12 @@ func resourceDatabaseInstanceV1() *schema.Resource {
 					},
 				},
 			},
+			"configuration": &schema.Schema{
```
jtopjian: The valid input for this would be the UUID of the configuration, right? If so, let's change this to `configuration_id`.

eugenetaranov: thanks, done

eugenetaranov:
@jtopjian added test, please review

jtopjian:
@eugenetaranov Quick update: I hope to review this one in the next day or two.

animationzl:
recheck

openstack/resource_openstack_db_instance_v1.go
```diff
@@ -292,6 +305,31 @@ func resourceDatabaseInstanceV1Read(d *schema.ResourceData, meta interface{}) er
...
+
+		instances.DetachConfigurationGroup(databaseV1Client, d.Id())
+		log.Printf("Detaching configuration %v from the instance %v", old, d.Id())
+
+		if new != "" {
+			instances.AttachConfigurationGroup(databaseV1Client, d.Id(), new.(string))
+			log.Printf("Attaching configuration %v to the instance %v", new, d.Id())
+		}
+
+		instances.Restart(databaseV1Client, d.Id())
```
jtopjian: @eugenetaranov I've been thinking about this for the past few days. Here are some thoughts I have:

* I can't find a lot of details, but what I've read says that _some_ configuration updates require a restart while others can be applied without a restart. Restarting upon every change might cause unneeded restarts.
*  Having someone apply this in production might cause some serious downtime. 

There are two ways we can handle this:

1. Keep this as it is and document a big warning. Some other Terraform resources already do this.
2. Remove this and have the user/administrator perform a restart on their own.

Thoughts?

eugenetaranov-travis: @jtopjian thanks for update!

I am voting for #2.
But maybe I can add an extra parameter to db instance resource "restart = true/false" with false as a default?

eugenetaranov: @jtopjian thanks for update!

I am voting for #2.
But maybe I can add an extra parameter to db instance resource "restart = true/false" with false as a default?

jtopjian: @eugenetaranov do you want to submit a patch which removes the restart?

eugenetaranov: yeah, will do @jtopjian , no worries

eugenetaranov:
@jtopjian removed instance restart, please take a look

jtopjian:
recheck trove

openstack/resource_openstack_db_instance_v1.go
```diff
@@ -292,6 +304,28 @@ func resourceDatabaseInstanceV1Read(d *schema.ResourceData, meta interface{}) er
...
+	config := meta.(*Config)
+	databaseV1Client, err := config.databaseV1Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating database client: %s", err)
+	}
+
+	if d.HasChange("configuration_id") {
+		old, new := d.GetChange("configuration_id")
+
+		instances.DetachConfigurationGroup(databaseV1Client, d.Id())
```
jtopjian: Similarly:

```go
err := instances.DetachConfigurationGroup(databaseV1Client, d.Id()).ExtractErr()
if err != nil {
  return err
}
```

openstack/resource_openstack_db_instance_v1.go
```diff
@@ -292,6 +304,28 @@ func resourceDatabaseInstanceV1Read(d *schema.ResourceData, meta interface{}) er
...
+		log.Printf("Detaching configuration %v from the instance %v", old, d.Id())
+
+		if new != "" {
+			instances.AttachConfigurationGroup(databaseV1Client, d.Id(), new.(string))
+			log.Printf("Attaching configuration %v to the instance %v", new, d.Id())
```
jtopjian: Similarly:

```go
err = instances.AttachConfigurationGroup(databaseV1Client, d.Id(), new.(string)).ExtractErr()
if err != nil {
  return err
}
```

openstack/resource_openstack_db_instance_v1.go
```diff
@@ -264,6 +271,11 @@ func resourceDatabaseInstanceV1Create(d *schema.ResourceData, meta interface{})
 			instance.ID, err)
 	}
 
+	if configuration, ok := d.GetOk("configuration_id"); ok {
+		instances.AttachConfigurationGroup(databaseV1Client, instance.ID, configuration.(string))
```
jtopjian: I think this should be:

```go
err := instances.AttachConfigurationGroup(databaseV1Client, instance.ID, configuration.(string)).ExtractErr()
if err != nil {
  return err
}
```

eugenetaranov:
thanks @jtopjian, updated

jtopjian:
LGTM! Nice work :)

PR #223, opened 2018-01-10
Use HasChange for openstack_fw_rule Update

niuzhenguo:
This changes to use HasChange to construct the update options instead
of sending the whole body even if the unchanged attributes.

PR #224, opened 2018-01-12
Use HasChange for openstack_fw_rule Update

jtopjian:
Amends #223 

@niuzhenguo Github is showing "unknown repository" for your PR so I've opened a new PR with some fixes. This patch seems to satisfy both of our use cases.

Can you test this and let me know if it works for you?

### DIFF
--- a/openstack/resource_openstack_fw_rule_v1.go
+++ b/openstack/resource_openstack_fw_rule_v1.go
@@ -171,46 +171,70 @@ func resourceFWRuleV1Update(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
 	}
 
+	var requireExtraFields bool
+
 	var updateOpts rules.UpdateOpts
 	if d.HasChange("name") {
 		name := d.Get("name").(string)
 		updateOpts.Name = &name
 	}
+
 	if d.HasChange("description") {
 		description := d.Get("description").(string)
 		updateOpts.Description = &description
 	}
+
 	if d.HasChange("protocol") {
 		protocol := d.Get("protocol").(string)
 		updateOpts.Protocol = &protocol
 	}
+
 	if d.HasChange("action") {
 		action := d.Get("action").(string)
 		updateOpts.Action = &action
 	}
+
 	if d.HasChange("ip_version") {
 		ipVersion := resourceFWRuleV1DetermineIPVersion(d.Get("ip_version").(int))
 		updateOpts.IPVersion = &ipVersion
 	}
+
 	if d.HasChange("source_ip_address") {
+		requireExtraFields = true
 		sourceIPAddress := d.Get("source_ip_address").(string)
 		updateOpts.SourceIPAddress = &sourceIPAddress
 	}
+
 	if d.HasChange("source_port") {
+		requireExtraFields = true
 		sourcePort := d.Get("source_port").(string)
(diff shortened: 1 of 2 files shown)


niuzhenguo:
@jtopjian Thanks very much! Yes, it fixes the error of my case when only updating rule name.

But I got another error when updating source_port/destination_port, and maybe that's the real problem.

I run "terraform apply" with "source_port" removed from the original terraform file and got the error:
```
"NeutronError": {
     "detail": "",
     "message": "Invalid input for source_port. Reason: Port '' is not a valid number.",
     "type": "HTTPBadRequest"
}
```

And seems the root cause is because gophercloud use a Request Body like
```
"firewall_rule": {
     "source_port": ""
}
```

instead of 
```
"firewall_rule": {
     "source_port": null
}
```

jtopjian:
@niuzhenguo To clarify: you are trying to remove the source port from a rule?

niuzhenguo:
@jtopjian Yes, just like chaning from https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_fw_rule_v1_test.go#L179-L190 to https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_fw_rule_v1_test.go#L171-L175

jtopjian:
The acceptance test suite for this is passing for me on Pike. However, I do agree with you that something is wrong with how Gophercloud is handling update options:

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/fwaas/rules/requests.go#L151-L163

The way it's configured will not allow "null" to ever be passed for any field -- only empty strings. In addition to that, Neutron seems to want some fields to always be present (for example protocol+port).

This is probably going to require updates to both Gophercloud and Terraform. Rather than continue to make guesses, we're going to have to dig into the Neutron code to figure out exactly how updates work. That could take some time.

niuzhenguo:
@jtopjian Yes, I'm ok with this PR and it's really fix my case. We can add another patch later, thanks!

jtopjian:
> Yes, I'm ok with this PR and it's really fix my case. We can add another patch later, thanks!

I realize that it will prolong the issue you have, and I apologize for that, but I would prefer not to merge this as-is. I'd rather spend some time to fully resolve the issue. I'm afraid that if I merge this PR as it is, it might actually cause more bugs.

niuzhenguo:
@jtopjian np, thanks for sticking with this!

jtopjian:
@niuzhenguo I apologize for the delay on this.

I found the area of neutron/fwaas code which handles updates:

https://github.com/openstack/neutron-fwaas/blob/master/neutron_fwaas/db/firewall/firewall_db.py#L496-L532

Given the conditionals applied in that function, I have tuned Terraform's update call to match them. Can you test this and let me now if you see any issues?

niuzhenguo:
@jtopjian thanks! it's ok on my tests with your new change but still can't remove source_port or destination_port even with protocol, as gophercloud use empty string instead of null.

jtopjian:
@niuzhenguo Thank you for testing. Removal of source and destination ports should now be possible. Can you verify when you have time?

niuzhenguo:
@jtopjian It works for me now, thanks very much!

jtopjian:
@niuzhenguo Awesome! Thank you for testing :)

@fatmcgav This is ready for review whenever you have time.

Issue #225, opened 2018-01-17
openstack_networking_subnet_v2: dns nameservers order is not preserved


yanndegat:
The order of the dns nameservers given as input is not preserved by the resource, thus
breaking the dns order consistency that user shall expect 
see: https://docs.openstack.org/neutron/pike/contributor/internals/dns_order.html

### Terraform Version
Terraform v0.11.1
+ provider.ignition v1.0.0
+ provider.local v1.1.0
+ provider.null v1.0.0
+ provider.openstack v1.1.0
+ provider.ovh v0.2.0
+ provider.template v1.0.0
+ provider.tls v1.0.1

### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking_subnet_v2:

Schema type TypeSet is to be blamed:
"dns_nameservers": &schema.Schema{
				Type:     schema.TypeSet,
				Optional: true,
				ForceNew: false,
				Elem:     &schema.Schema{Type: schema.TypeString},
				Set:      schema.HashString,
			},

hcl
```
 resource "openstack_networking_subnet_v2" "subnet" {

   name       = "testdns_order_subnet}"
   network_id = "${var.network_id}"
   cidr       = "${var.subnet_cidr}"
   ip_version = 4
   no_gateway = true

   dns_nameservers = ["10.0.16.4", "213.186.33.99"]
}
```

output:
neutron subnet-show b74ebf81-42fe-4add-bc7f-8ffdbbfd227a                                                                                             ✘ 1
neutron CLI is deprecated and will be removed in the future. Use openstack CLI instead.
+-------------------+------------------------------------------------------+
| Field             | Value                                                |
+-------------------+------------------------------------------------------+
| allocation_pools  | {"start": "10.0.48.2", "end": "10.0.63.254"}         |
| cidr              | 10.0.48.0/20                                         |
| created_at        | 2018-01-16T14:33:23Z                                 |
| description       |                                                      |
| dns_nameservers   | 213.186.33.99                                        |
|                   | 10.0.16.4                                            |
| enable_dhcp       | True                                                 |

Expected output
neutron subnet-show b74ebf81-42fe-4add-bc7f-8ffdbbfd227a                                                                                             ✘ 1
neutron CLI is deprecated and will be removed in the future. Use openstack CLI instead.
+-------------------+------------------------------------------------------+
| Field             | Value                                                |
+-------------------+------------------------------------------------------+
[...]

PR #226, opened 2018-01-17
 Subnet_v2: Fix dns nameservers order consistency #225 

yanndegat:


openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -356,14 +362,28 @@ func resourceSubnetAllocationPoolsV2(d *schema.ResourceData) []subnets.Allocatio
...
-	for i, raw := range rawDNSN.List() {
+	rawDNSN := d.Get("dns_nameservers").([]interface{})
+	dnsn := make([]string, len(rawDNSN))
+	for i, raw := range rawDNSN {
 		dnsn[i] = raw.(string)
 	}
 	return dnsn
 }
 
+func resourceSubnetDNSNameserversV2CheckIsSet(d *schema.ResourceData) error {
```
jtopjian: I think this would work better as a Validation function in the schema.

yanndegat: hi @jtopjian 
that's what i tried first
but i got a "Validate func are not supported on TypeList and TypeSet" error

yanndegat: https://github.com/hashicorp/terraform/blob/dd5882e5b26a1cfa0ba5f7d0b9b92a22834ce5d3/helper/schema/schema.go#L730

PR #234, opened 2018-02-06
Docs: additional_properties for instance scheduler hints

jtopjian:


jtopjian:
The test failure is ignorable - especially since this is only updating a doc.

Issue #242, opened 2018-02-14
Enhancement: add subnetpools management


ozerovandrei:
Hello.

It would be great if this provider will be able to manage Neutron Subnetpools.
Also subnet resource should has the subnetpool id in it's schema.

### Terraform Version
Terraform v0.11.3

### Affected Resource(s)
openstack_networking_subnet_v2
openstack_networking_subnetpool_v2

### Terraform Configuration Files
-

### Debug Output
-

### Panic Output
-

### Expected Behavior
Management of Neutron Subnetpools resources

### Actual Behavior
Nothing yet.

### Steps to Reproduce
-

### Important Factoids
-

### References
[General info](https://docs.openstack.org/neutron/pike/admin/config-subnet-pools.html)
[Neutron API](https://developer.openstack.org/api-ref/network/v2/index.html#subnet-pools-extension-subnetpools)
[Gophercloud#672](https://github.com/gophercloud/gophercloud/issues/672)

ozerovandrei:
I want to implement this. I've done the subnetpool stuff inside the Gophercloud so I'm familiar with it.
Also there is no subnet pool info in the subnet resource inside Gophercloud, so I guess it's also should be fixed on the Gophercloud's side.

ozerovandrei:
I've just created the issue to address subnetpool id absence in the subnet resource: [Gophercloud#763](https://github.com/gophercloud/gophercloud/issues/763)

ozerovandrei:
Implemented in #243 and #249.

PR #243, opened 2018-02-16
Networking v2: Add subnetpools support

ozerovandrei:
Add the support for Neutron subnetpools:
  - add the resourceNetworkingSubnetPoolV2
  - add the dataSourceNetworkingSubnetPoolV2
  - add acceptance tests and a documentation

For: #242

### DIFF
--- a/openstack/data_source_openstack_networking_subnetpool_v2.go
+++ b/openstack/data_source_openstack_networking_subnetpool_v2.go
@@ -31,12 +31,6 @@ func dataSourceNetworkingSubnetPoolV2() *schema.Resource {
 				Computed: true,
 				ForceNew: false,
 			},
-			"tenant_id": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
-				ForceNew: true,
-			},
 			"project_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
@@ -142,24 +136,8 @@ func dataSourceNetworkingSubnetPoolV2Read(d *schema.ResourceData, meta interface
 		listOpts.DefaultQuota = v.(int)
 	}
 
-	if v, ok := d.GetOk("tenant_id"); ok {
-		listOpts.TenantID = v.(string)
-	}
-
 	if v, ok := d.GetOk("project_id"); ok {
 		listOpts.ProjectID = v.(string)
-	}
-
-	if v, ok := d.GetOk("created_at"); ok {
-		listOpts.CreatedAt = v.(string)
-	}
-
-	if v, ok := d.GetOk("updated_at"); ok {
-		listOpts.UpdatedAt = v.(string)
-	}
-
-	if v, ok := d.GetOk("prefixes"); ok {
-		listOpts.Prefixes = v.([]string)
 	}
 
 	if v, ok := d.GetOk("default_prefixlen"); ok {
@@ -225,11 +203,9 @@ func dataSourceNetworkingSubnetPoolV2Read(d *schema.ResourceData, meta interface
 
 	d.Set("name", subnetPool.Name)
 	d.Set("default_quota", subnetPool.DefaultQuota)
-	d.Set("tenant_id", subnetPool.TenantID)
 	d.Set("project_id", subnetPool.ProjectID)
 	d.Set("created_at", subnetPool.CreatedAt)
 	d.Set("updated_at", subnetPool.UpdatedAt)
-	d.Set("prefixes", subnetPool.Prefixes)
 	d.Set("default_prefixlen", subnetPool.DefaultPrefixLen)
 	d.Set("min_prefixlen", subnetPool.MinPrefixLen)
 	d.Set("max_prefixlen", subnetPool.MaxPrefixLen)
@@ -240,6 +216,10 @@ func dataSourceNetworkingSubnetPoolV2Read(d *schema.ResourceData, meta interface
 	d.Set("description", subnetPool.Description)
 	d.Set("revision_number", subnetPool.RevisionNumber)
 	d.Set("region", GetRegion(d, config))
+
+	if err := d.Set("prefixes", subnetPool.Prefixes); err != nil {
+		log.Printf("[WARN] unable to set prefixes: %s", err)
+	}
 
 	return nil
 }
--- a/openstack/data_source_openstack_networking_subnetpool_v2.go
+++ b/openstack/data_source_openstack_networking_subnetpool_v2.go
@@ -157,19 +157,19 @@ func dataSourceNetworkingSubnetPoolV2Read(d *schema.ResourceData, meta interface
 	}
 
 	if v, ok := d.GetOk("ip_version"); ok {
-		listOpts.IPversion = v.(int)
+		listOpts.IPVersion = v.(int)
 	}
 
 	if v, ok := d.GetOk("shared"); ok {
-		listOpts.Shared = v.(bool)
+		listOpts.Shared = v.(*bool)
 	}
 
 	if v, ok := d.GetOk("description"); ok {
 		listOpts.Description = v.(string)
 	}
 
-	if v, ok := d.GetOk("is_default"); ok {
-		listOpts.IsDefault = v.(bool)
+	if v, ok := d.GetOk("default"); ok {
+		listOpts.IsDefault = v.(*bool)
 	}
 
 	if v, ok := d.GetOk("revision_number"); ok {
--- a/openstack/data_source_openstack_networking_subnetpool_v2.go
+++ b/openstack/data_source_openstack_networking_subnetpool_v2.go
@@ -39,19 +39,16 @@ func dataSourceNetworkingSubnetPoolV2() *schema.Resource {
 			},
 			"created_at": &schema.Schema{
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 				ForceNew: false,
 			},
 			"updated_at": &schema.Schema{
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 				ForceNew: false,
 			},
 			"prefixes": &schema.Schema{
 				Type:     schema.TypeSet,
-				Optional: true,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
@@ -114,7 +111,6 @@ func dataSourceNetworkingSubnetPoolV2() *schema.Resource {
 			},
 			"revision_number": &schema.Schema{
 				Type:     schema.TypeInt,
-				Optional: true,
 				Computed: true,
 				ForceNew: false,
 			},
@@ -161,19 +157,17 @@ func dataSourceNetworkingSubnetPoolV2Read(d *schema.ResourceData, meta interface
 	}
 
 	if v, ok := d.GetOk("shared"); ok {
-		listOpts.Shared = v.(*bool)
+		shared := v.(bool)
+		listOpts.Shared = &shared
 	}
 
 	if v, ok := d.GetOk("description"); ok {
 		listOpts.Description = v.(string)
 	}
 
-	if v, ok := d.GetOk("default"); ok {
-		listOpts.IsDefault = v.(*bool)
-	}
-
-	if v, ok := d.GetOk("revision_number"); ok {
-		listOpts.RevisionNumber = v.(int)
+	if v, ok := d.GetOk("is_default"); ok {
+		isDefault := v.(bool)
+		listOpts.IsDefault = &isDefault
 	}
 
 	pages, err := subnetpools.List(networkingClient, listOpts).AllPages()
--- a/openstack/data_source_openstack_networking_subnetpool_v2.go
+++ b/openstack/data_source_openstack_networking_subnetpool_v2.go
@@ -1,0 +1,245 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/subnetpools"
+)
+
+func dataSourceNetworkingSubnetPoolV2() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNetworkingSubnetPoolV2Read,
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"default_quota": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"tenant_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"created_at": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"updated_at": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"prefixes": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+				ForceNew: false,
+			},
+			"default_prefixlen": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"min_prefixlen": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"max_prefixlen": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"address_scope_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"ip_version": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(int)
+					if value != 4 && value != 6 {
+						errors = append(errors, fmt.Errorf(
+							"Only 4 and 6 are supported values for 'ip_version'"))
+					}
+					return
+				},
+			},
+			"shared": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"is_default": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"revision_number": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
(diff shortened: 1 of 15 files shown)


ozerovandrei:
recheck

jtopjian:
@ozerovandrei The build failure error is a little hard to find. You'll have to look in the raw output (`job-output.txt.gz` here: http://logs.openlabtesting.org/logs/43/243/d502eaf5e8d16dc2d2eef1f49b5d0be74ca510a7/check/terraform-provider-openstack-acceptance-test/3209b2f/)

```
* data.openstack_networking_subnetpool_v2.subnetpool_1: data.openstack_networking_subnetpool_v2.subnetpool_1: Your query returned more than one result. Please try a more specific search criteria
```

Searching for keyword "fail" will eventually lead to the error message.

If it is passing for you locally, then the issue is because the CI environment is creating a subnet that looks similar. It might take a few tries to get it correct. I had to do this with the Flavor acceptance tests a few weeks ago.

ozerovandrei:
@jtopjian yeah it's running smoothly in my environment.
It seems that it's failing on a datasource test query that searches for all subnetpools with `ip_version=4`.
I've just made less broad queries.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,245 @@
...
+				Computed: true,
+				ForceNew: true,
+			},
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"created_at": &schema.Schema{
```
jtopjian: This might be a difficult one for the user to use. I recommend either removing it (I can't see a user querying a subnet pool by a specific date) or we'll need to ensure the correct date format is given.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,245 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"updated_at": &schema.Schema{
```
jtopjian: Same as `created_at`.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,245 @@
...
+	log.Printf("[DEBUG] Retrieved Subnetpool %s: %+v", subnetPool.ID, subnetPool)
+	d.SetId(subnetPool.ID)
+
+	d.Set("name", subnetPool.Name)
+	d.Set("default_quota", subnetPool.DefaultQuota)
+	d.Set("tenant_id", subnetPool.TenantID)
+	d.Set("project_id", subnetPool.ProjectID)
+	d.Set("created_at", subnetPool.CreatedAt)
+	d.Set("updated_at", subnetPool.UpdatedAt)
+	d.Set("prefixes", subnetPool.Prefixes)
```
jtopjian: When setting a non-scalar attribute, we now need to check for an error:

```go
if err := d.set("prefixes", subnetPool.Prefixes); err != nil {
  log.Printf("[WARN] unable to set prefixes: %s", err)
}
```

openstack/data_source_openstack_networking_subnetpool_v2_test.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil
+	}
+}
+
+const testAccOpenStackNetworkingSubnetPoolV2DataSourceSubnetPool = `
+resource "openstack_networking_subnetpool_v2" "subnetpool_1" {
+	name = "subnetpool_1"
+	description = "terraform subnetpool acceptance test"
+
+	prefixes = ["10.10.0.0/16", "10.11.11.0/24"]
```
jtopjian: Thank you for specifically testing this to make sure the type cast works.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+				ForceNew: true,
+				Computed: true,
+			},
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"created_at": &schema.Schema{
```
jtopjian: Since  you can't set `created_at`, this should be a "read-only" / computed-only attribute. You can do this by setting it by removing the `Optional` field in the schema.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+				Computed: true,
+			},
+			"updated_at": &schema.Schema{
```
jtopjian: Same as `created_at`.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
+			"is_default": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
+			},
+			"revision_number": &schema.Schema{
```
jtopjian: This is another "read-only" / computed-only attribute.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+	log.Printf("[DEBUG] Waiting for Subnetpool (%s) to become available", s.ID)
+	stateConf := &resource.StateChangeConf{
+		Target:     []string{"ACTIVE"},
+		Refresh:    waitForSubnetPoolActive(networkingClient, s.ID),
+		Timeout:    d.Timeout(schema.TimeoutCreate),
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
```
jtopjian: We should check for an error here.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+
+	log.Printf("[DEBUG] Retrieved Subnetpool %s: %#v", d.Id(), s)
+
+	d.Set("name", s.Name)
+	d.Set("default_quota", s.DefaultQuota)
+	d.Set("tenant_id", s.TenantID)
+	d.Set("project_id", s.ProjectID)
+	d.Set("created_at", s.CreatedAt)
+	d.Set("updated_at", s.UpdatedAt)
+	d.Set("prefixes", s.Prefixes)
```
jtopjian: The same error check needs done here as in the data source.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,245 @@
...
+				Computed: true,
+				ForceNew: false,
+			},
+			"default_quota": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"tenant_id": &schema.Schema{
```
jtopjian: Although List can probably accept both `tenant_id` and `project_id`, let's standardize on just `project_id`, so let's remove this.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+			},
+			"default_quota": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: false,
+			},
+			"tenant_id": &schema.Schema{
```
jtopjian: Same as the data source, let's remove this and keep `project_id`.

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,362 @@
...
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error deleting OpenStack Neutron Subnetpool: %s", err)
+	}
+
+	d.SetId("")
```
jtopjian: Small nit: this is no longer needed in the `Delete` calls. Terraform automatically does it.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,245 @@
...
+	if v, ok := d.GetOk("created_at"); ok {
+		listOpts.CreatedAt = v.(string)
+	}
+
+	if v, ok := d.GetOk("updated_at"); ok {
+		listOpts.UpdatedAt = v.(string)
+	}
+
+	if v, ok := d.GetOk("prefixes"); ok {
+		listOpts.Prefixes = v.([]string)
```
jtopjian: To protect against a strange type cast error, I wonder if we should call `resourceSubnetPoolPrefixesV2` here.

ozerovandrei:
@jtopjian thank you for a review.
I've never contributed to terraform or it's providers so these comments are really helpful.

ozerovandrei:
@jtopjian I rechecked list query with `prefixes`. It seems that it's not working. At least in my environment. Neutron just throws 500 internal server error and writes to log something like that:
```
File ".../neutron/objects/subnetpool.py", line 70, in modify_fields_to_db
ERROR neutron.api.v2.resource     for prefix in result['prefixes']
ERROR neutron.api.v2.resource KeyError: 'id'
```
So I removed `prefixes` from the list options.

I added `project_id` to update opts and fixed all issues that you commented.

jtopjian:
@ozerovandrei 

> I rechecked list query with prefixes. It seems that it's not working.

That's strange.

Can you try testing with both `TF_LOG=DEBUG` and `OS_DEBUG=1` set? That will display the API requests and responses. I wonder if the `prefixes` field isn't being formatted correctly. 

There will be _a lot_ of output, so it's best to pipe all output (stderr and stdout) to a file for review.

ozerovandrei:
@jtopjian I got the same error with a simple curl query to list subnetpools filtered by prefixes in a `stable/newton` release.
I installed a simple devstack with `stable/queens` release and right now I'm getting a different error if I'm trying to make a query to list subnetpools filtered by prefixes.:
```
{"NotImplementedError": {"message": "in_() not yet supported for relationships.  For a simple many-to-one, use in_() against the set of foreign key values.", "type": "NotImplementedError", "detail": ""}}
```.

I guess that I need to remove `prefixes` from `subnetpools.ListOpts` in the Gophercloud then. Sorry, I've never tested query with that option after I implemented `List()` for subnetpools 😞 .

**Comment by jtopjian** [comment]
> I guess that I need to remove prefixes from subnetpools.ListOpts in the Gophercloud then. Sorry, I've never tested query with that option after I implemented List() for subnetpools 😞 .

It's no problem at all. There was probably something in the code that made it look as though a prefix query was possible or I would have never merged the code 😄 No harm done.

**Comment by jtopjian** [comment]
@ozerovandrei I've submitted https://github.com/gophercloud/gophercloud/pull/789 which addresses the changes needed :)

**Comment by ozerovandrei** [comment]
@jtopjian that's great. Thank you!

**Comment by ozerovandrei** [comment]
@jtopjian do I need to change something else? I've removed `created_at`, `updated_at` and `prefixes` from requests.

**Comment by jtopjian** [comment]
@ozerovandrei https://github.com/gophercloud/gophercloud/pull/789 was merged. If you haven't, can you do another vendor import just so the latest Gophercloud subnetpool package is bundled?

**Comment by ozerovandrei** [comment]
@jtopjian I haven't noticed that `` and `` became `*bool` types.
How to better deal with that? Is something like that is a good choice (saw that in `subnets`):
```
data_source_openstack_networking_subnetpool_v2:
...
if v, ok := d.GetOk("shared"); ok {
    shared := true
    listOpts.Shared = &shared
}

if v, ok := d.GetOk("not_shared"); ok {
    shared := false
    listOpts.Shared = &shared
}

if v, ok := d.GetOk("default"); ok {
    isDefault := false
    listOpts.IsDefault = &isDefault
}

if v, ok := d.GetOk("not_default"); ok {
    isDefault := true
    listOpts.IsDefault = &isDefault
}
...
```
(Maybe `private` instead of `not_shared` is better).

**Comment by jtopjian** [comment]
@ozerovandrei What's wrong with just having a single option of `shared` and passing the value (either `true` or `false`) to `ListOpts`?

**Comment by ozerovandrei** [comment]
@jtopjian nothig wrong. It was an error assumption. All fixed 😄

File: openstack/data_source_openstack_networking_subnetpool_v2.go [hunk 14]
```diff
@@ -0,0 +1,225 @@
...
+			},
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"created_at": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
jtopjian: This line needs to be removed to prevent the user from querying by `created_at`.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,225 @@
...
+				Computed: true,
+				ForceNew: false,
+			},
+			"updated_at": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
jtopjian: Same as `created_at`.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,225 @@
...
+				Computed: true,
+				ForceNew: false,
+			},
+			"prefixes": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
```
jtopjian: This needs removed, too.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if v, ok := d.GetOk("shared"); ok {
+		listOpts.Shared = v.(*bool)
+	}
+
+	if v, ok := d.GetOk("description"); ok {
+		listOpts.Description = v.(string)
+	}
+
+	if v, ok := d.GetOk("default"); ok {
```
jtopjian: Should this be `is_default`?

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,225 @@
...
+			},
+			"is_default": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+				ForceNew: false,
+			},
+			"revision_number": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
```
jtopjian: I'm not getting an error when querying with `revision_number`, but not getting any results, either. It's probably best to remove `Optional` and make it a computed-only field.

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if v, ok := d.GetOk("description"); ok {
+		listOpts.Description = v.(string)
+	}
+
+	if v, ok := d.GetOk("default"); ok {
+		listOpts.IsDefault = v.(*bool)
+	}
+
+	if v, ok := d.GetOk("revision_number"); ok {
```
jtopjian: This will need removed

openstack/data_source_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,225 @@
...
+	if v, ok := d.GetOk("address_scope_id"); ok {
+		listOpts.AddressScopeID = v.(string)
+	}
+
+	if v, ok := d.GetOk("ip_version"); ok {
+		listOpts.IPVersion = v.(int)
+	}
+
+	if v, ok := d.GetOk("shared"); ok {
+		listOpts.Shared = v.(*bool)
```
jtopjian: Does this work? If so, that's great. But if not, you'll have to do:

```go
shared := v.(bool)
listOpts.Shared = &shared
```

openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -0,0 +1,363 @@
...
+		updateOpts.AddressScopeID = &v
+	}
+
+	if d.HasChange("description") {
+		v := d.Get("description").(string)
+		updateOpts.Description = &v
+	}
+
+	if d.HasChange("is_default") {
+		isDefaultRaw := d.Get("is_default").(string)
```
jtopjian: You should be able to just do

```go
isDefault := d.Get("is_default").(bool)
```

There's really no need to do `GetOk` here, so `Get` is fine.

website/docs/d/networking_subnetpool_v2.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+* `ip_version` - The IP protocol version.
+
+* `shared` - (Optional) Whether this subnetpool is shared across all projects.
+
+* `description` - (Optional) The human-readable description for the subnetpool.
+
+* `is_default` - (Optional) Whether the subnetpool is default subnetpool or not.
+
+* `revision_number` - The revision number of the subnetpool.
```
jtopjian: This will need removed, too

website/docs/d/networking_subnetpool_v2.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+* `prefixes` - See Argument Reference above.
+* `default_prefixlen` - See Argument Reference above.
+* `min_prefixlen` - See Argument Reference above.
+* `max_prefixlen` - See Argument Reference above.
+* `address_scope_id` - See Argument Reference above.
+* `ip_version` -The IP protocol version.
+* `shared` - See Argument Reference above.
+* `description` - See Argument Reference above.
+* `is_default` - See Argument Reference above.
+* `revision_number` - See Argument Reference above.
```
jtopjian: And this will need updated with `The revision number of the subnetpool.`

website/docs/r/networking_subnetpool_v2.html.markdown
```diff
@@ -0,0 +1,106 @@
...
+# openstack\_networking\_subnetpool_v2
+
+Manages a V2 Neutron subnetpool resource within OpenStack.
+
+## Example Usage
+
+```hcl
+resource "openstack_networking_subnetpool_v2" "subnetpool_1" {
+  name = "my_ipv6_pool"
+  prefixes = ["fdf7:b13d:dead:beef::/64", "fd65:86cc:a334:39b7::/64"]
```
jtopjian: I think `ip_version` defaults to "4", so in order for the above example to be valid, you'll have to add `ip_version = 6` here.

openstack/resource_openstack_networking_subnetpool_v2_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	default_quota = 4
+
+	default_prefixlen = 25
+	min_prefixlen = 24
+	max_prefixlen = 30
+}
+`
+
+const testAccNetworkingV2SubnetPoolPrefixLenghts = `
```
jtopjian: This is a really good test to have - nice. There's a typo in the name, though: `Lengths`.

Maybe a third step should be added which updates some of the fields? That way it tests the update functionality.

openstack/data_source_openstack_networking_subnetpool_v2.go
ozerovandrei: Sorry for the bad testing. Again. Yeah, it's not working, will fix that. (Same in "is_default").

openstack/resource_openstack_networking_subnetpool_v2.go
ozerovandrei: I don't get it. There is no `GetOk` in resource_openstack_networking_subnetpool_v2.go.

jtopjian: Sorry - my comment was confusing. I was confirming that using `Get` is fine.

But the important change that needs done here is to cast to a `bool` rather than a `string`.

jtopjian:
@ozerovandrei This looks good! Thank you for your patience and work with this. These are really nice additions to Terraform :)

ozerovandrei:
@jtopjian thank you for such a thorough review 😄

lunderhage:
Would it be possible to get some example about how to create a subnet using a subnetpool? The documentation does not say much for now.

jtopjian:
@lunderhage I apologize for the delay -- I lost track of this.

We've just merged a doc fix for this here: https://github.com/terraform-providers/terraform-provider-openstack/pull/312

It'll appear on the site next week.

PR #248, opened 2018-02-20
[WIP] Extend identity resources & data sources

gonzolino:
This PR adds resources `openstack_identity_role_v3` and `openstack_identity_user_role_v3` and data sources `openstack_identity_project_v3`, `openstack_identity_role_v3` and `openstack_identity_user_v3` in an effort to improve support for the identity service in openstack.

It is still a WIP. Currently there are tests missing and a `openstack_identity_group_role_v3` resource (alternatively, I am thinking about removing `openstack_identity_user_role_v3` and replacing it with a resource `openstack_identity_role_assignment_v3` that handles role assignments for users and groups).

Also, I am hoping for some feedback from maintainers, since this is my first PR on a terraform provider and I suspect that I might have made some rookie mistakes in the code ;)

The `openstack_identity_user_role_v3` resource should fix #237.

PR #250, opened 2018-02-25
Add data source and resource for roles

gonzolino:
Adds `openstack_identity_role_v3` data source & resource.

Docs are still missing.

### DIFF
--- a/openstack/data_source_openstack_identity_role_v3.go
+++ b/openstack/data_source_openstack_identity_role_v3.go
@@ -1,0 +1,89 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceIdentityRoleV3() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceIdentityRoleV3Read,
+
+		Schema: map[string]*schema.Schema{
+			"domain_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+// dataSourceIdentityRoleV3Read performs the role lookup.
+func dataSourceIdentityRoleV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.identityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	listOpts := roles.ListOpts{
+		DomainID: d.Get("domain_id").(string),
+		Name:     d.Get("name").(string),
+	}
+
+	log.Printf("[DEBUG] List Options: %#v", listOpts)
+
+	var role roles.Role
+	allPages, err := roles.List(identityClient, listOpts).AllPages()
+	if err != nil {
+		return fmt.Errorf("Unable to query roles: %s", err)
+	}
+
+	allRoles, err := roles.ExtractRoles(allPages)
+	if err != nil {
+		return fmt.Errorf("Unable to retrieve roles: %s", err)
+	}
+
+	if len(allRoles) < 1 {
+		return fmt.Errorf("Your query returned no results. " +
+			"Please change your search criteria and try again.")
+	}
+
+	if len(allRoles) > 1 {
+		log.Printf("[DEBUG] Multiple results found: %#v", allRoles)
+		return fmt.Errorf("Your query returned more than one result. Please try a more " +
+			"specific search criteria, or set `most_recent` attribute to true.")
+	}
+	role = allRoles[0]
+
+	log.Printf("[DEBUG] Single Role found: %s", role.ID)
+	return dataSourceIdentityRoleV3Attributes(d, config, &role)
+}
+
+// dataSourceIdentityRoleV3Attributes populates the fields of an Role resource.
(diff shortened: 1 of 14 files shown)


jtopjian:
@gonzolino Thank you for the work you've put in with these resources :smile: 

I haven't done a review yet, but I noticed that the builds are failing. This is because the CI environment doesn't run as an "admin" user, so resources which require admin-level privileges need to be ignored (for now. that should be changed at some point).

There's an example of how to do that here:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_identity_user_v3_test.go#L22-L25

Because the acceptance tests will be ignored, I'll run the tests manually when they're ready for review.

Let me know if you have any questions or need any help :)

gonzolino:
@jtopjian Thanks for the hint with the admin only acceptance tests! I updated the patch, so the test should be ignored now.

The tests I added for the role resource & data source are rather simple at the moment. It would be nice if you could have a look at them in your review, and if you have ideas for additional tests just let me know :)

openstack/data_source_openstack_identity_role_v3_test.go
```diff
@@ -0,0 +1,50 @@
...
+			return fmt.Errorf("Role data source ID not set")
+		}
+
+		return nil
+	}
+}
+
+const testAccOpenStackIdentityV3RoleDataSource_basic = `
+data "openstack_identity_role_v3" "role_1" {
+    name = "Member"
```
jtopjian: Let's change this to `admin`. Sounds silly, but `Member` might not exist in a testing environment. For example, mine has `_member_`. `admin`, however, is definitely a standard name.

openstack/resource_openstack_identity_role_v3.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceIdentityRoleV3() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceIdentityRoleV3Create,
+		Read:   resourceIdentityRoleV3Read,
+		Update: resourceIdentityRoleV3Update,
+		Delete: resourceIdentityRoleV3Delete,
+		Importer: &schema.ResourceImporter{
```
jtopjian: In order to ensure `import` works correctly, can you add an import test? See the `import_*_test.go` files as an example.

openstack/resource_openstack_identity_role_v3.go
```diff
@@ -0,0 +1,118 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
```
jtopjian: We'll need to add the `region` attribute here. Just copy and paste it from another resource -- they're all the same configuration.

openstack/data_source_openstack_identity_role_v3.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceIdentityRoleV3() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceIdentityRoleV3Read,
+
+		Schema: map[string]*schema.Schema{
```
jtopjian: We'll need to add the region attribute here. Just copy and paste it from another resource -- they're all the same configuration.

openstack/resource_openstack_identity_role_v3_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
+)
+
+func TestAccIdentityV3Role_basic(t *testing.T) {
```
jtopjian: Can you add a second step here which updates the name?

gonzolino:
@jtopjian Thanks for your comments!

I tried to address all of them and it would be nice if you could have another look.

jtopjian:
@gonzolino I apologize for the late reply.

This looks really good. I tested it and it works well. It'll need to be rebased with master -- there was a PR merged the other day which clashes with `openstack/provider.go`. Let me know if you need help with this.

Once the merge conflict is resolved and docs are added, this is good to go. :)

gonzolino:
@jtopjian No worries :)
Rebase is done and docs are added. So the PR is ready for the next review round 👍

openstack/resource_openstack_identity_role_v3.go
```diff
@@ -0,0 +1,125 @@
...
+func resourceIdentityRoleV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.identityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	role, err := roles.Get(identityClient, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "project")
```
jtopjian: `s/project/role`

openstack/resource_openstack_identity_role_v3.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if hasChange {
+		_, err := roles.Update(identityClient, d.Id(), updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating OpenStack role: %s", err)
+		}
+	}
+
+	return resourceIdentityProjectV3Read(d, meta)
```
jtopjian: `s/resourceIdentityProjectV3Read/resourceIdentityRoleV3Read`

The Update step of the acceptance test was failing because of this - took me a while to spot it :smile:

jtopjian:
recheck

gonzolino:
@jtopjian Sorry for taking so long, I hope all your comments are addressed now :)

jtopjian:
@gonzolino Nice work! This looks good to me.

PR #251, opened 2018-02-25
Add project data source

gonzolino:
Adds data source `openstack_identity_project_v3`.

### DIFF
--- a/openstack/data_source_openstack_identity_project_v3.go
+++ b/openstack/data_source_openstack_identity_project_v3.go
@@ -1,0 +1,117 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceIdentityProjectV3() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceIdentityProjectV3Read,
+
+		Schema: map[string]*schema.Schema{
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"domain_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			"is_domain": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"parent_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+// dataSourceIdentityProjectV3Read performs the project lookup.
+func dataSourceIdentityProjectV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.identityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	enabled := d.Get("enabled").(bool)
+	isDomain := d.Get("is_domain").(bool)
+	listOpts := projects.ListOpts{
+		DomainID: d.Get("domain_id").(string),
+		Enabled:  &enabled,
+		IsDomain: &isDomain,
+		Name:     d.Get("name").(string),
+		ParentID: d.Get("parent_id").(string),
+	}
+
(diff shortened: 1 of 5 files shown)


openstack/data_source_openstack_identity_project_v3.go
```diff
@@ -0,0 +1,113 @@
...
+	return dataSourceIdentityProjectV3Attributes(d, &project)
+}
+
+// dataSourceIdentityProjectV3Attributes populates the fields of an Project resource.
+func dataSourceIdentityProjectV3Attributes(d *schema.ResourceData, project *projects.Project) error {
+	log.Printf("[DEBUG] openstack_identity_project_v3 details: %#v", project)
+
+	d.SetId(project.ID)
+	d.Set("is_domain", project.IsDomain)
+	d.Set("description", project.Description)
```
jtopjian: This will need to be added to the schema, too, just like the user `default_project_id`.

gonzolino: Done. I added the `Computed` option just like with `default_project_id`.

openstack/data_source_openstack_identity_project_v3_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccOpenStackIdentityV3ProjectDataSource_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3DataSourceID("data.openstack_identity_project_v3.project_1"),
+					resource.TestCheckResourceAttr(
+						"data.openstack_identity_project_v3.project_1", "name", "demo"),
+					resource.TestCheckResourceAttr(
+						"openstack_identity_project_v3.project_1", "description", ""),
```
jtopjian: This should be `data.openstack_identity_project_v3`. Same for the others below this :)

jtopjian:
@gonzolino I wanted to check with this to see if it was ready for another review or if you're still working on it?

If the former, this looks really good. I'm running into issues with the acceptance test again and I think I know how we can resolve it (and the future ones, too): 

since it's possible to create a project, the acceptance test can have two steps: the first step will create a random project and the second step will verify information about it. You can see an example of this with the network data source:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_networking_network_v2_test.go#L11-L31

Let me know if you have any problems implementing it - I'd be happy to either help or submit a separate patch for it. While the acceptance tests are important, I'd hate for you to spend more time on them than the actual data source :)

gonzolino:
@jtopjian Thanks for the review! Yes, all three PRs are currently ready for review and I am open for suggestions :)

Good idea with the creation of a random project for the acceptance test!
I'll try that out later. I might adapt the acceptance tests for my other two PRs in a similar way.

gonzolino:
@jtopjian I adapted the acceptance tests. Would be good to get some feedback :)
If it looks good, I will also adapt the acceptance tests in my other PRs.

openstack/data_source_openstack_identity_project_v3_test.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccOpenStackIdentityProjectV3DataSource_project(projectName, projectDescription),
+			},
+			resource.TestStep{
+				Config: testAccOpenStackIdentityProjectV3DataSource_basic(projectName, projectDescription),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3DataSourceID("data.openstack_identity_project_v3.project_1"),
```
jtopjian: Looks like this should be `testAccCheckIdentityV3ProjectDataSourceID`.

`testAccCheckIdentityV3DataSourceID` is from the Role data source -- we should probably update that function name to include the word "Role".

openstack/data_source_openstack_identity_project_v3_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+`, name, description)
+}
+
+func testAccOpenStackIdentityProjectV3DataSource_basic(name, description string) string {
+	return fmt.Sprintf(`
+	%s
+
+	data "openstack_identity_project_v3" "project_1" {
+      name = "%s"
```
jtopjian: This should be

```
name = "${openstack_identity_project_v3.project_1.name}"
```

This will create an explicit relationship between the resource and data source. Without it, the resource is destroyed first which invalidates the data source (no results found).

openstack/data_source_openstack_identity_project_v3_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+`, testAccOpenStackIdentityProjectV3DataSource_project(name, description), name)
```
jtopjian: With the above change in place, this line will be

```
, testAccOpenStackIdentityProjectV3DataSource_project(name, description))
```

jtopjian:
@gonzolino This looks good to me 😄 

Did you have any other questions or any further updates or is this ready to be merged?

gonzolino:
@jtopjian If you ay it looks good it can be merged :)

jtopjian:
Sounds good! Thank you for your work on this.

PR #252, opened 2018-02-25
Add user data source

gonzolino:
Adds data source `openstack_identity_user_v3`.

### DIFF
--- a/openstack/data_source_openstack_identity_user_v3.go
+++ b/openstack/data_source_openstack_identity_user_v3.go
@@ -1,0 +1,163 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/users"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceIdentityUserV3() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceIdentityUserV3Read,
+
+		Schema: map[string]*schema.Schema{
+			"default_project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"domain_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"enabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			"idp_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"password_expires_at": &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validatePasswordExpiresAtQuery,
+			},
+
+			"protocol_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"unique_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+// dataSourceIdentityUserV3Read performs the user lookup.
+func dataSourceIdentityUserV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.identityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	enabled := d.Get("enabled").(bool)
+	listOpts := users.ListOpts{
+		DomainID:          d.Get("domain_id").(string),
+		Enabled:           &enabled,
+		IdPID:             d.Get("idp_id").(string),
+		Name:              d.Get("name").(string),
+		PasswordExpiresAt: d.Get("password_expires_at").(string),
+		ProtocolID:        d.Get("protocol_id").(string),
+		UniqueID:          d.Get("unique_id").(string),
+	}
+
+	log.Printf("[DEBUG] List Options: %#v", listOpts)
+
+	var user users.User
+	allPages, err := users.List(identityClient, listOpts).AllPages()
+	if err != nil {
+		return fmt.Errorf("Unable to query users: %s", err)
+	}
+
+	allUsers, err := users.ExtractUsers(allPages)
+	if err != nil {
+		return fmt.Errorf("Unable to retrieve users: %s", err)
+	}
+
+	if len(allUsers) < 1 {
+		return fmt.Errorf("Your query returned no results. " +
+			"Please change your search criteria and try again.")
+	}
+
+	if len(allUsers) > 1 {
+		log.Printf("[DEBUG] Multiple results found: %#v", allUsers)
+		return fmt.Errorf("Your query returned more than one result")
+	}
+	user = allUsers[0]
+
+	log.Printf("[DEBUG] Single user found: %s", user.ID)
+	return dataSourceIdentityUserV3Attributes(d, &user)
+}
+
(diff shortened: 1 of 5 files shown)


openstack/data_source_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,124 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"password_expires_at": &schema.Schema{
```
jtopjian: This might be a tricky one to handle. In order for this to work, the user will have to input a valid time stamp formatted as a string. There's a chance the user will enter this incorrectly, so you'll need to do some internal checking that the string is formatted correctly.

Next, the result of `user.PasswordExpiresAt` is of `time.Time`. If Terraform doesn't automatically convert this to a string, you'll need to do it internally.

If in doubt, just omit `password_expires_at` altogether. It's always possible for someone to dedicate a PR just to this later on.

openstack/data_source_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,124 @@
...
+// dataSourceIdentityUserV3Attributes populates the fields of an User resource.
+func dataSourceIdentityUserV3Attributes(d *schema.ResourceData, user *users.User) error {
+	log.Printf("[DEBUG] openstack_identity_user_v3 details: %#v", user)
+
+	d.SetId(user.ID)
+	d.Set("default_project_id", user.DefaultProjectID)
+	d.Set("description", user.Description)
+	d.Set("domain_id", user.DomainID)
+	d.Set("enabled", user.Enabled)
+	d.Set("extra", user.Extra)
```
jtopjian: I've already tested the `extra` field with roles and it won't work. Let's remove this for now.

openstack/data_source_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,124 @@
...
+	log.Printf("[DEBUG] Single user found: %s", user.ID)
+	return dataSourceIdentityUserV3Attributes(d, &user)
+}
+
+// dataSourceIdentityUserV3Attributes populates the fields of an User resource.
+func dataSourceIdentityUserV3Attributes(d *schema.ResourceData, user *users.User) error {
+	log.Printf("[DEBUG] openstack_identity_user_v3 details: %#v", user)
+
+	d.SetId(user.ID)
+	d.Set("default_project_id", user.DefaultProjectID)
```
jtopjian: This is missing from the schema. Since it's not possible to use this in the `ListOpts`, this means it's a read-only/computed-only field, so you can define it as:

```go
			"default_project_id": &schema.Schema{
				Type:     schema.TypeString,
				Computed: true,
			},
```

Note how `Computed` is the only option.

openstack/data_source_openstack_identity_user_v3.go
gonzolino: Ok, I tried to add a validation check to ensure that the input matches the format explained in to API docs: https://developer.openstack.org/api-ref/identity/v3/#list-users
Also, I added a call to `user.PasswordExpiresAt.Format()` to convert `time.Time` to `string`.
Hope this works.

openstack/data_source_openstack_identity_user_v3.go
gonzolino: Done

openstack/data_source_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,164 @@
...
+	d.Set("description", user.Description)
+	d.Set("domain_id", user.DomainID)
+	d.Set("enabled", user.Enabled)
+	d.Set("name", user.Name)
+	d.Set("password_expires_at", user.PasswordExpiresAt.Format(time.RFC3339))
+
+	return nil
+}
+
+func validatePasswordExpiresAtQuery(passwordExpiresAt string) error {
```
jtopjian: If `password_expires_at` is not set (like in the acceptance tests), this function returns an error. Maybe it should only be checked if the string is not empty?

Also, would it be possible to use this as a [`ValidateFunc`](https://github.com/terraform-providers/terraform-provider-openstack/search?utf8=%E2%9C%93&q=validatefunc&type=) in the schema? That would catch errors earlier on.

website/docs/d/identity_user_v3.html.markdown
```diff
@@ -0,0 +1,58 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+		ProtocolID:        d.Get("protocol_id").(string),
+		UniqueID:          d.Get("unique_id").(string),
+
+- * `description` - (Optional) A description of the user.
+
+- * `default_project_id` - (Optional) The default project this user belongs to.
```
jtopjian: This line and the line above it should be shifted to the left.

The `d.Get` lines above should probably be removed :)

openstack/data_source_openstack_identity_user_v3_test.go
```diff
@@ -0,0 +1,109 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityUserV3DataSourceID("data.openstack_identity_user_v3.user_1"),
+					resource.TestCheckResourceAttr(
+						"data.openstack_identity_user_v3.user_1", "name", userName),
+					resource.TestCheckResourceAttr(
+						"data.openstack_identity_user_v3.user_1", "domain_id", "default"),
+					resource.TestCheckResourceAttr(
+						"data.openstack_identity_user_v3.user_1", "enabled", "true"),
+					resource.TestCheckResourceAttr(
+						"data.openstack_identity_user_v3.user_1", "region", "default"),
```
jtopjian: I'm getting an error when running the tests that `region` can't be found. I suggest either removing this check or add a `d.Set("region", GetRegion(d, config))`.

openstack/data_source_openstack_identity_user_v3.go
gonzolino: Thats cool, I didn't know about ValidateFunc.
I changed it to be used a a ValidateFunc. Hope it works!

website/docs/d/identity_user_v3.html.markdown
gonzolino: done

openstack/data_source_openstack_identity_user_v3_test.go
gonzolino: I removed the check.

openstack/data_source_openstack_identity_user_v3.go
```diff
@@ -0,0 +1,165 @@
...
+	d.Set("name", user.Name)
+	d.Set("password_expires_at", user.PasswordExpiresAt.Format(time.RFC3339))
+
+	return nil
+}
+
+// Ensure that password_expires_at query matches format explained in
+// https://developer.openstack.org/api-ref/identity/v3/#list-users
+func validatePasswordExpiresAtQuery(v interface{}, k string) (ws []string, errors []error) {
+	re, err := regexp.Compile(`(\w{2,3}):\d{4}-\{2}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z`)
```
jtopjian: I had to change the regex to this to work:

```go
regexp.Compile(`(\w{2,3}):\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z`)
```

to do a `strings.SplitN(v.(string), 2)` and then parse the second part as a time stamp?

gonzolino: Oh you are right, there was an error in my regex!
What do you mean with `strings.SplitN(v.(string), 2)`? Split the string at the ':' so that we get operator and timestamp and can validate the timestamp?

jtopjian: oops - sorry.

```go
strings.SplitN(v.(string), ":", 2)
```

Right, split at the first `:` so that instead of using a regex, you could convert the string to a timestamp and then format it to what the OpenStack API wants.

gonzolino: Sounds like a good idea :)
I'll update the PR when I'm back in the office tomorrow.

mrhillsman:
recheck

gonzolino:
recheck

jtopjian:
@gonzolino Yep, this is good to go! Very nice work on this -- thank you for your work and patience. :smile:

PR #254, opened 2018-02-26
Removed default values on tenant_id

### LINKED ISSUE #253: Conflict with environment variables in a data source
### Terraform Version
Terraform v0.11.3
+ provider.null v1.0.0
+ provider.openstack v1.2.0
+ provider.template v1.0.0

### Affected Resource(s)
data "openstack_networking_subnet_v2"

```hcl
data "openstack_networking_subnet_v2" "public_subnet" {
  subnet_id = "AN_EXISTING_PUBLIC_SUBNET_OUTSIDE_TENANT"
}
output "p" {
  value = "${data.openstack_networking_subnet_v2.public_subnet.*.id}"
}

```

### Debug Output
-

### Expected Behavior
It shouldn't fail to find this existing subnet.
### Actual Behavior
Because i have setup OS_TENANT_ID it's getting added as a filter even tough i'm not specifying it in the tf configuration. I believe the issue would be fixed if the multi values here where removed https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_networking_subnet_v2.go#L48-L57

I think shouldn't be considered a breaking change because in no documentation this behaviour is specified as the expected one.

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

xocasdashdash:
Closes #253

### DIFF
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -40,12 +40,8 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Optional: true,
 			},
 			"tenant_id": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				DefaultFunc: schema.MultiEnvDefaultFunc([]string{
-					"OS_TENANT_ID",
-					"OS_PROJECT_ID",
-				}, ""),
+				Type:        schema.TypeString,
+				Optional:    true,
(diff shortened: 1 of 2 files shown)


a2ron:
+1

jtopjian:
recheck

jtopjian:
LGTM - thank you for fixing this :)

Issue #99, opened 2017-09-05
Feature request: support the extra_dhcp_opts parameter for the port resource 


ghost:
### Affected Resource(s)
- openstack_networking_port_v2

### Actual Behavior
There is no extra_dhcp_opts in the schema of the resourceNetworkingPortV2.
As the Neutron API expects a list of maps for the extra_dhcp_opts parameter it is currently not possible to provide the value via the value_specs parameter (which only accepts strings for the parameter values)

### Expected Behavior
There is a way to provida a value for the extra_dhcp_opts parameter when creating or updating the port resource

fatmcgav:
@salikov-201511 At a quick glance, it looks like `extra_dhcp_opts` is also missing from the Gophercloud [port package](https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/ports). 

So support would need to be added to the Gophercloud library before we could add it into the `openstack` provider. 

If you feel up for it, PR's are welcome both here and upstream in Gophercloud.
Otherwise hopefully someone will be able to get to it soon...

salikov-201511:
@fatmcgav Thanks. Unfortunately I won't be able to produce a PR for this anytime soon as I have zero experience with Go language. Hopefully someone will be able to implement this.

jtopjian:
This was added in #258

Issue #246, opened 2018-02-19
Feature request: assign QoS policy to openstack_networking_port_v2


DSpeichert:
### Terraform Version
Terraform v0.11.3

### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking_port_v2
- openstack_network_qos_v2 (new if needed?)

### Expected Behavior
It would be great if it was possible to attach pre-existing QoS policies to a port.
Full management of QoS policies does not need to be added at first, those are typically created by admins and shared for use by regular users.

Simply adding `qos_policy_id` on `openstack_networking_port_v2` would be a good first step.
More advanced would be resolving QoS policy ID by its name which requires additional API call and possible implementing a new `openstack_network_qos_v2` resource type.

### Actual Behavior
Feature missing

### Important Factoids
- QoS policies don't seem to be supported in gophercloud :(
- QoS is an extension of Neutron, but available by default
- Only one QoS policy can be applied at a time to a port

### References
A typical request to do that is:
`PUT /v2.0/ports/c7b7c8e3-e32f-4715-96a6-da028abebed6`
```
{"port": {"qos_policy_id": "4c7998b6-f46b-4728-afd2-782618644769"}}
```

jtopjian:
@DSpeichert Thanks for the request. This is possible to do, but support for QoS will need added to the upstream Gophercloud library first.

ozerovandrei:
QoS progress will be tracked in #760

ozerovandrei:
This feature is implemented in #781 and will be available in `1.20.0` version.

PR #258, opened 2018-03-06
Networking V2: add DHCP opts to the port resource

ozerovandrei:
Add the extra DHCP opt field to the network port resource.
Provide tests and documentation.

For: #99.


jtopjian:
@ozerovandrei Thanks for putting this together. 

This one almost drove me insane :smile: 

If you've tried running the acceptance tests, you'll notice Terraform is crashing. You'll need to do a vendor update of the Gophercloud "ports" package to pull in the updates which allow for extensions to work. That took me around 2 hours to realize :smile: 

Some of the test fixtures need updating, too. For example, the "update" test only has the port resource. You'll also need to include the network and subnet resources, too, or else they get deleted in "step 2".

I also have some ideas for how the Create and Update calls here can be simplified so the "if" conditionals can be removed, but we can work on that later.

And just a heads up: I'm going to want to get #244 merged before this PR, so it's possible this PR will need a rebase once that's done.

ozerovandrei:
@jtopjian yeah I tried to run the tests before and I couldn't understand what is the problem. So I pushed my first set of changes to see if CI can run it.
Thank you for the explanation. All fixed now.
And sure it can wait for the #244.

jtopjian:
@ozerovandrei I noticed that the tests only ever append dhcp options and never remove any - have you tried testing the removal of a dhcp option?

jtopjian:
@ozerovandrei I dug into this a little and discovered that while the current code will allow the port to be updated with new dhcp values, it's not possible to remove existing values from a port.

I've opened a PR here to fix this:

https://github.com/gophercloud/gophercloud/pull/817

ozerovandrei:
@jtopjian I've never tried it. Thank you for discovering that issue.

ozerovandrei:
Tests failed because Devstack crashed for some reason in the installation phase:
```
[Call Trace]
./stack.sh:1451:run_phase
/opt/stack/new/devstack/functions-common:1705:run_plugins
/opt/stack/new/devstack/functions-common:1672:source
/opt/stack/new/manila/devstack/plugin.sh:1160:install_manila_tempest_plugin
/opt/stack/new/manila/devstack/plugin.sh:851:git_clone
/opt/stack/new/devstack/functions-common:534:die
[ERROR] /opt/stack/new/devstack/functions-common:534 Cloning not allowed in this configuration
Error on exit
```

ozerovandrei:
recheck

ozerovandrei:
It's the same error. You can check it in devstack-early.txt.gz

jtopjian:
@ozerovandrei Let's give it a day and retry the tests - might be fixed by then. If not, I will notify OpenLab.

ozerovandrei:
recheck

animationzl:
OpenLab failure is fixed now, for the reason you can refer to this issue
https://github.com/theopenlab/openlab-zuul-jobs/issues/95

Feel sorry for the late recover, will make OpenLab more stable and responsive.

ozerovandrei:
@jtopjian moving all Neutron extensions from a single port resource sounds like a good idea.
We will need to have more port-related stuff in future, for example - QoS: #246 .

Let's postpone this PR for later. Maybe you can start separating port extensions to different resources and after that I will refactor changes for this PR to also move them to a different resource?
I mean, I need some common example of how those extensions are going to look to apply the same schema for this PR.

jtopjian:
@ozerovandrei That sounds good. Let's leave this PR open for the time being. I'm not sure if splitting up the port resource will actually work.

ozerovandrei:
@jtopjian do you have any new ideas regarding port resource and whether it should be split among new resources?

In my opinion, it's better to create new resources and mark current allowed_address_pair as deprecated. We can also remove all deprecated stuff if we're going to release a next major release of the provider (which will include new server resource, hopefully).

jtopjian:
> In my opinion, it's better to create new resources and mark current allowed_address_pair as deprecated. 

Agreed. Let's proceed with exploring this.

> We can also remove all deprecated stuff if we're going to release a next major release of the provider (which will include new server resource, hopefully).

A new server resource (with a new name) will mean we won't have to do a major release. But you're right - we won't be able to remove the deprecated arguments (`allowed_address_pair` etc) until a major release. That's OK, though.

ozerovandrei:
recheck

ozerovandrei:
@jtopjian this is ready for a review.

I also added `extend`/`flatten` functions into separate file as discussed in the #456. Please let me know if I got this idea right.

jtopjian:
@ozerovandrei Instead of having a resource that contains all of the DHCP options for a port, I was thinking of having a resource that defined a single DHCP option for a port:

```hcl
resource "openstack_networking_port_extra_dhcp_option_v2" "port_1_option_a" {
  name = "optionA"
  value = "valueA"
  port_id = "${openstack_networking_port_v2.port_1.id}"
}
```

The main reason for doing this is to effectively eliminate the `TypeSet` and make an individual DHCP option a first-class resource in Terraform. I understand that this would cause manifests/configurations to be larger and more verbose, as well as create more API traffic, but it seems that most users always run up against limitations with `TypeSet` and `TypeList` and turning those types into separate resources _seems_ like the best solution (I might be wrong. In addition, Terraform 0.12 will allow users much more flexibility with Sets and Lists, so perhaps this design limitation is no longer an issue).

By having a separate resource that still contains the `extra_dhcp_opts` argument/attribute `TypeSet`, it creates a one-to-one mapping of the parent port resource and DHCP options resource. Unless I'm missing something, it will never be possible for a port to have more than one DHCP option resource -- I think if the user would try to define more than one, it would cause some errors.

If the idea is to keep DHCP options as a `TypeSet` but reduce the amount of code and complexity in the port resource (which is totally valid), then we can define all DHCP option stuff in `openstack/extra_dhcp_options_v2.go`, just like you've done, and integrate it into the port resource. 

Thoughts?

ozerovandrei:
Moving each DHCP option to a separate resource sounds hard because it will be very different from the OpenStack API model where each port has several DHCP options in the key value form.

I've started to think of implementation and I guess that it will be difficult because we need to manage deletion of options somehow. But we can't add/delete separate options for a single port by their names, we always need to send a single Update request: [UpdateExtraDHCPOpt](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/extradhcpopts/requests.go#L68).
So in case of deletion we will need to implement some additional helper methods that will read all options from a port and populate the new `UpdateExtraDHCPOpt` without the option that we want to delete. This looks much harder than working with a `TypeSet` 😅 

> it will never be possible for a port to have more than one DHCP option resource -- I think if the user would try to define more than one, it would cause some errors.

You're right, if user tries to create more than one DHCP options resources for the same port, he will have a problems with unclean Terraform state.

> If the idea is to keep DHCP options as a TypeSet but reduce the amount of code and complexity in the port resource (which is totally valid), then we can define all DHCP option stuff in openstack/extra_dhcp_options_v2.go, just like you've done, and integrate it into the port resource.

That sounds like a better variant. How should I proceed with this? I think in that case It's better to remove DHCP options resource and just move all helper methods to the `extra_dhcp_options_v2.go`?
So we will have DHCP options as port attribute?

jtopjian:
> But we can't add/delete separate options for a single port by their names, we always need to send a single Update request: UpdateExtraDHCPOpt.

You're right. I totally overlooked this when I first thought about a new resource for this.

> So in case of deletion we will need to implement some additional helper methods that will read all options from a port and populate the new UpdateExtraDHCPOpt without the option that we want to delete. This looks much harder than working with a TypeSet

Not only is it much harder, it's actually not possible :smiley: Unfortunately we can't pull information from other resources.

> I think in that case It's better to remove DHCP options resource and just move all helper methods to the extra_dhcp_options_v2.go? So we will have DHCP options as port attribute?

Yes, this looks like the best way to go. I was hoping that we would be able to simplify the "port" resource, but the best we can do is start refactoring the code in the port resource to make it more clean and maintainable.

Thank you for researching this. :)

ozerovandrei:
@jtopjian alright, it's ready.

ozerovandrei:
@jtopjian well, it seems that we can't use the following code:
```
dhcpOptsSet := &schema.Set{}
dhcpOptsSet.Add(map[string]interface{}{
    "ip_version": dhcpOpt.IPVersion,
    "opt_name":   dhcpOpt.OptName,
    "opt_value":  dhcpOpt.OptValue,
})
```

It panics because of nil-pointer reference while trying to `Add` new elements: [helper/schema/set.go#L211](https://github.com/hashicorp/terraform/blob/v0.11.10/helper/schema/set.go#L211)

This stuff only worked when we initialized empty set like this:
```
dhcpOptsSet := &schema.Set{
    F: networkingPortDHCPOptsV2HashSetFunc(),		
}
```

Tests from the `networking_port_v2_test.go` are broken too because of that.

jtopjian:
@ozerovandrei ahh interesting. In that case, just use a basic `map[string]interface{}` instead of a `schema.Set`.

ozerovandrei:
@jtopjian thank you for helping me with that!

About testing helper: do we need to implement it inside this repo or maybe `stretchr/testify/assert` will be enough?
We can use [ElementsMatch](https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch) to compare `[]extradhcpopts.CreateExtraDHCPOpt` and other same structures.

jtopjian:
> About testing helper: do we need to implement it inside this repo or maybe stretchr/testify/assert will be enough?

I think testify would be best. I recall someone else asking about testify in Slack. I can double-check if it would be OK to include it in this repo.

ozerovandrei:
@jtopjian apart from it, I think this PR is ready.

jtopjian:
@ozerovandrei yay! Really nice work on this. Thank you!

jtopjian:
@ozerovandrei Actually, upon hitting "merge" I realized it would be better to name the field singular: `extra_dhcp_opt` since each option is only a single option. I'm going to do a quick PR that updates this.

PR #265, opened 2018-03-13
Add role assignment resource

gonzolino:
Adds resource openstack_identity_role_assignment_v3.

### DIFF
--- a/openstack/resource_openstack_identity_role_assignment_v3_test.go
+++ b/openstack/resource_openstack_identity_role_assignment_v3_test.go
@@ -1,0 +1,155 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
+
+	"github.com/gophercloud/gophercloud/pagination"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/users"
+)
+
+func TestAccIdentityV3RoleAssignment_basic(t *testing.T) {
+	var role roles.Role
+	var user users.User
+	var project projects.Project
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckAdminOnly(t)
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckIdentityV3RoleAssignmentDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccIdentityV3RoleAssignment_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3RoleAssignmentExists("openstack_identity_role_assignment_v3.role_assignment_1", &role, &user, &project),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "project_id", &project.ID),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "user_id", &user.ID),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "role_id", &role.ID),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckIdentityV3RoleAssignmentDestroy(s *terraform.State) error {
+	config := testAccProvider.Meta().(*Config)
+	identityClient, err := config.identityV3Client(OS_REGION_NAME)
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "openstack_identity_role_assignment_v3" {
+			continue
+		}
+
+		_, err := roles.Get(identityClient, rs.Primary.ID).Extract()
+		if err == nil {
+			return fmt.Errorf("Role assignment still exists")
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckIdentityV3RoleAssignmentExists(n string, role *roles.Role, user *users.User, project *projects.Project) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No ID is set")
+		}
+
+		config := testAccProvider.Meta().(*Config)
+		identityClient, err := config.identityV3Client(OS_REGION_NAME)
+		if err != nil {
+			return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+		}
+
+		domainID, projectID, groupID, userID, roleID := extractRoleAssignmentID(rs.Primary.ID)
+
+		var opts roles.ListAssignmentsOpts
+		opts = roles.ListAssignmentsOpts{
+			GroupID:        groupID,
+			ScopeDomainID:  domainID,
+			ScopeProjectID: projectID,
+			UserID:         userID,
+		}
+
+		pager := roles.ListAssignments(identityClient, opts)
+		var assignment roles.RoleAssignment
+
+		err = pager.EachPage(func(page pagination.Page) (bool, error) {
+			assignmentList, err := roles.ExtractRoleAssignments(page)
+			if err != nil {
+				return false, err
+			}
+
+			for _, a := range assignmentList {
+				if a.Role.ID == roleID {
+					assignment = a
+					return false, nil
+				}
+			}
+
+			return true, nil
+		})
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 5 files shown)


openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+					testAccCheckIdentityV3RoleAssignmentExists("openstack_identity_role_assignment_v3.role_assignment_1", &role, &user, &project1),
+					resource.TestCheckResourceAttr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "project_id", project1.Name),
+					resource.TestCheckResourceAttr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "user_id", user.Name),
+					resource.TestCheckResourceAttr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "role_id", role.Name),
+				),
+			},
+			resource.TestStep{
```
jtopjian: Since all attributes are `ForceNew`, we really don't need to test updating.

openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+		identityClient, err := config.identityV3Client(OS_REGION_NAME)
+		if err != nil {
+			return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+		}
+
+		split := strings.Split(rs.Primary.ID, "/")
+		projectID, userID, roleID := split[1], split[3], split[4]
+
+		foundProject, err := projects.Get(identityClient, projectID).Extract()
+		if err == nil {
```
jtopjian:
```go
if err != nil
```

😄

openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+			return err
+		}
+		foundUser, err := users.Get(identityClient, userID).Extract()
+		if err == nil {
```
jtopjian: Same here

openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+			return err
+		}
+		foundRole, err := roles.Get(identityClient, roleID).Extract()
+		if err == nil {
```
jtopjian: Same here

openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+						"openstack_identity_role_assignment_v3.role_assignment_1", "user_id", user.Name),
+					resource.TestCheckResourceAttr(
+						"openstack_identity_role_assignment_v3.role_assignment_1", "role_id", role.Name),
+				),
+			},
+			resource.TestStep{
+				Config: testAccIdentityV3RoleAssignment_update,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3RoleAssignmentExists("openstack_identity_role_assignment_v3.role_assignment_1", &role, &user, &project2),
+					resource.TestCheckResourceAttr(
```
jtopjian: This needs to be `resource.TestCheckResourcePtrAttr` and the value to check needs to be `&project.ID`.

openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+		config := testAccProvider.Meta().(*Config)
+		identityClient, err := config.identityV3Client(OS_REGION_NAME)
+		if err != nil {
+			return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+		}
+
+		split := strings.Split(rs.Primary.ID, "/")
+		projectID, userID, roleID := split[1], split[3], split[4]
+
+		foundProject, err := projects.Get(identityClient, projectID).Extract()
```
jtopjian: Instead of checking for each resource individually, would it make more sense to do a `ListOpts` and ensure there's an assigned result?

openstack/resource_openstack_identity_role_assignment_v3_test.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil
+	}
+}
+
+const testAccIdentityV3RoleAssignment_basic = `
+resource "openstack_identity_role_assignment_v3" "role_assignment_1" {
+  user_id = "demo"
+  project_id = "demo"
+  role_id = "admin"
+}
```
jtopjian: We can do a more elaborate "full-stack" test:

```hcl
resource "openstack_identity_project_v3" "project_1" {
  name = "project_1"
}

resource "openstack_identity_user_v3" "user_1" {
  name = "user_1"
  default_project_id = "${openstack_identity_project_v3.project_1.id}"
}

resource "openstack_identity_role_v3" "role_1" {
  name = "role_1"
}

resource "openstack_identity_role_v3" "role_2" {
  name = "role_2"
}

resource "openstack_identity_role_assignment_v3" "role_assignment_1" {
  user_id = "${openstack_identity_user_v3.user_1.id}"
  project_id = "${openstack_identity_project_v3.project_1.id}"
  role_id = "${openstack_identity_role_v3.role_1.id}"
}
```

website/docs/r/identity_role_assignment_v3.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+DomainID:  domainID,
+		GroupID:   groupID,
+		ProjectID: projectID,
+		UserID:    userID,
```
jtopjian: These items should probably be removed :)

website/docs/r/identity_role_assignment_v3.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+Note: You _must_ have admin privileges in your OpenStack cloud to use
+this resource.
+
+## Example Usage
+
+```hcl
+resource "openstack_identity_role_assignment_v3" "role_1" {
+  user_id = "admin"
+  domain_id = "default"
+  role_id = "admin"
```
jtopjian: Maybe use the "full-stack" test fixture to provide a more verbose example here.

openstack/resource_openstack_identity_role_assignment_v3_test.go
gonzolino: thats true, removed the update test

openstack/resource_openstack_identity_role_assignment_v3_test.go
gonzolino: Ok, changed that bit (and the other two `resource.TestCheckResourceAttr` calls. Somehow I confused name and ID here 😛 

But one question: Is there any particular reason why I should use `resource.TestCheckResourceAttrPtr("openstack_identity_role_assignment_v3.role_assignment_1", "project_id", &project.ID)` instead of `resource.TestCheckResourceAttr("openstack_identity_role_assignment_v3.role_assignment_1", "project_id", project.ID)` (without the pointer)?

website/docs/r/identity_role_assignment_v3.html.markdown
gonzolino: done

website/docs/r/identity_role_assignment_v3.html.markdown
gonzolino: done

openstack/resource_openstack_identity_role_assignment_v3_test.go
gonzolino: done

openstack/resource_openstack_identity_role_assignment_v3_test.go
gonzolino: Ok that makes sense.
I still need to work on that though, since the acceptance tests are currently failing for me with this.

gonzolino:
@jtopjian Thanks for your comments and sorry for the delay, I was busy with other stuff the last days.

I updated my patches but there is still work left to do, since the acceptance tests are failing for me. I'll try to fix that tomorrow.

openstack/resource_openstack_identity_role_assignment_v3_test.go
jtopjian: The comment for the `TestCheckResourceAttrPtr` function explains it better than I can :)

https://github.com/hashicorp/terraform/blob/7d1f594f54ab741ff021b82b94d73d7dc2fb8651/helper/resource/testing.go#L997-L1004

It took me a while to find that function. I was trying to debug a test for an hour or two and wondering why the value was always `""`. To get around that, `TestCheckResourceAttrPtr` is used.

jtopjian:
@gonzolino Absolutely no problem at all -- I've been swamped with work myself. Thank you so much for the updates to this :)

I think the following should get the test to work:

```go
    p, err := projects.Get(identityClient, assignment.Scope.Project.ID).Extract()
    if err != nil {
      return fmt.Errorf("Project not found")
    }
    *project = *p

    u, err := users.Get(identityClient, assignment.User.ID).Extract()
    if err != nil {
      return fmt.Errorf("User not found")
    }
    *user = *u

    r, err := roles.Get(identityClient, assignment.Role.ID).Extract()
    if err != nil {
      return fmt.Errorf("Role not found")
    }
    *role = *r
```

gonzolino:
recheck

jtopjian:
@gonzolino This looks good to me. Were there any other changes you wanted to make or any other questions?

jtopjian:
Nope - this looks great! I was just double-checking :)

Thank you very much for your work on this!

PR #269, opened 2018-03-26
Filter image with  MemberStatus field

bewiwi:
To get ID of a shared image we need to filter with member_status = all

Documentation:
https://specs.openstack.org/openstack/glance-specs/specs/api/v2/sharing-image-api-v2.html#list-shared-images

```
visibility=shared&member_status=all: show all images shared with me regardless of my member status
```

It's a mandatory feature if image was shared with your tenant and you only know the name.

### DIFF
--- a/openstack/data_source_openstack_images_image_v2.go
+++ b/openstack/data_source_openstack_images_image_v2.go
@@ -34,6 +34,12 @@ func dataSourceImagesImageV2() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"member_status": {
+				Type:     schema.TypeString,
+				Optional: true,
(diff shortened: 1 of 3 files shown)


jtopjian:
@bewiwi Thanks for adding this! This looks good to me.

Would you be able to add this new field to the argument reference here: https://raw.githubusercontent.com/terraform-providers/terraform-provider-openstack/master/website/docs/d/images_image_v2.html.markdown ?

That'll save me an extra PR. :)

bewiwi:
Updated with doc

jtopjian:
LGTM - thank you again :)

PR #270, opened 2018-03-26
[WIP] Vpnaas: ipsecpolicy resource

simonre:
Part 1 of a series that addresses #2 

@jtopjian could you maybe take a quick look at this and tell me if there are any obvious issues?
 I still need to add documentation and the policy lifetime needs to be added to the Create, Read and Update operations. Did I miss anything else?


openstack/provider.go
```diff
@@ -212,6 +212,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_networking_subnetpool_v2":        resourceNetworkingSubnetPoolV2(),
 			"openstack_objectstorage_container_v1":      resourceObjectStorageContainerV1(),
 			"openstack_objectstorage_object_v1":         resourceObjectStorageObjectV1(),
+			"openstack_ipsec_policy_v2":                 resourceIPSecPolicyV2(),
```
jtopjian: Do you have any objections to calling this `openstack_vpnaas_ipsec_policy_v2`?

openstack/resource_openstack_ipsec_policy_v2.go
```diff
@@ -0,0 +1,323 @@
...
+		return err
+	}
+
+	log.Printf("[DEBUG] IPSec policy created: %#v", policy)
+
+	d.SetId(policy.ID)
+
+	return resourceIPSecPolicyV2Read(d, meta)
+}
+func IPSecPolicyV2GetTransformProtocol(trp string) ipsecpolicies.TransformProtocol {
```
jtopjian: nit: the order of the functions should be:

* Create
* Read
* Update
* Delete
* everything else

openstack/resource_openstack_ipsec_policy_v2.go
```diff
@@ -0,0 +1,323 @@
...
+		opts.PFS = d.Get("pfs").(ipsecpolicies.PFS)
+	}
+
+	if d.HasChange("encapsulation_mode") {
+		opts.EncapsulationMode = d.Get("encapsulation_mode").(ipsecpolicies.EncapsulationMode)
+	}
+
+	log.Printf("[DEBUG] Updating IPSec policy with id %s: %#v", d.Id(), opts)
+
+	err = ipsecpolicies.Update(networkingClient, d.Id(), opts).Err
```
jtopjian: It's more common to do:

```go
_, err := ipsecpolicies.Update(...).Extract()
```

Additionally, this update call should be contained in an `if` conditional:

```go
if hasChange {
...
}
```

While `Update` _should_ only be called when something has changed, adding the `hasChange` part is a second layer of protection.

openstack/resource_openstack_ipsec_policy_v2.go
```diff
@@ -0,0 +1,323 @@
...
+func resourceIPSecPolicyV2Update(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	opts := ipsecpolicies.UpdateOpts{}
+
+	if d.HasChange("name") {
```
jtopjian: When doing a series of `HasChange` checks like this, you could do:

```go
var hasChange bool

if d.HasChange("...") {
  hasChange = true
  opts.Foo = x
}
```

openstack/resource_openstack_ipsec_policy_v2.go
```diff
@@ -0,0 +1,323 @@
...
+		MapValueSpecs(d),
+	}
+
+	log.Printf("[DEBUG] Create IPSec policy: %#v", opts)
+
+	policy, err := ipsecpolicies.Create(networkingClient, opts).Extract()
+	if err != nil {
+		return err
+	}
+
```
jtopjian: Should a `WaitForState` be added here?

openstack/resource_openstack_ipsec_policy_v2_test.go
```diff
@@ -0,0 +1,156 @@
...
+		config := testAccProvider.Meta().(*Config)
+		networkingClient, err := config.networkingV2Client(OS_REGION_NAME)
+		if err != nil {
+			return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+		}
+
+		var found *ipsecpolicies.Policy
+		for i := 0; i < 5; i++ {
+			// IPSec policy creation is asynchronous. Retry some times
+			// if we get a 404 error. Fail on any other error.
```
jtopjian: Doing a `WaitForState` during create should remove the need for this.

openstack/resource_openstack_ipsec_policy_v2_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+`
+
+const testAccIPSecPolicyV2_timeout = `
+resource "openstack_ipsec_policy_v2" "policy_1" {
+  timeouts {
+    create = "5m"
+  }
+}
```
jtopjian: This can probably be safely removed. The other timeout tests were all added when Terraform first supported `timeouts` and it was more important to explicitly do a test. We can now be quite confident that `timeouts` works as intended :)

simonre:
@jtopjian I think this is ready for review. I tried to fix all the points you mentioned.

openstack/resource_openstack_vpnaas_ipsec_policy_v2.go
```diff
@@ -0,0 +1,411 @@
...
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	encapsulationMode := IPSecPolicyV2GetEncapsulationMode(d.Get("encapsulation_mode").(string))
+	authAlgorithm := IPSecPolicyV2GetAuthAlgorithm(d.Get("auth_algorithm").(string))
+	encryptionAlgorithm := IPSecPolicyV2GetEncryptionAlgorithm(d.Get("encryption_algorithm").(string))
+	pfs := IPSecPolicyV2GetPFS(d.Get("pfs").(string))
+	transformProtocol := IPSecPolicyV2GetTransformProtocol(d.Get("transform_protocol").(string))
+	lifetime := IPSecPolicyV2GetLifetimeCreateOpts(d.Get("lifetime").(*schema.Set))
+	log.Printf("LIFETIME: %+v", lifetime)
```
jtopjian: nit: we should probably remove this line.

openstack/resource_openstack_vpnaas_ipsec_policy_v2.go
```diff
@@ -0,0 +1,411 @@
...
+	d.Set("region", GetRegion(d, config))
+
+	// Set the lifetime
+	var lifetimeMap map[string]interface{}
+	lifetimeMap = make(map[string]interface{})
+	lifetimeMap["units"] = policy.Lifetime.Units
+	lifetimeMap["value"] = policy.Lifetime.Value
+	var lifetime []map[string]interface{}
+	lifetime = append(lifetime, lifetimeMap)
+	d.Set("lifetime", &lifetime)
```
jtopjian: Something we only recently started doing was checking for an error when setting non "scalar" attributes:

```go
if err := d.Set("lifetime", &lifetime); err != nil {
  log.Printf("[WARN] unable to set lifetime")
  // or return err if it should be a hard error.
}
```

openstack/resource_openstack_vpnaas_ipsec_policy_v2.go
```diff
@@ -0,0 +1,411 @@
...
+	if d.HasChange("lifetime") {
+		lifetime := IPSecPolicyV2GetLifetimeUpdateOpts(d.Get("lifetime").(*schema.Set))
+		opts.Lifetime = &lifetime
+		hasChange = true
+	}
+
+	log.Printf("[DEBUG] Updating IPSec policy with id %s: %#v", d.Id(), opts)
+
+	if hasChange {
+		_, err = ipsecpolicies.Update(networkingClient, d.Id(), opts).Extract()
```
jtopjian: Should there be a wait for state after updating?

openstack/resource_openstack_vpnaas_ipsec_policy_v2.go
```diff
@@ -0,0 +1,411 @@
...
+	return func() (interface{}, string, error) {
+		policy, err := ipsecpolicies.Get(networkingClient, id).Extract()
+		if err != nil {
+			return "", "PENDING_CREATE", nil
+		}
+		return policy, "ACTIVE", nil
+	}
+}
+
+func IPSecPolicyV2GetTransformProtocol(trp string) ipsecpolicies.TransformProtocol {
```
jtopjian: nit: these function names should be something like:

`resourceIPSecPolicyV2TransformProtocol`

openstack/resource_openstack_vpnaas_ipsec_policy_v2_test.go
```diff
@@ -0,0 +1,180 @@
...
+			return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+		}
+
+		found, err := ipsecpolicies.Get(networkingClient, rs.Primary.ID).Extract()
+		switch {
+		case name != found.Name:
+			err = fmt.Errorf("Expected name <%s>, but found <%s>", name, found.Name)
+		case description != found.Description:
+			err = fmt.Errorf("Expected description <%s>, but found <%s>", description, found.Description)
+		}
```
jtopjian: This is really good to check if the attributes match! 

An alternative way of doing it can be seen here:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_identity_role_assignment_v3_test.go#L33-L39

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_identity_role_assignment_v3_test.go#L120

This way, you're returning the full found resource as a pointer and can compare each attribute individually in the test block.

simonre: Do you want me to do it this way to in this and future commits? I'm not quite clear on that.

jtopjian: I apologize for being unclear. Yes, it's best to use the method I described since that's how other tests are set up.

simonre: No worries. I'll change it then.

jtopjian:
@simonre One thing I forgot to make note off: you'll need to add a new section to

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/openstack.erb

with the VPN resources and then add list items for each one.

jtopjian:
@simonre This looks good to go! Thank you :)

I'm going to hold off on merging this as we're going to have a new release in the next few days (tomorrow, possibly). The VPN resources will begin getting merged into the release after that.

PR #283, opened 2018-04-13
Networking v2: Add disable_snat to router resource

jtopjian:
This commit adds the disable_snat argument to the openstack_networking_router_v2
resource.

enable_snat is retrieved using d.GetOk because we want to only set it
if the user explicitly enabled it in their Terraform configuration.
However, a value of "false" is not able to be read using this method.

Therefore, we need to add a new argument to reverse the logic and allow
a value of false to be sent in the API request.


jtopjian:
recheck

PR #290, opened 2018-04-17
Authentication Update

### LINKED ISSUE #275: OS_DOMAIN_* values incorrectly used
This provider currently does not source the domain values in openstack correctly and then pass them along to gophercloud.

The values as they exists in the official libraries are:
```
# Who the user is:
OS_USERNAME
OS_USER_ID
# what domain the user is in:
OS_USER_DOMAIN_NAME
OS_USER_DOMAIN_ID

# what the project scope is:
OS_PROJECT_NAME
OS_PROJECT_ID
# what the domain for the project is
OS_PROJECT_DOMAIN_NAME
OS_PROJECT_DOMAIN_ID

# what the domain scope is (mutually exclusive with project scope)
OS_DOMAIN_NAME
OS_DOMAIN_ID
```

While on the other hand what this terraform provider does is boil all those different values down to just domain_id and domain_name (this is wrong):
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider.go#L82-L103

Gophercloud supports all these values correctly itself, the terraform provider just isn't correctly passing them along.

Domain scope isn't useful to terraform, so that's safe to ignore (OS_DOMAIN_ID, OS_DOMAIN_NAME).

### Expected Behavior
Terraform should be able to correctly handle domain values when talking to OpenStack and use them correctly through gophercloud.

### Actual Behavior
As per:
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider.go#L82-L103
This provider is entirely ignoring the unique meaning of each domain value and simply grouping them all under domain_id and domain_name. Gophercloud sort of then handles this rather than assuming they are meant to domain scope it seems, but I'm not sure on exactly what happens there.

### Steps to Reproduce
I haven't tried to reproduce it, the code is clear as to what happens, but attempting to auth with a user in one domain, to a project in another domain is one way to test this (roles can be assigned for a user in a project that is in a different domain). Or even making sure multiple users of the same name in different domains are correctly used when domain values are provided.

### References
see discussion in: 
terraform-providers/terraform-provider-openstack#271

and continuation in:
gophercloud/gophercloud#252

jtopjian:
This commit fixes authentication issues seen with user and project
domain scoping. It does this by adding new authentication
arguments which can be used to specify user and project domain
names and IDs.

Fixes #275


jtopjian:
The test failure is transient. TCP connection failure during the test. Will try a recheck anyway.

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
@fatmcgav This one is ready for review, too.

The last commit is the actual changes to Terraform. Namely, it is using the new `clientconfig` package to handle authentication configuration. The `clientconfig` implementation was a result of some in-depth work I did here: https://github.com/gophercloud/utils/pull/27

In general, `clientconfig` can be used for all user-facing Gophercloud-based apps. 

If you want to verify this yourself outside of testing, compile a binary and try using any form of authentication. I've tested this across three different clouds, each with different auth configurations, and all supported forms (hard-coded auth settings, env variables, `clouds.yaml` entries) have worked.

I'm quite confident that this is good to go and will resolve all previous authentication issues that have been reported.

Let me know if you have any questions :)

PR #294, opened 2018-04-18
Networking v2: Make Subnet CIDR Optional

### LINKED ISSUE #282: Subnet shouldn't require cidr when using a pool
The CLI allows creation of a subnet with the CIDR coming from a subnet pool rather than being explicitly declared, but CIDR is required in TF config even when a pool is specified. 

https://docs.openstack.org/ocata/networking-guide/config-subnet-pools.html

> openstack subnet create --subnet-pool demo-subnetpool4 --network demo-network1 demo-subnet1

### Terraform Version

```bash
$ terraform -v
Terraform v0.11.7
+ provider.openstack v1.3.0
```

### Affected Resource(s)
- opc_instance
- opc_storage_volume

```hcl
provider "openstack" {
  auth_url    = "${var.openstack_auth_url}"
  insecure    = "${var.openstack_insecure}"
  domain_name = "${var.project_domain_name}"
  tenant_name = "${var.project_name}"
}

data "openstack_networking_network_v2" "tenant_network" {
  name = "${var.tenant_network_name}"
}

data "openstack_networking_subnetpool_v2" "pool23" {
  name = "${var.openstack_subnet_pool}"
}

resource "openstack_networking_subnet_v2" "tenant_subnet" {
  name            = "tenant-subnet-${var.project_name}-01"
  tenant_id       = "${var.project_id}"
  network_id      = "${data.openstack_networking_network_v2.tenant_network.id}"
  subnetpool_id   = "${data.openstack_networking_subnetpool_v2.pool23.id}"
  ip_version = 4
}
```

variables.tf:

```hcl
variable "openstack_auth_url" {
  description = "The endpoint url to connect to OpenStack."
  default     = "https://openstack.example.com:5000/v3"
}

variable "openstack_insecure" {
  type        = "string"
  description = "Accept self signed SSL certificates?"
  default     = "true"
}

variable "tenant_network_name" {
  description = "Openstack Network Name"
  default     = "tenant-network-demo"
}

variable "openstack_subnet_pool" {
  description = "Name of the subnet pool"
  default     = "VIOINT_Tenant_pool23"
}

variable "project_domain" {
  description = "The domain of the Project."
  default     = "default"
}

variable "project_name" {
  description = "The name of the Project."
}

variable "project_id" {
  description = "The ID of the Project"
}
```

### Expected Behavior

Subnet should should be created with next available CIDR from the pool provided.

### Actual Behavior

```
Error: openstack_networking_subnet_v2.tenant_subnet: "cidr": required field is not set
```

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Running VMWare Integrated OpenStack (ViO) with Ocata OpenStack release.

jtopjian:
This commit makes the subnet CIDR optional so that a subnet can
be created with a subnet pool and have the pool automatically
delegate a CIDR.

The CIDR argument/attribute is now also computed to reflect the
CIDR supplied by the pool.

Fixes #282

### DIFF
--- a/openstack/resource_openstack_networking_secgroup_rule_v2.go
+++ b/openstack/resource_openstack_networking_secgroup_rule_v2.go
@@ -116,7 +116,7 @@ func resourceNetworkingSecGroupRuleV2Create(d *schema.ResourceData, meta interfa
 		PortRangeMax:   d.Get("port_range_max").(int),
 		RemoteGroupID:  d.Get("remote_group_id").(string),
 		RemoteIPPrefix: d.Get("remote_ip_prefix").(string),
-		TenantID:       d.Get("tenant_id").(string),
+		ProjectID:      d.Get("tenant_id").(string),
 	}
 
 	if v, ok := d.GetOk("direction"); ok {
--- a/openstack/resource_openstack_networking_subnet_v2.go
+++ b/openstack/resource_openstack_networking_subnet_v2.go
@@ -41,7 +41,8 @@ func resourceNetworkingSubnetV2() *schema.Resource {
 			},
 			"cidr": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Computed: true,
 				ForceNew: true,
 			},
 			"name": &schema.Schema{
@@ -163,7 +164,6 @@ func resourceNetworkingSubnetV2Create(d *schema.ResourceData, meta interface{}) 
 	createOpts := SubnetCreateOpts{
 		subnets.CreateOpts{
 			NetworkID:       d.Get("network_id").(string),
-			CIDR:            d.Get("cidr").(string),
 			Name:            d.Get("name").(string),
 			TenantID:        d.Get("tenant_id").(string),
 			IPv6AddressMode: d.Get("ipv6_address_mode").(string),
@@ -175,6 +175,11 @@ func resourceNetworkingSubnetV2Create(d *schema.ResourceData, meta interface{}) 
 			EnableDHCP:      nil,
 		},
 		MapValueSpecs(d),
+	}
+
+	if v, ok := d.GetOk("cidr"); ok {
+		cidr := v.(string)
+		createOpts.CIDR = cidr
 	}
 
 	if v, ok := d.GetOk("gateway_ip"); ok {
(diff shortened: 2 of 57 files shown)


louis-gounot:
Hello,

Documentation has not been updated accordingly.

CIDR is now optionnal. If you want dynamic CIDR from subnetpool with custom `prefixlen`, you MUST provide `prefixlen` as a `value_specs` attribute (`prefixlen` is an integer). If not provided, default prefixlen from subnetpool will be used (`default_prefixlen` attibute of `openstack_networking_subnetpool_v2`). As a side effect (due to default OpenStack beahaviour, not provider) default gateway (if allocated) will be first IP of allocated CIDR.

For those wanting to use dynamic CIDR allocation from subnetpool, you can use something like this :

```hcl
resource "openstack_networking_subnetpool_v2" "subnetpool" {
  name = "subnetpool"
  ip_version = 4
  max_prefixlen = 28
  min_prefixlen = 24
  default_prefixlen = 26
  is_default = "false"
  shared = "1"
  prefixes = [
    "172.16.0.0/16"
  ]
}

  name = "network"
}

resource "openstack_networking_subnet_v2" "subnet" {
  name = "subnet"
  network_id = "${openstack_networking_network_v2.network.id}"
  subnetpool_id = "${openstack_networking_subnetpool_v2.subnetpool.id}"
  value_specs {
    prefixlen = "25"
  }
}
```

By the way, many thanks to @jtopjian for this addition !!! 👍

jtopjian:
@louis-gounot Thanks for the feedback!

I think there are two issues here:

1. Documentation was not updated in this PR. I've opened #332 to address this.

2. Inability to specify `prefixlen`. This deserves more detail:

What you've outlined works, but it's not the way we'd like to see users implementing this. `vendor_specs` is reserved for vendor-specific options _but_ it can also be used to inject key/value pairs that are not currently supported upstream. This way of using `vendor_specs` is kept undocumented because we don't want to see users relying on it.

The best solution is to get `prefixlen` added to the upstream Gophercloud library and then to add a first-class field of `prefixlen` to the `openstack_networking_subnet_v2` resource. I've opened https://github.com/terraform-providers/terraform-provider-openstack/issues/333 to track this.

Let me know if my assumptions are correct here :)

PR #297, opened 2018-04-25
LBaaS v2: Lower Delays for Load Balancer Resources

### LINKED ISSUE #295: Unnecessary waiting prior to initial state checks
When creating openstack_lb_member_v2 resources, terraform-provider-openstack will [do the following:](https://github.com/terraform-providers/terraform-provider-openstack/blob/ad0d5ed871c3f2209fd20abf75fc87931a44cc51/openstack/resource_openstack_lb_member_v2.go#L94-L149)

1. [Wait for the load balancer state to be ACTIVE](https://github.com/terraform-providers/terraform-provider-openstack/blob/ad0d5ed871c3f2209fd20abf75fc87931a44cc51/openstack/resource_openstack_lb_member_v2.go#L121)
2. [Create the pool member](https://github.com/terraform-providers/terraform-provider-openstack/blob/ad0d5ed871c3f2209fd20abf75fc87931a44cc51/openstack/resource_openstack_lb_member_v2.go#L129)
3. [Wait for the load balancer state to be ACTIVE, again](https://github.com/terraform-providers/terraform-provider-openstack/blob/ad0d5ed871c3f2209fd20abf75fc87931a44cc51/openstack/resource_openstack_lb_member_v2.go#L141)

This works fine, functionally.  However, there's one small improvement I would like to suggest.

Each call to `waitForLBV2viaPool` (to wait for the load balancer state to become ACTIVE) will [wait 5 seconds](https://github.com/terraform-providers/terraform-provider-openstack/blob/570573a5f080c97b09f8b21dbc849a6293db7847/openstack/lb_v2_shared.go#L185) before it checks the state for the first time. This is fine for step 3 since changes were just made in step 2, _but_ it wastes 5 seconds for step 1. For step 1, the load balancer is generally _already_ ACTIVE, since no changes have been performed on it yet.  So there's no reason to wait 5 seconds before the first state check in step 1

I would like to suggest changing the behavior so that it does NOT wait 5 seconds prior to the state check in step 1. To be clear, it still should wait for the load balancer state to be ACTIVE, it just should not wait 5 seconds before the first check. This change would allow pool member creation to be performed 5 seconds faster. While this might not seem like much, it definitely adds up over time when creating lots pool members. This will be particularly helpful for me, since I deal with an LBaaS provider that has some serious concurrency issues that we have to workaround by reducing the terraform parallelism when creating/updating/deleting pool members.

The same change could be made for updating/deleting pool members (and possibly other resources as well).

### Terraform Version

terraform 0.11.7
terraform-provider-openstack 1.3.0

### Affected Resource(s)

openstack_lb_member_v2

Any config with an openstack_lb_member_v2 resource.

### Debug Output
https://gist.github.com/philsttr/89124e17d68136de01b19ddd17e0d918

### Panic Output
none

### Expected Behavior
Don't wait 5 seconds before checking loadbalancer state before creating/updating/deleting pool members

### Actual Behavior
Waiting 5 seconds before checking loadbalancer state before creating/updating/deleting pool members

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
none, other than we're using a hardware LBaaS provider that has some problems handling lots of concurrent requests.
We therefore have to reduce terraform parallelism to 1 when creating/updating/deleting pool members.
This makes 5 additional seconds per pool member modification particularly painful.

### References
none

jtopjian:
This commit lowers the delays for the load balancer resources
while waiting for a certain state. For the LoadBalancer resource
in particular, the delay is set to 0.

Fixes #295 

I tested setting all delays to 0 and everything worked, but I want to be cautious here. #295 specifically requested the load balancer delay to be set to 0, so I've implemented that. For everything else, I've lowered it to 1 second.

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -23,7 +23,7 @@ func waitForLBV2Listener(networkingClient *gophercloud.ServiceClient, id string,
 		Pending:    pending,
 		Refresh:    resourceLBV2ListenerRefreshFunc(networkingClient, id),
 		Timeout:    timeout,
-		Delay:      5 * time.Second,
+		Delay:      1 * time.Second,
 		MinTimeout: 1 * time.Second,
 	}
 
@@ -63,7 +63,7 @@ func waitForLBV2LoadBalancer(networkingClient *gophercloud.ServiceClient, id str
 		Pending:    pending,
 		Refresh:    resourceLBV2LoadBalancerRefreshFunc(networkingClient, id),
 		Timeout:    timeout,
-		Delay:      5 * time.Second,
+		Delay:      0,
 		MinTimeout: 1 * time.Second,
 	}
 
@@ -102,7 +102,7 @@ func waitForLBV2Member(networkingClient *gophercloud.ServiceClient, poolID, memb
 		Pending:    pending,
 		Refresh:    resourceLBV2MemberRefreshFunc(networkingClient, poolID, memberID),
 		Timeout:    timeout,
-		Delay:      5 * time.Second,
+		Delay:      1 * time.Second,
 		MinTimeout: 1 * time.Second,
 	}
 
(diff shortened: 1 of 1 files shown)


jtopjian:
recheck lbaas

jtopjian:
recheck lbaas

jtopjian:
I've tested this already outside of OpenLab and it works. 

I'm now seeing if the OpenLab lbaas checks are working and it looks like there might be some problem. OpenLab isn't making it to the point of running the tests, so any failure here is ignorable.

PR #300, opened 2018-04-27
Vpnaas: VPN service resource

simonre:
Part 2 of a series that addresses #2

### DIFF
--- a/openstack/resource_openstack_vpnaas_service_v2.go
+++ b/openstack/resource_openstack_vpnaas_service_v2.go
@@ -124,6 +124,10 @@ func resourceServiceV2Create(d *schema.ResourceData, meta interface{}) error {
 	}
 	_, err = stateConf.WaitForState()
 
+	if err != nil {
+		return err
+	}
+
 	log.Printf("[DEBUG] Service created: %#v", service)
 
 	d.SetId(service.ID)
@@ -210,6 +214,10 @@ func resourceServiceV2Update(d *schema.ResourceData, meta interface{}) error {
 			MinTimeout: 2 * time.Second,
 		}
 		_, err = stateConf.WaitForState()
+
+		if err != nil {
+			return err
+		}
 
 		log.Printf("[DEBUG] Updated service with id %s", d.Id())
 	}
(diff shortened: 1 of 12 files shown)


jtopjian:
> Unless I'm misinterpreting the openlab logs I don't think that this is a problem with my code (?)

That's correct -- you can ignore the results of OpenLab acceptance tests with these VPNaaS PRs.

openstack/resource_openstack_vpnaas_service_v2.go
```diff
@@ -0,0 +1,286 @@
...
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"NOT_CREATED"},
+		Target:     []string{"PENDING_CREATE"},
+		Refresh:    waitForServiceCreation(networkingClient, service.ID),
+		Timeout:    d.Timeout(schema.TimeoutCreate),
+		Delay:      0,
+		MinTimeout: 2 * time.Second,
+	}
+	_, err = stateConf.WaitForState()
```
jtopjian:
```go
if err != nil {
  return err
}
```

openstack/resource_openstack_vpnaas_service_v2.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+		stateConf := &resource.StateChangeConf{
+			Pending:    []string{"PENDING_UPDATE"},
+			Target:     []string{"UPDATED"},
+			Refresh:    waitForServiceUpdate(networkingClient, service.ID),
+			Timeout:    d.Timeout(schema.TimeoutCreate),
+			Delay:      0,
+			MinTimeout: 2 * time.Second,
+		}
+		_, err = stateConf.WaitForState()
```
jtopjian:
```go
if err != nil {
  return err
}
```

simonre:
@jtopjian Apart from the test issue this is ready for review again.

jtopjian:
@simonre I have no idea why I misread your notes twice.

Since the problem is only with updating, then I'm OK with omitting update tests. If we can test creation, then that's good enough and will exercise a good amount of code in this resource.

simonre:
@jtopjian I removed the update test and rebased the code.

PR #310, opened 2018-05-02
Compute v2: Add wait_until_associated to floating IP association

### LINKED ISSUE #88: FloatingIP trys to attach even if it is already attached and results in error
Hi there,

### Terraform Version
`Terraform v0.10.2`
`provider-openstack_v0.2.0_x4`

### Affected Resource(s)
- `openstack_compute_floatingip_associate_v2`

```hcl
resource "openstack_compute_floatingip_associate_v2" "fip_1" {
  count       = "${var.instance_count}"
  floating_ip = "${element(openstack_networking_floatingip_v2.fip.*.address, count.index)}"
  instance_id = "${element(openstack_compute_instance_v2.webserver.*.id, count.index)}"
}
```

On my current OpenStack provider, floating IP association works, but the resource is unable to realized in subsequent runs that the IP is still associated. Since the resource uses the compute API and check for the external IP adress in the `addresses` list returned, this isn't working for me, since the external network is not listed. Also the APIs used in this resource are mostly marked as deprecated in OpenStack. This is also stated in this comment: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_floatingip_associate_v2.go#L73-L75

For me, this results in Terraform trying to associate the floating IP again, which then results in an error. I don't have that deep of an OpenStack knowledge but aren't the API options that allow for a more generic check if a floating IP is attached?

jtopjian:
This commit adds the wait_until_associated argument to the
openstack_compute_floatingip_associate_v2 resource. This is useful
in environments where the OpenStack environment won't automatically
wait for the floating IP association to finish before returning.

Use this option to instead have Terraform wait until the association
has finished.

Fixes #88

### DIFF
--- a/openstack/resource_openstack_compute_floatingip_associate_v2.go
+++ b/openstack/resource_openstack_compute_floatingip_associate_v2.go
@@ -4,12 +4,15 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
 	nfloatingips "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
-	"github.com/hashicorp/terraform/helper/schema"
 )
 
 func resourceComputeFloatingIPAssociateV2() *schema.Resource {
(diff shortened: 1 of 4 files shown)


openstack/resource_openstack_compute_floatingip_associate_v2.go
```diff
@@ -34,16 +41,25 @@ func resourceComputeFloatingIPAssociateV2() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+
```
fatmcgav: Are these spaces necessary?

jtopjian: They're not _necessary_, but they follow the pattern of other providers. This is something on my list of cleanups I'll get to one day :)

fatmcgav: Ah, ok... Just so happened that the ones I checked _didn't_ have spaces ;( hehe...

jtopjian:
recheck

PR #312, opened 2018-05-03
Docs: Add subnetpool example

jtopjian:
For https://github.com/terraform-providers/terraform-provider-openstack/pull/243#issuecomment-376563210


jtopjian:
The build failure is ignorable here. Will try again, though.

jtopjian:
recheck

PR #313, opened 2018-05-03
Networking v2: Add floating IP associate resource

### LINKED ISSUE #259: openstack_networking_port_floatingip_associate_v2
### Terraform Version
Terraform v0.11.3
+ provider.null v1.0.0
+ provider.openstack v1.2.0
+ provider.powerdns v0.1.0
+ provider.template v1.0.0

### Affected Resource(s)

Today one can associate existing floating IPs to a virtual machine, but not to a port. Of course one can associate an existing port to a floating IP when setting up the floating IP. However in my case the floating IPs are fixed assigned to the tenants and should not be changed as they are referenced by firwalls outside OpenStack. Thus recreating a floating IPs when rebuilding resources within the tenant is not an option and it would be really great if an **openstack_networking_port_floatingip_associate_v2** resource could be introduced to realize the assignment in the same way it is realized by **openstack_compute_floatingip_associate_v2**

Regards

jtopjian:
This commit adds the openstack_networking_floatingip_associate_v2
resource. This is useful for situations where users already have
an allocated floating IP or if they are unable to use the
openstack_networking_floatingip_v2 resource because their cloud
pre-allocates floating IPs.

Fixes #259


scat70:
Hi ! Many thanks for the implementation! With regards to the code unlike the resource "openstack_compute_floatingip_associate_v2" which expects a "floating_ip" the new resources expects  a "floatingip_id". From my perspective I would be a little bit easier to use if instead of the ID once again just the IP could be used. Was this spossible ?

jtopjian:
> With regards to the code unlike the resource "openstack_compute_floatingip_associate_v2" which expects a "floating_ip" the new resources expects a "floatingip_id".

This is a good point. To provide consistency between the two resources, I'll change this to `floating_ip` and expect an IP address rather than a UUID.

scat70:
Perfect, many thanks !

jtopjian:
@scat70 I've made the changes. Let me know if you have any other comments :)

jtopjian:
recheck

Issue #322, opened 2018-05-11
Update vendor for block storage V3


flashvoid:
Hi there,

Support for [Cinder V3](https://github.com/gophercloud/gophercloud/pull/415 ) was around for half a year and i would like to use it on terraform. 

I would send a PR but i don't see a documentation on how to upgrade vendor.

jtopjian:
@flashvoid hmm... yeah, you're right. There's basic documentation in the Terraform core [README](https://github.com/hashicorp/terraform#dependencies) -- we should probably add that here.

I've been meaning to do a full Gophercloud update but haven't had a chance. Because of that, I recommend only vendoring the blockstorage/v3 packages (unless other updates are required):

```shell
$ govendor fetch github.com/gophercloud/gophercloud/openstack/blockstorage/v3/...
```

The three dots should be used.

If you do that to a clean checkout of this provider, you'll notice that nothing actually happens. It's not until you begin creating block storage v3 resources that `govendor` will pull in the required libraries.

When committing resources+vendoring, it's recommended to have the vendor update in a separate commit:

```shell
$ git add vendor
$ git commit -m "Updating Gophercloud for blockstorage resource"
```

Let me know if you run into any problems or have any questions.

flashvoid:
@jtopjian i think i figured vendor stuff out. Would you prefer single PR with 2 commits or separate PRs for vendor update and openstack resources?

jtopjian:
@flashvoid A single PR with one resource + vendor works :)

flashvoid:
@jtopjian  ouch, and how do i update `openstack/client.go` to get `NewBlockStorageV3`
i suppose that doing `govendor fetch github.com/gophercloud/gophercloud/openstack/...` not advised?

jtopjian:
@flashvoid Sorry about that, `...` means "current directory and all subdirectories".

If you just want `github.com/gophercloud/gophercloud/openstack/`, you can do:

```shell
$ govendor fetch github.com/gophercloud/gophercloud/openstack/
```

However, one thing to note is that the `./openstack` directory will be updated in this PR: https://github.com/terraform-providers/terraform-provider-openstack/pull/320

I'll see if I can get this reviewed tomorrow so we don't run into a merge conflict :)

swapnilgm:
Why don't we use go dependency management tool like [dep](https://github.com/golang/dep)?

jtopjian:
[...]

PR #323, opened 2018-05-11
New/update blockstorage vendor

flashvoid:
This PR updates block storage related vendor and introduces CinderV3 resources.

Issue: https://github.com/terraform-providers/terraform-provider-openstack/issues/322

* V3 volumes could be resized on-line, without detaching them first. [see](https://blueprints.launchpad.net/cinder/+spec/extend-attached-volume)


flashvoid:
broken vendor

PR #324, opened 2018-05-11
New/blockstorage v3 resources

flashvoid:
This PR updates blockstorage/extensions and introduces CinderV3 resources.

Issue: https://github.com/terraform-providers/terraform-provider-openstack/issues/322

* V3 volumes could be resized on-line, without detaching them first. [see](https://blueprints.launchpad.net/cinder/+spec/extend-attached-volume)

### DIFF
--- a/openstack/resource_openstack_blockstorage_volume_attach_v3.go
+++ b/openstack/resource_openstack_blockstorage_volume_attach_v3.go
@@ -13,11 +13,11 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
-func resourceBlockStorageVolumeAttachV3() *schema.Resource {
+func resourceBlockStorageVolumeAttachV2() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceBlockStorageVolumeAttachV3Create,
-		Read:   resourceBlockStorageVolumeAttachV3Read,
-		Delete: resourceBlockStorageVolumeAttachV3Delete,
+		Create: resourceBlockStorageVolumeAttachV2Create,
+		Read:   resourceBlockStorageVolumeAttachV2Read,
+		Delete: resourceBlockStorageVolumeAttachV2Delete,
 
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(10 * time.Minute),
@@ -134,9 +134,9 @@ func resourceBlockStorageVolumeAttachV3() *schema.Resource {
 	}
 }
 
-func resourceBlockStorageVolumeAttachV3Create(d *schema.ResourceData, meta interface{}) error {
+func resourceBlockStorageVolumeAttachV2Create(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
-	client, err := config.blockStorageV3Client(GetRegion(d, config))
+	client, err := config.blockStorageV2Client(GetRegion(d, config))
 	if err != nil {
 		return fmt.Errorf("Error creating OpenStack block storage client: %s", err)
 	}
@@ -212,7 +212,7 @@ func resourceBlockStorageVolumeAttachV3Create(d *schema.ResourceData, meta inter
 	}
 
 	// Once the connection has been made, tell Cinder to mark the volume as attached.
-	attachMode, err := blockStorageVolumeAttachV3AttachMode(d.Get("attach_mode").(string))
+	attachMode, err := blockStorageVolumeAttachV2AttachMode(d.Get("attach_mode").(string))
 	if err != nil {
 		return nil
 	}
@@ -235,7 +235,7 @@ func resourceBlockStorageVolumeAttachV3Create(d *schema.ResourceData, meta inter
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"available", "attaching"},
 		Target:     []string{"in-use"},
-		Refresh:    VolumeV3StateRefreshFunc(client, volumeId),
+		Refresh:    VolumeV2StateRefreshFunc(client, volumeId),
 		Timeout:    d.Timeout(schema.TimeoutCreate),
 		Delay:      10 * time.Second,
 		MinTimeout: 3 * time.Second,
@@ -271,17 +271,17 @@ func resourceBlockStorageVolumeAttachV3Create(d *schema.ResourceData, meta inter
 	id := fmt.Sprintf("%s/%s", volumeId, attachmentId)
 	d.SetId(id)
 
-	return resourceBlockStorageVolumeAttachV3Read(d, meta)
+	return resourceBlockStorageVolumeAttachV2Read(d, meta)
 }
 
-func resourceBlockStorageVolumeAttachV3Read(d *schema.ResourceData, meta interface{}) error {
+func resourceBlockStorageVolumeAttachV2Read(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*Config)
-	client, err := config.blockStorageV3Client(GetRegion(d, config))
+	client, err := config.blockStorageV2Client(GetRegion(d, config))
 	if err != nil {
 		return fmt.Errorf("Error creating OpenStack block storage client: %s", err)
 	}
 
-	volumeId, attachmentId, err := blockStorageVolumeAttachV3ParseId(d.Id())
+	volumeId, attachmentId, err := blockStorageVolumeAttachV2ParseId(d.Id())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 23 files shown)


jtopjian:
@flashvoid Thanks for submitting this and for working through the vendor issues :)

I did a quick scan and it looks like `openstack/resource_openstack_blockstorage_volume_attach_v2.go` (and test file) is being modified for v3?

jtopjian:
Ah, additionally it looks like the new attach v3 resources are configured for v2 :)

flashvoid:
Ouch, wrong branch. Will fix.

flashvoid:
@jtopjian i think last test failed due to random networking issue on test platform.
```
* openstack_images_image_v2.image_1: Error opening file for Image: Error downloading image "https://releases.rancher.com/os/latest/rancheros-openstack.img" to file "/root/.terraform/image_cache/9ff941e9296703b0630bc96bbfda1453.img": read tcp 192.168.0.105:55540-&gt;104.24.16.51:443: read: connection reset by peer
```

I ran acceptance tests locally and it went fine.

jtopjian:
Yeah, there are two tests that randomly fail. I'll rerun the tests now.

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
recheck

openstack/resource_openstack_blockstorage_volume_attach_v3.go
```diff
@@ -0,0 +1,414 @@
...
+				ForceNew: true,
+			},
+
+			"volume_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"instance_id": &schema.Schema{
```
jtopjian: Let's remove this argument entirely from the schema. It's only in the v2 resource for legacy reasons.

openstack/resource_openstack_blockstorage_volume_v3.go
```diff
@@ -0,0 +1,380 @@
...
+	updateOpts := volumes.UpdateOpts{
+		Name:        d.Get("name").(string),
+		Description: d.Get("description").(string),
+	}
+
+	if d.HasChange("metadata") {
+		updateOpts.Metadata = resourceVolumeMetadataV3(d)
+	}
+
+	if d.HasChange("size") {
```
jtopjian: Can you add an acceptance test for this? It should be as simple as duplicating the `_basic` test and changing the size in the second step. Let me know if you need help or run into problems, though.

flashvoid:
Hey @jtopjian, take a look at acceptance test. I integrated the check into `TestAccBlockStorageV3Volume_basic` instead of making a new test. Also it only tests offline path since online resize requires a Cinder and specific storage driver support.

Documentation is coming later.

jtopjian:
What happens when the user has an attached/in-use volume on an OpenStack cloud running a release earlier than Pike? Will they see an error? If so, maybe we should have some way for the user to opt-in to online resizing -- maybe a boolean schema argument of `enable_online_resize`. 

We should probably add another test case to test resizing with attached volumes, too.

Thoughts?

flashvoid:
The user should see the same behavior as in volume_v2, namely attempt to resize attached volume produces an error "_volume_must_be_in_`available`_state_". By providing opt-in all we will do is to provide a user with a different error.

Test also doesn't help much, offline resize is tested and online resize test will fail way too often since it requires specific configuration of Cinder, Nova and storage drivers (and only a few are supported so far). Maybe later, when volume_V4 is out?

So, assuming that optin only for sake of producing different error doesn't make sense and final goal is to prevent user from seeing an error - there are some thoughts.

Normal procedure in openstack right now is to take your volume offline, resize it and than attach it back. Can/should it be done by terraform?
1) Detaching volume through terraform might produce unclean state (e.g. if you running postgress there and didn't stop the process, haven't done xfs_freeze/sync)
2) Volume attached after resize won't be mounted automatically, should terraform login and attempt to mount stuff back? it feels wrong.
3) This will trigger volume_attach object recreation during volume resize - not sure if it's bad but have a bad feeling about this as well.

Another option is to fallback to v2 schema - namely destroy the volume and create it again, personally i don't see why would anyone want this - it's destroying data during `resize` operation, but this is a known behavior and falling back to it at least produces less surprize.

So i'm not sure about either option. If you have a strong feeling for this or any other solution i'm happy to do it.



P.S. one way that would work for me is to report and error when online resize requested but Cinder doesn't support it, but reporting this error during **terraform plan**. That's why i started all this discussion in "Microversion". Is there a way to do it right now?

jtopjian:
All really good points.

> online resize test will fail way too often since it requires specific configuration of Cinder, Nova and storage drivers (and only a few are supported so far). Maybe later, when volume_V4 is out?

Good point. We could probably have an environment variable trigger when this test would run (`OS_CINDER_RESIZE=1` or something like that) but if online resizing isn't widely available, then let's forgo testing for now. But at a minimum, I trust that the support your adding has been tested with a working environment? :)

> Can/should it be done by terraform?

This is a blurry area, especially when we're talking about block storage volumes. If the discussion was about containers (or perhaps even instances), one could make the argument that these should be easy to recreate from scratch. But block storage is different since it's frequently used to hold state which might be hard to recreate.

So I think being able to support online resizing is important.

We could/should probably _not_ support offline resizing. As you've outlined, it's a complicated process. Instead, the recommended path would be for the user to add a new volume to their Terraform configuration, attach it through Terraform, then copy the data (`rsync` or some other means) outside of Terraform, then remove the original volume from their configuration. I imagine this is what most users are doing at the moment.

> By providing opt-in all we will do is to provide a user with a different error.

Correct, but the key here is that it's a Terraform error and not an OpenStack error. The difference is that Terraform will halt _before_ it tried to resize the volume. There might be some OpenStack cloud that did something strange where it detached the volume before an error was triggered or something like that. If we guard the resize from happening at all, we can be sure that the volume is never touched.

Of course, this error won't happen until the "apply" stage. That's OK. It's not the best behavior, but it's better than nothing. There are already other areas where this kind of error happens.

> one way that would work for me is to report and error when online resize requested but Cinder doesn't support it, but reporting this error during terraform plan

This would be the best case scenario.

It would require Gophercloud to do microversion discovery, which is on the roadmap. But even if it had that support, it would be difficult to trigger this error during the "plan" stage (this is more of a Terraform "limitation" in how the plan phase works). I think the best that could be done is the same type of error that `enable_online_resize` would provide, but without having to have that option in the schema.

I think we should proceed with the `enable_online_resize` option. This will cause the user to have to research if online resizing is possible in their cloud and opt-in to any side effects that might happen as a result. 

If, at some point, this is option is no longer needed, we can always deprecated and remove the option down the road. I would much rather be in this situation then to have users confused about why we're now supporting resizing but it's not working / triggering an error from OpenStack.

flashvoid:
> But at a minimum, I trust that the support you're adding has been tested in a working environment?

It happened to work in current devstack with lvm driver - it should be easy to test for anyone, maybe making a test that is off-be-default isn't a bad idea just for people who know why they enable it.

> I think we should proceed with the enable_online_resize option

Ok, i can do that. You said you don't want it in schema (makes sense), but how should it be implemented (Environment variable/terraform variable)? can you point to an example?

jtopjian:
> You said you don't want it in schema (makes sense)

That must have been a miscommunication on my part. `enable_online_resize` should be in the schema as a `TypeBool`.

flashvoid:
@jtopjian so i added the option and also provided a test for online resize guarded by `OS_ONLINE_RESIZE`. 

This test is slow and i suggest only using it as a convenience way to determine if given environment supports the feature.

openstack/resource_openstack_blockstorage_volume_v3.go
```diff
@@ -0,0 +1,401 @@
...
+		extendOpts := volumeactions.ExtendSizeOpts{d.Get("size").(int)}
+		err = volumeactions.ExtendSize(blockStorageClient, d.Id(), extendOpts).ExtractErr()
+		if err != nil {
+			return fmt.Errorf(
+				"Error extending volume (%s) size (%s)",
+				d.Id(), err)
+		}
+
+		stateConf := &resource.StateChangeConf{
+			Pending:    []string{"available", "in-use"},
```
jtopjian: I just want to confirm that this is correct. The status here should be set to whatever status will be displayed _while_ the volume is resizing -- would that be something like "resizing"?

"available" and "in-use" would be the end-state statuses.

flashvoid: Yeah, there is another status "extending". which indicates that process going.
Ok i'll try that.

flashvoid:
@jtopjian ok, done

PR #327, opened 2018-05-14
Identity Auth Scope: fix flipped User Domin ID/Name values

jbcraig:


jtopjian:
@jbcraig Thanks so much for catching these and cleaning these up. I'll merge this once testing is done.

jbcraig:
No problem, happy to help

Issue #2, opened 2017-06-13
Provider: Openstack - VPNaaS


hashibot:
_This issue was originally opened by @sandywater as hashicorp/terraform#4503. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I would like to request support for VPNaaS in Terraform.  Please !!! :smile:

hashibot:
_This comment was originally opened by @nicoaina as https://github.com/hashicorp/terraform/issues/4503#issuecomment-246379984. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

:+1:

hashibot:
_This comment was originally opened by @jsecchiero as https://github.com/hashicorp/terraform/issues/4503#issuecomment-279717841. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

:+1:

hashibot:
_This comment was originally opened by @kessenza as https://github.com/hashicorp/terraform/issues/4503#issuecomment-279718506. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

👍

hashibot:
_This comment was originally opened by @fnzv as https://github.com/hashicorp/terraform/issues/4503#issuecomment-279719049. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

:+1:

hashibot:
_This comment was originally opened by @jtopjian as https://github.com/hashicorp/terraform/issues/4503#issuecomment-279729291. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

Hi all,

I appreciate the enthusiasm of all the :+1: 's, however I think it's preferable to simply use the "Add a reaction" button.

In order for VPNaaS to be added, it first needs to be added to https://github.com/gophercloud/gophercloud. New extensions are added to Gophercloud as time permits for the contributors. If someone would like to start there, it would certainly help expedite this.

[...]

PR #330, opened 2018-05-24
Vpnaas: IPsec site connection resource

simonre:
Part 5 out of 5 of a series that addresses #2 

@jtopjian This has the same problem as #300 and https://github.com/gophercloud/gophercloud/pull/784 so I left out the "update" acceptance test. Let me know if this is not acceptable or you have a better idea :)

This is ready for review!

### DIFF
--- a/openstack/resource_openstack_vpnaas_site_connection_test.go
+++ b/openstack/resource_openstack_vpnaas_site_connection_test.go
@@ -137,9 +137,9 @@ var testAccSiteConnectionV2_basic = fmt.Sprintf(`
 		vpnservice_id = "${openstack_vpnaas_service_v2.service_1.id}"
 		psk = "secret"
 		peer_address = "192.168.10.1"
+		peer_id = "192.168.10.1"
 		local_ep_group_id = "${openstack_vpnaas_endpoint_group_v2.group_2.id}"
 		peer_ep_group_id = "${openstack_vpnaas_endpoint_group_v2.group_1.id}"
 		depends_on = ["openstack_networking_router_interface_v2.router_interface_1"]
-		peer_id = "9b00d6b0-6c93-4ca5-9747-b8ade7bb514f"
 	}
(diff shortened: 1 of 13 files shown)


simonre:
@jtopjian No, they are not. Not sure how I missed that. The fields in the docs are actually more correct than the schema fields when it comes to that.

simonre:
@jtopjian I made some of the parameters required

openstack/resource_openstack_vpnaas_site_connection_test.go
```diff
@@ -140,5 +140,6 @@ var testAccSiteConnectionV2_basic = fmt.Sprintf(`
 		local_ep_group_id = "${openstack_vpnaas_endpoint_group_v2.group_2.id}"
 		peer_ep_group_id = "${openstack_vpnaas_endpoint_group_v2.group_1.id}"
 		depends_on = ["openstack_networking_router_interface_v2.router_interface_1"]
+		peer_id = "9b00d6b0-6c93-4ca5-9747-b8ade7bb514f"
```
jtopjian: Is it possible to dynamically determine this? Otherwise the test will never work correctly.

simonre: It was set to a value it shouldn't be set to anyways. It should have been set to the same value as peer_address

simonre:
@jtopjian This is ready for review again :)

jtopjian:
@simonre I apologize for the delay. This looks good to me. Thank you! :)

simonre:
@jtopjian no worries. As far as I can see, that's everything implemented for #2 then or can you think of anything else?

jtopjian:
@simonre if it looks good to you, I'll close #2. We can open additional issues if anything comes up.

Thank you very much for your work on this over the past few months! This was a large amount of work and I really appreciate your time.

PR #332, opened 2018-05-30
Docs: Fixing CIDR documentation in subnet resource

jtopjian:
As mentioned in https://github.com/terraform-providers/terraform-provider-openstack/pull/294#issuecomment-392710506


jtopjian:
@fatmcgav This is ready for review at your convenience :)

PR #340, opened 2018-06-10
Add support to modify SSL settings via clouds.yaml

### LINKED ISSUE #339: Add support to modify SSL settings via clouds.yaml
### Terraform Version
```
Terraform v0.11.7
+ provider.openstack v1.5.0
```

### Affected Resource(s)
- provider

```hcl
provider "openstack" {
  cloud = "test"
}

data "openstack_networking_network_v2" "public" {
  name = "public"
}
```

clouds.yaml
```json
clouds:
  test:
    auth:
      auth_url: https://192.168.1.254:5000/v3
      username: "test"
      password: "test"
      project_name: "test"
      domain_name: "Default"
    region_name: "RegionOne"
    interface: "public"
    identity_api_version: 3
    verify: False
```

### Expected Behavior
Refresh of resources successful and no changes expected when running `terraform plan`

### Actual Behavior
Got error: `x509: certificate signed by unknown authority`

### Steps to Reproduce
1. `terraform plan`

### References
Upstream configuration is already implemented by:
- https://github.com/gophercloud/utils/issues/45

haxorof:
Fixes #339

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -29,7 +29,7 @@ type Config struct {
 	DomainName        string
 	EndpointType      string
 	IdentityEndpoint  string
-	Insecure          bool
+	Insecure          *bool
 	Password          string
 	Region            string
 	Swauth            bool
@@ -81,6 +81,23 @@ func (c *Config) LoadAndValidate() error {
 		if c.Region == "" && cloud.RegionName != "" {
 			c.Region = cloud.RegionName
 		}
+
+		if c.CACertFile == "" && cloud.CACertFile != "" {
+			c.CACertFile = cloud.CACertFile
+		}
+
+		if c.ClientCertFile == "" && cloud.ClientCertFile != "" {
+			c.ClientCertFile = cloud.ClientCertFile
+		}
+
+		if c.ClientKeyFile == "" && cloud.ClientKeyFile != "" {
+			c.ClientKeyFile = cloud.ClientKeyFile
+		}
+
+		if c.Insecure == nil && cloud.Verify != nil {
+			// Insecure == No verification
+			c.Insecure = func(b bool) *bool { return &b }(!*cloud.Verify)
+		}
 	} else {
 		authInfo := &clientconfig.AuthInfo{
 			AuthURL:           c.IdentityEndpoint,
@@ -125,8 +142,10 @@ func (c *Config) LoadAndValidate() error {
 		config.RootCAs = caCertPool
 	}
 
-	if c.Insecure {
-		config.InsecureSkipVerify = true
+	if c.Insecure == nil {
+		config.InsecureSkipVerify = false
+	} else {
+		config.InsecureSkipVerify = *c.Insecure
 	}
 
 	if c.ClientCertFile != "" && c.ClientKeyFile != "" {
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -95,8 +95,8 @@ func (c *Config) LoadAndValidate() error {
 		}
 
 		if c.Insecure == nil && cloud.Verify != nil {
-			// Insecure == No verification
-			c.Insecure = func(b bool) *bool { return &b }(!*cloud.Verify)
+			v := (!*cloud.Verify)
+			c.Insecure = &v
 		}
 	} else {
 		authInfo := &clientconfig.AuthInfo{
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -330,10 +330,6 @@ func toBoolPtr(d *schema.ResourceData, key string) *bool {
 }
 
 func configureProvider(d *schema.ResourceData) (interface{}, error) {
-	// Config order of precedence
-	// 1. Terraform config
-	// 2. System Environment
-	// 3. OpenStack cloud config (only if cloud is defined)
 	config := Config{
 		CACertFile:        d.Get("cacert_file").(string),
 		ClientCertFile:    d.Get("cert").(string),
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -137,7 +137,7 @@ func Provider() terraform.ResourceProvider {
 			"insecure": &schema.Schema{
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("OS_INSECURE", ""),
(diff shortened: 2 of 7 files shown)


jtopjian:
recheck

openstack/config.go
```diff
@@ -81,6 +81,23 @@ func (c *Config) LoadAndValidate() error {
...
+			c.ClientCertFile = cloud.ClientCertFile
+		}
+
+		if c.ClientKeyFile == "" && cloud.ClientKeyFile != "" {
+			c.ClientKeyFile = cloud.ClientKeyFile
+		}
+
+		if c.Insecure == nil && cloud.Verify != nil {
+			// Insecure == No verification
+			c.Insecure = func(b bool) *bool { return &b }(!*cloud.Verify)
```
jtopjian: I think doing something more explicit might be better:

```go
v := (!*cloud.Verify)
c.Insecure = &v
```

This will help with readability.

openstack/provider.go
```diff
@@ -330,7 +343,7 @@ func configureProvider(d *schema.ResourceData) (interface{}, error) {
 		DomainName:        d.Get("domain_name").(string),
 		EndpointType:      d.Get("endpoint_type").(string),
 		IdentityEndpoint:  d.Get("auth_url").(string),
-		Insecure:          d.Get("insecure").(bool),
+		Insecure:          toBoolPtr(d, "insecure"),
```
jtopjian: The way this is handled throughout this provider is to do:

```go
insecure := d.Get("insecure")
config := Config{
  ...
  Insecure: &insecure,
}
```

openstack/provider.go
```diff
@@ -320,7 +320,20 @@ func init() {
...
+		return &tmpValue
+	}
+	return nil
+}
+
 func configureProvider(d *schema.ResourceData) (interface{}, error) {
+	// Config order of precedence
+	// 1. Terraform config
+	// 2. System Environment
+	// 3. OpenStack cloud config (only if cloud is defined)
```
jtopjian: Let's remove this comment as it could be confusing depending on interpretation.

For example: the Terraform schema handles reading in and setting environment variables, so it's arguable whether 1 implicitly does 2 or if there's an explicit step 2 that someone won't be able to find :)

openstack/config.go
haxorof: Yes, that's easier to read. :smile:
Will fix.

openstack/provider.go
haxorof: No problem. Will remove it

openstack/provider.go
haxorof: I was testing a bit but I did not see the Get function returned nil, so that is the reason I did the function. I will check that again and change it if the above works.

jtopjian: Thanks for looking into it. If you can, please also test the possible combinations of this so we don't run into a crash of some sort. This might turn out to be a little tricky so let me know if you need help.

haxorof: Just tried to write the code you suggested but that gives me compilation error:
`cannot use &insecure (type *interface {}) as type *bool in field value`

Do you have any other suggestion? If so please share because I can change as long as it fulfills the expected behavior.

jtopjian: You might have to do something like:

```go
var insecure *bool
if v, ok := d.Get("insecure").(*bool); ok {
    insecure = v
}

config := Config{
  ...
  Insecure: insecure,
}
```

jtopjian: On second thought, you might have to do 

```go
var insecure bool
```

But I'm not sure. Maybe another option is to use [`GetOkExists`](https://github.com/hashicorp/terraform/blob/0c39b69d50ad6d54c4fe61d310cd45183dbf9b8c/helper/schema/resource_data.go#L109-L123) instead of `GetOk`, but I'm not sure how that'll work when combined with the `EnvDefaultFunc`.

jtopjian: This seems to be working for me:

```diff
diff --git a/openstack/provider.go b/openstack/provider.go
index a277b70..43ec3d0 100644
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -320,15 +320,6 @@ func init() {
        }
 }

-func toBoolPtr(d *schema.ResourceData, key string) *bool {
-       in, inExists := d.GetOkExists(key)
-       if inExists {
-               var tmpValue = in.(bool)
-               return &tmpValue
-       }
-       return nil
-}
-
 func configureProvider(d *schema.ResourceData) (interface{}, error) {
        config := Config{
                CACertFile:        d.Get("cacert_file").(string),
@@ -339,7 +330,6 @@ func configureProvider(d *schema.ResourceData) (interface{}, error) {
                DomainName:        d.Get("domain_name").(string),
                EndpointType:      d.Get("endpoint_type").(string),
                IdentityEndpoint:  d.Get("auth_url").(string),
-               Insecure:          toBoolPtr(d, "insecure"),
                Password:          d.Get("password").(string),
                ProjectDomainID:   d.Get("project_domain_id").(string),
                ProjectDomainName: d.Get("project_domain_name").(string),
@@ -355,6 +345,12 @@ func configureProvider(d *schema.ResourceData) (interface{}, error) {
                useOctavia:        d.Get("use_octavia").(bool),
        }

+       v, ok := d.GetOkExists("insecure")
+       if ok {
+               insecure := v.(bool)
+               config.Insecure = &insecure
+       }
+
        if err := config.LoadAndValidate(); err != nil {
                return nil, err
        }
```

`GetOkExists` seems to be working as it should when the environment variable is either set or not. 

If this isn't working for you, let me know which use-case it's failing for.

haxorof: hmm.. okay. I will do a test with that again and let you know about the result. Might be that during the different tests I did I got confused. :smile: Thanks @jtopjian

haxorof:
Updated and did some quick tests:

```
| env (insecure) | provider | clouds.yml (verify) | expected | result |
| -------------- | -------- | ------------------- | -------- | ------ |
| -              | -        | -                   | secure   | OK     |
| true           | -        | -                   | insecure | OK     |
| -              | true     | -                   | insecure | OK     |
| -              | -        | false               | insecure | OK     |
| -              | -        | true                | secure   | OK     |
```

Cheers! 😄

jtopjian:
Awesome test matrix 😄 

This looks good to me -- really nice work!

Issue #347, opened 2018-06-20
enhancement request: allow resize for stopped VMs


bogdansavu:
Hello,

It is possible to update the openstack_compute_instance_v2 to allow us to resize the VMs even if there are stopped. 

### Terraform Version
Terraform v0.11.7.

### Affected Resource(s)
- openstack_compute_instance_v2

### Expected Behavior
The resize VM should succeed if the VM is stopped(SHUTOFF state).

### Actual Behavior
* openstack_compute_instance_v2.vm1: Error waiting for instance (de8985e5-e06a-492d-8c1c-15b5b85c72ea) to confirm resize: unexpected state 'SHUTOFF', wanted target 'ACTIVE'. last error: %!s(<nil>)

### Steps to fix the problem:
Edit resource_openstack_compute_instance_v2.go and update the line 730 and change:
from:
Target:     []string{"ACTIVE"},
to:
Target:     []string{"ACTIVE", "SHUTOFF"},

jtopjian:
@bogdansavu Thanks for reporting this. To confirm: you've tested the proposed fix and it works?

bogdansavu:
Yes , I tested it today and worked fine.

jtopjian:
@bogdansavu Thanks for the confirmation! If you'd like to open a PR for this, I'd be happy to merge it. If not, I can get this fixed in the next day or two.

jtopjian:
Implemented in #348

PR #348, opened 2018-06-20
allow resize for stopped VMs (#347)

bogdansavu:


jtopjian:
recheck

jtopjian:
recheck

jtopjian:
@bogdansavu What happens if a user specifies a state of `shutoff` when the instance is initially created?

Additionally, where you have:

```
d.Set("state", strings.ToLower(server.Status))
```

What happens if `server.Status` is `ERROR`?

bogdansavu:
@jtopjian I created a new PR for (#349).

Issue #349, opened 2018-06-20
enhancement request: allow vm state change


bogdansavu:
Hello,

### Terraform Version
Terraform v0.11.7

### Affected Resource(s)
- resource_openstack_compute_instance_v2

### Expected Behavior
Have a property to allow us to change the vm state for an instance.

### Actual Behavior
There is not any option to modify the vm state.

### Solution
Implement a new property like vm_state to allow us to change the state to ACTIVE or SHUTOFF.

bogdansavu:
I implemented the state property (optional) , that allow us to change the VM state.

jtopjian:
Implemented in #350

PR #350, opened 2018-06-23
allow vm state change (#349)

bogdansavu:


bogdansavu:
What happens if a user specifies a state of shutoff when the instance is initially created?
The VM is created and then stopped, the final state will be shutoff.

Additionally, where you have:

d.Set("state", strings.ToLower(server.Status))
What happens if server.Status is ERROR?
edit: with regard to the above, the StateChangeConf should handle filtering instances of type ERROR.

Updated the code , if we have any other state than "active" and "shutoff" the vm state change will fail.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -346,6 +346,19 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Type:     schema.TypeMap,
 				Computed: true,
 			},
+			"state": &schema.Schema{
```
jtopjian: I think we need to have a default value of "active" here.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -475,6 +488,32 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
 			server.ID, err)
 	}
 
+	if d.Get("state").(string) != "" {
```
jtopjian: With a default value in the schema, this can be removed.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -475,6 +488,32 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
...
+		vmState := d.Get("state").(string)
+		if strings.ToLower(vmState) == "shutoff" {
+			err = startstop.Stop(computeClient, d.Id()).ExtractErr()
+			if err != nil {
+				log.Printf("[WARN] Error stopping OpenStack instance: %s", err)
```
jtopjian: If the action failed, we should return an error rather than log one.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -475,6 +488,32 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
...
+			} else {
+				stopStateConf := &resource.StateChangeConf{
+					//Pending:    []string{"ACTIVE"},
+					Target:     []string{"SHUTOFF"},
+					Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
+					Timeout:    3 * time.Minute,
```
jtopjian: `Timeout` should be:

```
Timeout: d.Timeout(schema.TimeoutCreate),
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -475,6 +488,32 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
...
+					Delay:      10 * time.Second,
+					MinTimeout: 3 * time.Second,
+				}
+
+				log.Printf("[DEBUG] Waiting for instance (%s) to stop", d.Id())
+				_, err = stopStateConf.WaitForState()
+				if err != nil {
+					log.Printf("[WARN] Error waiting for instance (%s) to stop: %s, proceeding to delete", d.Id(), err)
```
jtopjian: An error should be returned rather than logged.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -577,6 +616,16 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
 	// Set the region
 	d.Set("region", GetRegion(d, config))
 
+	//Set the current state:
```
jtopjian: nit: 

```
// Set the current state
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -577,6 +616,16 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
...
+	if d.Get("state").(string) != "" {
```
jtopjian: This `if` statement can be removed.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -577,6 +616,16 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
...
+		if strings.ToLower(server.Status) != "active" && strings.ToLower(server.Status) != "shutoff" {
+			return fmt.Errorf("Error changing the instance(%s) state. Invalid current state:%s ", d.Id(), server.Status)
+		} else {
+			d.Set("state", strings.ToLower(server.Status))
+		}
```
jtopjian: Because you're returning an error, you can remove the `else` statement. I think we need to re-word the error message, too. Maybe something like:

```go
if strings.ToLower(server.Status) != "active" && strings.ToLower(server.Status) != "shutoff" {
  return fmt.Errorf("Unable to set state argument. Instance (%s) has an invalid state: %s", d.Id(), server.Status)
}

d.Set("state", strings.ToLower(server.Status))
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
 		}
 	}
 
+	if d.Get("state").(string) != "" {
```
jtopjian: This can be removed if the schema has a default value.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+		if d.HasChange("state") {
+			vmState := d.Get("state").(string)
+			if strings.ToLower(vmState) == "shutoff" {
+				err = startstop.Stop(computeClient, d.Id()).ExtractErr()
+				if err != nil {
+					log.Printf("[WARN] Error stopping OpenStack instance: %s", err)
```
jtopjian: We should return an error rather than log one.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+				} else {
+					stopStateConf := &resource.StateChangeConf{
+						//Pending:    []string{"ACTIVE"},
+						Target:     []string{"SHUTOFF"},
+						Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
+						Timeout:    3 * time.Minute,
```
jtopjian: This should be:

```
Timeout: d.Timeout(schema.TimeoutUpdate),
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+						Delay:      10 * time.Second,
+						MinTimeout: 3 * time.Second,
+					}
+
+					log.Printf("[DEBUG] Waiting for instance (%s) to stop", d.Id())
+					_, err = stopStateConf.WaitForState()
+					if err != nil {
+						log.Printf("[WARN] Error waiting for instance (%s) to stop: %s, proceeding to delete", d.Id(), err)
```
jtopjian: We should return an error rather than log one.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+					}
+				}
+
+			}
+			if strings.ToLower(vmState) == "active" {
+				err = startstop.Start(computeClient, d.Id()).ExtractErr()
+				if err != nil {
+					log.Printf("[WARN] Error starting OpenStack instance: %s", err)
```
jtopjian: We should return an error rather than log one.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+				} else {
+					startStateConf := &resource.StateChangeConf{
+						//Pending:    []string{"SHUTOFF"},
+						Target:     []string{"ACTIVE"},
+						Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
+						Timeout:    3 * time.Minute,
```
jtopjian: This should be:

```
Timeout: d.Timeout(schema.TimeoutUpdate),
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +648,57 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+						Delay:      10 * time.Second,
+						MinTimeout: 3 * time.Second,
+					}
+
+					log.Printf("[DEBUG] Waiting for instance (%s) to start", d.Id())
+					_, err = startStateConf.WaitForState()
+					if err != nil {
+						log.Printf("[WARN] Error waiting for instance (%s) to start: %s, proceeding to delete", d.Id(), err)
```
jtopjian: We should return an error rather than log one.

bogdansavu:
@jtopjian Thank for your review, I run some additional tests, and if the resource has provisioners the deployment is timed out because the VM is shutdown and the ssh is not working. The VM will be always active first time , and then we can change the state. I'll implement new changes.

bogdansavu:
@jtopjian Please let me know if is ok to have initial state 'active' even if we set the state 'shutoff'. In this case we can deploy and customise an instance, and run the apply again to shutoff the instance. If not I don't know how to stop the instance after provisioners. If is ok I'll try to create some acceptance tests deploy, change state to shutoff , change state to active and destroy. Thank you.

jtopjian:
@bogdansavu Good idea about testing with provisioners. 

> Please let me know if is ok to have initial state 'active' even if we set the state 'shutoff'

I'm not a fan of this since the initial state of the instance is not the true desired state that the user requested. The easy way around this is to document that instances declared with an initial state of "shutoff" cannot be provisioned. 

I'll ask for some advice internally tomorrow or Tuesday and will get back to you. :)

bogdansavu:
@jtopjian I think will be better to implement also shutoff state for initial instance.  Update the documentation and specify , if the initial state is the shutoff the VM will be stopped immediately after build and the provisioners like remote-exec or files are not supported for shutoff VMs.

jtopjian:
@bogdansavu Perfect. That's the same input I got from asking around, too. :)

If the user _really_ wanted to provision a shutoff instance, they can always use cloudinit/user_data.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -475,6 +489,30 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
...
+				Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
+				Timeout:    d.Timeout(schema.TimeoutCreate),
+				Delay:      10 * time.Second,
+				MinTimeout: 3 * time.Second,
+			}
+
+			log.Printf("[DEBUG] Waiting for instance (%s) to stop", d.Id())
+			_, err = stopStateConf.WaitForState()
+			if err != nil {
+				return fmt.Errorf("Error waiting for instance (%s) to stop: %s, proceeding to delete", d.Id(), err)
```
jtopjian: I don't think the instance will actually be deleted, so let's remove that portion of the error message.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -475,6 +489,30 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
 			server.ID, err)
 	}
 
+	vmState := d.Get("state").(string)
+	if strings.ToLower(vmState) == "shutoff" {
+		err = startstop.Stop(computeClient, d.Id()).ExtractErr()
+		if err != nil {
+			return fmt.Errorf("Error stopping OpenStack instance: %s", err)
+		} else {
```
jtopjian: since you are returning an error in the previous line, you don't need `else`:

```go
if err != nil {
   return ...
}

stopStateConf := ...
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +644,52 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
 		}
 	}
 
+	if d.HasChange("state") {
+		vmState := d.Get("state").(string)
+		if strings.ToLower(vmState) == "shutoff" {
+			err = startstop.Stop(computeClient, d.Id()).ExtractErr()
+			if err != nil {
+				return fmt.Errorf("Error stopping OpenStack instance: %s", err)
+			} else {
```
jtopjian: `else` can be removed here.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +644,52 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+				if err != nil {
+					return fmt.Errorf("Error waiting for instance (%s) to stop: %s, proceeding to delete", d.Id(), err)
+				}
+			}
+		}
+		if strings.ToLower(vmState) == "active" {
+			err = startstop.Start(computeClient, d.Id()).ExtractErr()
+			if err != nil {
+				return fmt.Errorf("Error starting OpenStack instance: %s", err)
+			} else {
```
jtopjian: `else` can be removed here.

openstack/resource_openstack_compute_instance_v2_test.go
```diff
@@ -40,6 +41,45 @@ func TestAccComputeV2Instance_basic(t *testing.T) {
 	})
 }
 
+func TestAccComputeV2Instance_state(t *testing.T) {
```
jtopjian: This is a really good test -- nice work.

Let's rename this to `TestAccComputeV2Instance_initialStateActive` and make a second test called `TestAccComputeV2Instance_initialStateShutoff` which starts with the instance in `shutoff` state.

website/docs/r/compute_instance_v2.html.markdown
```diff
@@ -354,6 +354,8 @@ The following arguments are supported:
     forcefully deleted. This is useful for environments that have reclaim / soft
     deletion enabled.
 
+* `state` - (Optional) Provide the VM state. Only 'active' and 'shutoff' are supported values. *Note*: If the initial state is the shutoff the VM will be stopped immediately after build and the provisioners like remote-exec or files are not supported.
```
jtopjian: doc nit: break the lines at 80 characters.

bogdansavu:
@jtopjian Never is to late to change , I also think the power_state is better. I did the changes, please review the code and let me know if I should do any other change. Thank you.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +642,50 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+		if strings.ToLower(vmState) == "shutoff" {
+			err = startstop.Stop(computeClient, d.Id()).ExtractErr()
+			if err != nil {
+				return fmt.Errorf("Error stopping OpenStack instance: %s", err)
+			}
+			stopStateConf := &resource.StateChangeConf{
+				//Pending:    []string{"ACTIVE"},
+				Target:     []string{"SHUTOFF"},
+				Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
+				Timeout:    d.Timeout(schema.TimeoutCreate),
```
jtopjian: `TimeoutUpdate` instead of `TimeoutCreate`

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -599,6 +642,50 @@ func resourceComputeInstanceV2Update(d *schema.ResourceData, meta interface{}) e
...
+		if strings.ToLower(vmState) == "active" {
+			err = startstop.Start(computeClient, d.Id()).ExtractErr()
+			if err != nil {
+				return fmt.Errorf("Error starting OpenStack instance: %s", err)
+			}
+			startStateConf := &resource.StateChangeConf{
+				//Pending:    []string{"SHUTOFF"},
+				Target:     []string{"ACTIVE"},
+				Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
+				Timeout:    d.Timeout(schema.TimeoutCreate),
```
jtopjian: `TimeoutUpdate` instead of `TimeoutCreate`

bogdansavu:
@jtopjian It is ok, don't worry, please review the latest changes and let me know if I need to do any other update. Thank you.

PR #353, opened 2018-06-27
Switch peer_ep_group_id to Optional and documentation fix

nfillot:
Hello,

First a big thanks to @simonre for his work on vpnaas.

I noticed two things : 
 - peer_ep_group_id should be optional in order to respect the [Openstack API](https://developer.openstack.org/api-ref/network/v2/index.html#create-ipsec-connection). It's currently impossible to specify only peer_cidrs when you have a VPN service with a subnet 
 - a minor nitpick inside documentation (Optional inside the schema but Required in the documentation).

Regards


jtopjian:
> First a big thanks to @simonre for his work on vpnaas.

+100

I'm OK with this change. 

Please let us know if you run into any other issues with these resources.

Issue #356, opened 2018-07-01
Can't create external network with terraform 


HassanIstanbouli:
### Terraform Version
11.0.7

### Affected Resource(s)

- openstack_networking_network_v2

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Expected Behavior
it seems like ,terraform dose not support "**External Network** " in OpenStack Provider because i can't create it . in other words there isn't an "External Network" Attribute .
i can make it shared but i can't make it external :(

### Actual Behavior
no attribute to do that .

### Important Factoids
OpenStack "Newton" release .

### References
https://www.terraform.io/docs/providers/openstack/r/networking_network_v2.html

jtopjian:
Implemented in #357

PR #357, opened 2018-07-01
Networking V2: support external network resource

ozerovandrei:
Add "router:external" OpenStack Networking service attribute to the resource_openstack_networking_network_v2.
Add tests for create and update of network with that attribute.

For #356

### DIFF
--- a/openstack/resource_openstack_networking_network_v2.go
+++ b/openstack/resource_openstack_networking_network_v2.go
@@ -55,7 +55,7 @@ func resourceNetworkingNetworkV2() *schema.Resource {
 				Computed: true,
 			},
 			"external": &schema.Schema{
-				Type:     schema.TypeString,
+				Type:     schema.TypeBool,
 				Optional: true,
 				ForceNew: false,
 				Computed: true,
@@ -140,27 +140,19 @@ func resourceNetworkingNetworkV2Create(d *schema.ResourceData, meta interface{})
 		createOpts.Shared = &shared
 	}
 
-	networkIsExternal := false
-	externalRaw := d.Get("external").(string)
-	if externalRaw != "" {
-		networkIsExternal, err = strconv.ParseBool(externalRaw)
-		if err != nil {
-			return fmt.Errorf("external, if provided, must be either 'true' or 'false': %v", err)
-		}
-	}
-
 	segments := resourceNetworkingNetworkV2Segments(d)
 
+	isExternal := d.Get("external").(bool)
 	n := &networks.Network{}
 	if len(segments) > 0 {
 		providerCreateOpts := provider.CreateOptsExt{
 			CreateOptsBuilder: createOpts,
 			Segments:          segments,
 		}
-		if networkIsExternal {
+		if isExternal {
 			createExternalOpts := external.CreateOptsExt{
 				CreateOptsBuilder: providerCreateOpts,
-				External:          &networkIsExternal,
+				External:          &isExternal,
 			}
 			log.Printf("[DEBUG] Create Options: %#v", createExternalOpts)
 			n, err = networks.Create(networkingClient, createExternalOpts).Extract()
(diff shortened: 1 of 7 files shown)


openstack/resource_openstack_networking_network_v2.go
```diff
@@ -53,6 +54,12 @@ func resourceNetworkingNetworkV2() *schema.Resource {
 				ForceNew: false,
 				Computed: true,
 			},
+			"external": &schema.Schema{
+				Type:     schema.TypeString,
```
jtopjian: We can use `TypeBool` here to automatically check the true/false values.

website/docs/r/networking_network_v2.html.markdown
```diff
@@ -75,6 +75,10 @@ The following arguments are supported:
     by any tenant or not. Changing this updates the sharing capabalities of the
     existing network.
 
+* `external` - (Optional)  Specifies whether the network resource has the
+    external routing facility. Changing this updates the external attribute of
+    the existing network.
```
jtopjian: "Valid values are `true` and `false`. Defaults to `false`."

openstack/resource_openstack_networking_network_v2.go
ozerovandrei: @jtopjian I didn't think about it when I copied checks from the `shared` attribute. Thank you, I'll fix it.

ozerovandrei:
@jtopjian I actually thought that I must ask you to edit Neutron's policy.json to allow those actions. But `testAccPreCheckAdminOnly(t)` looks good too. I'll add it to tests then.

jtopjian:
@ozerovandrei The downside is that those test aren't run in CI at this time, but that will most likely be enabled in the future.

ozerovandrei:
@jtopjian is there something else that needs to be fixed?

jtopjian:
@ozerovandrei I apologize, I haven't had time to do another review on this. I should be able to sometime today.

ozerovandrei:
@jtopjian no problem, I just thought maybe I missed something 😄

HassanIstanbouli:
Thanks @ozerovandrei  for your interest in this issue .
i hope this issue solved in the next Terraform release because currently i have to create the external network manually form openstack dashboard .
P.S : i saw a commit ( form you i guess ! ) in "terraform-provider-openstack" repository related to this issue but after a few hours it's gone !!! .
so what's happening now ?!!

ozerovandrei:
@HassanIstanbouli Hello. There is no need to worry, those changes will be added eventually.

HassanIstanbouli:
ok @ozerovandrei 
but by the way i would like to refer to another issue i faced with terraform : 
when you want to add an interface to a router you can't specify an optional IP to this interface ,it's automatically take the gateway IP you already specify when you create the subnet ,although you can specify it in openstack dashboard .
this issue cause an error if you want to link 2 router interfaces to the same subnet .
should i post a new issue related to this problem ?!

ozerovandrei:
@HassanIstanbouli its better to discuss that in a separate issue of course.

But you can create `networking_router_interface_v2` resources and specify both subnet and port inside them: [documentation](https://www.terraform.io/docs/providers/openstack/r/networking_router_interface_v2.html).
Attaching several ports from a single subnet to a single router sounds very strange to me (but I'm not a Neutron expert, so maybe this schema will work).
I guess that you should check this schema via the standard `openstack` client and if it works, you can try `networking_router_interface_v2` Terraform resource.

HassanIstanbouli:
@ozerovandrei  i told you , i tried that from openstack dashboard and i could specify an IP to it .
i need two interfaces connected to the same subnet because i want to isolate some instances in different network .
briefly my scenario is : 
i have 1 external network  and 2 internal network , the first internal have 2 instances and it connects to the external by a router ,the second internal have 2 instances and it connects to the first internal by another router . so the problem is that the second router can't attach second interface from second internal because the gateway for this second internal has already taken form first router interface . so the solution is to specify another IP (optional) , terraform doesn't have attribute to specify this optional IP .
35973170_1816751655057230_6223367398125207552_n

PR #358, opened 2018-07-01
Networking V2: support external network datasource

ozerovandrei:
Add "router:external" OpenStack Networking service attribute to the data_source_openstack_networking_network_v2
Add tests for datasource network with that attribute.

For #356

### DIFF
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -54,7 +54,7 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Computed: true,
 			},
 			"external": &schema.Schema{
-				Type:     schema.TypeString,
+				Type:     schema.TypeBool,
 				Optional: true,
 			},
 			"availability_zone_hints": &schema.Schema{
@@ -79,17 +79,10 @@ func dataSourceNetworkingNetworkV2Read(d *schema.ResourceData, meta interface{})
 		listOpts.Status = v.(string)
 	}
 
-	networkIsExternal := false
-	externalRaw := d.Get("external").(string)
-	if externalRaw != "" {
-		networkIsExternal, err = strconv.ParseBool(externalRaw)
-		if err != nil {
-			return fmt.Errorf("external, if provided, must be either 'true' or 'false': %v", err)
-		}
-	}
+	isExternal := d.Get("external").(bool)
 	listExternalOpts := external.ListOptsExt{
 		ListOptsBuilder: listOpts,
-		External:        &networkIsExternal,
+		External:        &isExternal,
 	}
 
 	pages, err := networks.List(networkingClient, listExternalOpts).AllPages()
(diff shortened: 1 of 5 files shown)


openstack/data_source_openstack_networking_network_v2.go
```diff
@@ -52,6 +53,10 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"external": &schema.Schema{
+				Type:     schema.TypeString,
```
jtopjian: Similar here, we can use a `TypeBool` field. I'm actually surprised `shared` and `admin_state_up` were added as strings. I'll make a note to fix these, too.

In addition, can you move this above `admin_state_up`? It's not obvious but the arguments are grouped by `Optional` (input) and `Computed` (output-only).

openstack/data_source_openstack_networking_network_v2_test.go
```diff
@@ -74,6 +74,30 @@ func TestAccOpenStackNetworkingNetworkV2DataSource_networkID(t *testing.T) {
 	})
 }
 
+func TestAccOpenStackNetworkingNetworkV2DataSource_external(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
```
jtopjian: This will need restricted to admin-only, too.

openstack/data_source_openstack_networking_network_v2.go
ozerovandrei: @jtopjian for sure. I haven't updated this yet only because I have been waiting for resource PR.

ozerovandrei:
@jtopjian that a nice catch and bug sounds interesting too 😄

jtopjian:
@ozerovandrei https://github.com/gophercloud/gophercloud/pull/1092 should fix the issue... I think.

I'll get that merged in the next day or so and we can vendor it here to confirm :)

jtopjian:
OK, merged!

jtopjian:
Looks like that Gophercloud patch did the trick :)

This looks good to me. Let me know if this is ready to be merged.

jrandall:
As noted in #382, this PR appears to have changed the default behaviour of terraform-provider-openstack in a backwards-incompatible way, such that anyone who has used this data source to refer to an external network is required to change their configuration to reflect that. 

Prior to this PR, external network data could be looked up like this:
```hcl
provider "openstack" {
  version     = "~> 1.6.0"
}

data "openstack_networking_network_v2" "external_network" {
  name = "public"
}
```

After this PR (e.g. in 1.7.0), the public network cannot be found because the API requests to openstack now default to searching only non-external networks. 

As of 1.7.0, a public network would need to be looked up by adding an `external = true` attribute like this: 
```hcl
provider "openstack" {
  version     = "~> 1.7.0"
}

data "openstack_networking_network_v2" "external_network" {
  name = "public"
  external = true
}
```

However, attempting to use that configuration with earlier versions results in an error:
```
$ terraform plan
Error: data.openstack_networking_network_v2.external_network: : invalid or unknown key: external
```

PR #359, opened 2018-07-01
Unify gophercloud revision

Frassle:
Projects that use dep (https://github.com/golang/dep/) such as pulumi (https://github.com/pulumi) can't use terraform-provider-openstack as a dependency due to a limitation in dep.

dep only supports one revision per project (i.e. git repo). Currently terraform-provider-openstack uses a lots of different revisions from gophercloud depending on which package is being imported. 

vgo (https://github.com/golang/vgo), which looks to be the successor to dep and will be in go 1.11, also only supports one revision per module. By default the whole repo is the module, so projects using vgo would also not be able to depend on terraform-provider-openstack and it's use of multiple revisions of gophercloud.

This PR simply unifies them all to the same revision 19abc56a8cd8d3630ab9735173cb92c0bf635f33, and fixes up one type mismatch in openstack/resource_openstack_compute_instance_v2.go due to the update.

PR #361, opened 2018-07-03
Object Storage v1: ETag Changes

jtopjian:
Gophercloud now allows the etag to be specified and omitted. This caused
the openstack_objecstorage_object_v1 resource to break because it was
sending the old etag with new content. This commit fixes this by
creating a separate Update function and conditionally sending the etag.

This should not break existing deployments.

Related to: #359 

/cc @yanndegat

### DIFF
--- a/openstack/resource_openstack_objectstorage_object_v1.go
+++ b/openstack/resource_openstack_objectstorage_object_v1.go
@@ -14,9 +14,9 @@ import (
 
 func resourceObjectStorageObjectV1() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceObjectStorageObjectV1Put,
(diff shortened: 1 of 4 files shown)


jtopjian:
recheck

jtopjian:
recheck

jtopjian:
@fatmcgav This is ready for review at your convenience :)

Frassle:
Any eta for this as it blocks #359?

Frassle:
Thanks @fatmcgav :)

jtopjian:
Thanks, Gav! Coincidentally I just requested a release, which is probably a good idea so we can look at this with a clean slate. I'll merge this once the release is out.

PR #365, opened 2018-07-03
Return the default network uuid on instance create

Frassle:
Currently when an instance has no "network" blocks it's assumed it will
connect to the default network, worked out by inspecting the instance
addresses and taking the network name from that. This doesn't return the
network uuid, but the first refresh will pick the uuid up because the
resource now has a network block (with the default network name).

This commit simply copies some of the logic used in
getAllInstanceNetworks to work out network uuids from network names to
do the same with the default network name. Thus instances have a full
network block returned with name and uuid on create.

### DIFF
--- a/openstack/compute_instance_v2_networking.go
+++ b/openstack/compute_instance_v2_networking.go
@@ -415,6 +415,15 @@ func flattenInstanceNetworks(
 					"fixed_ip_v6": instanceNIC.FixedIPv6,
 					"mac":         instanceNIC.MAC,
 				}
+
+				// Use the same method as getAllInstanceNetworks to get the network uuid
+				networkInfo, err := getInstanceNetworkInfo(d, meta, "name", instanceAddresses.NetworkName)
+				if err != nil {
+					log.Printf("[WARN] Error getting default network uuid: %s", err)
+				} else {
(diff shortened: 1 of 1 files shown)


openstack/compute_instance_v2_networking.go
```diff
@@ -415,6 +415,14 @@ func flattenInstanceNetworks(
 					"fixed_ip_v6": instanceNIC.FixedIPv6,
 					"mac":         instanceNIC.MAC,
 				}
+
+				// Use the same method as getAllInstanceNetworks to get the network uuid
+				networkInfo, err := getInstanceNetworkInfo(d, meta, "name", instanceAddresses.NetworkName)
+				if err != nil {
```
jtopjian: I seem to recall there were some custom clouds out in the wild that did not provide this information to users. Given that, instead of returning an error, how about we just log the error and not set the UUID?

I think it's still a good situation to be in: most users will now have the additional information, users who can't obtain that information won't see any changes.

Frassle: Sure I actually thought the same thing while sleeping on it.

jtopjian:
recheck

Frassle:
Might need a rebase to pull in your floating ip fix.

PR #371, opened 2018-07-10
Allow one to obtain a specific floating IP to associate with a loadbalancer

rkettelerij:
Since a load balancer is often the entry point to the rest of your infrastructure it's common to use a stable/fixed IP as the VIP of your load balancer. The OpenStack REST API supports this, but the Terraform OpenStack provider currently doesn't. 

This PR adds the ability to obtain a _specific_ networking floating IP from the pool in order to associate it with the VIP of a load balancer. E.g:
```
resource "openstack_networking_floatingip_v2" "https_lb_float_ip" {
  pool = "mypool"
  address = "192.0.2.0"
}
```

Also credits to @jagter for helping out with this PR.

### DIFF
--- a/website/docs/r/networking_floatingip_v2.html.markdown
+++ b/website/docs/r/networking_floatingip_v2.html.markdown
@@ -42,6 +42,8 @@ The following arguments are supported:
     belongs to the same tenant. Changing this creates a new floating IP (which
     may or may not have a different address)
 
+* `address` - (Optional) The actual/specific floating IP to obtain.
+
 * `fixed_ip` - Fixed IP of the port to associate with this floating IP. Required if
     the port has multiple fixed IPs.
 
--- a/website/docs/r/networking_floatingip_v2.html.markdown
+++ b/website/docs/r/networking_floatingip_v2.html.markdown
@@ -42,7 +42,10 @@ The following arguments are supported:
     belongs to the same tenant. Changing this creates a new floating IP (which
     may or may not have a different address)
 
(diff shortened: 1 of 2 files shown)


rkettelerij:
recheck

jtopjian:
@rkettelerij Thanks for your work on this! I will be able to review it in detail in the next day or so.

website/docs/r/networking_floatingip_v2.html.markdown
```diff
@@ -42,6 +42,8 @@ The following arguments are supported:
     belongs to the same tenant. Changing this creates a new floating IP (which
     may or may not have a different address)
 
+* `address` - (Optional) The actual/specific floating IP to obtain.
```
jtopjian: I noticed that a default OpenStack environment prohibits non-admin users from specifying the floating IP. Either the user must be an admin or a custom policy must be configured in the networking service. I think we should make note of this. Possibly something like:

```
(Optional) The actual/specific floating IP to obtain. By default,
non-admin user are not able to specify a floating IP, so you
must either be an admin user or have had a custom policy or role
applied to your OpenStack user or project.
```

Thoughts?

rkettelerij: That's a great addition. LGTM.

jtopjian: @rkettelerij Cool. Do you want to follow-up with a commit when you have time?

jtopjian:
@rkettelerij Thank you again.

Issue #333, opened 2018-05-30
Add prefixlen to subnet resource


jtopjian:
As reported, the `openstack_networking_subnet_v2` resource does not support the `prefixlen` attribute.

This will first need added to the upstream Gophercloud library.

PR #376, opened 2018-07-23
Added prefixlen documentation

baderbuddy:
Documented how to change the size of the CIDR.

### DIFF
--- a/website/docs/r/networking_subnet_v2.html.markdown
+++ b/website/docs/r/networking_subnet_v2.html.markdown
@@ -95,6 +95,10 @@ The `host_routes` block supports:
 
 * `next_hop` - (Required) The next hop in the route.
 
+The `value_specs` block supports:
+
+* `prefixlen` - (Optional) The size of the CIDR to allocate from the subnetpool.
+
 ## Attributes Reference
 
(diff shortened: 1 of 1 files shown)


jtopjian:
@baderbuddy Thanks for submitting this.

I'm hesitant about merging it, though. `prefixlen` is first-class attribute that needs to be merged into the upstream Gophercloud library. There's a tracking issue for this here: https://github.com/terraform-providers/terraform-provider-openstack/issues/333

`value_specs` is a way of supplying key/value pairs that individual OpenStack clouds have added that are not official attributes.

You can also use `value_specs` to sneak missing official attributes into the API request. Some users do this as a workaround, and that's totally fine, however it's not something we want to officially document. If we did that, when official support came out for `prefixlen`, then it would end up breaking existing configurations.

Let me know if this makes sense and/or if you have any questions :)

baderbuddy:
Yeah that makes perfect sense to me, I couldn't find any documentation anywhere so it's great to know you're working on this. Thanks!

jtopjian:
No problem at all. Thanks again for making the doc change, though - it's still very appreciated.

I'm going to close this but still happy to discuss further.

PR #379, opened 2018-07-30
Add resource for creating Swift tempurls

flaper87:
Arguably, TempUrls could be considered a resource. However, they do not
exist in the server, they cannot be updated, and they cannot be
deleted. This type of resource can only be generated and it has a
limited lifetime. Therefore, this has been implemented as a datasource.

Given a container name and an object name, it'll be possible to create
a tempurl for that object and store it as a data source.

/cc @jtopjian

### DIFF
--- a/openstack/resource_openstack_objectstorage_tempurl_v1_test.go
+++ b/openstack/resource_openstack_objectstorage_tempurl_v1_test.go
@@ -1,0 +1,90 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccOpenStackObjectStorageTempurlV1_basic(t *testing.T) {
+	objectName := "object"
+	containerName := "container"
+	ttl := 60
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccOpenStackObjectstorageTempurlV1Resource_basic(containerName, objectName, "get", ttl),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckObjectstorageTempurlV1ResourceID("openstack_objectstorage_tempurl_v1.tempurl_1"),
+					resource.TestCheckResourceAttr(
+						"openstack_objectstorage_tempurl_v1.tempurl_1", "method", "get"),
+					resource.TestCheckResourceAttr(
+						"openstack_objectstorage_tempurl_v1.tempurl_1", "container", containerName),
+					resource.TestCheckResourceAttr(
+						"openstack_objectstorage_tempurl_v1.tempurl_1", "object", objectName),
+				),
+			},
+			resource.TestStep{
+				Config: testAccOpenStackObjectstorageTempurlV1Resource_basic(containerName, objectName, "post", ttl),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckObjectstorageTempurlV1ResourceID("openstack_objectstorage_tempurl_v1.tempurl_1"),
+					resource.TestCheckResourceAttr(
+						"openstack_objectstorage_tempurl_v1.tempurl_1", "method", "post"),
+				),
+			},
+			/* TODO(flaper87): Find a good way to test the ttl expiration
+			            resource.TestStep{
+							Config: testAccOpenStackObjectstorageTempurlV1Resource_basic(containerName, objectName, "get", ),
+							Check: resource.ComposeTestCheckFunc(
+								resource.TestCheckResourceAttr(
+									"openstack_objectstorage_tempurl_v1.tempurl_1", "method", "get"),
+								testAccCheckObjectstorageTempurlV1Expired("openstack_objectstorage_tempurl_v1.tempurl_1", 1),
+							),
+						},*/
+		},
+	})
+}
+
+func testAccCheckObjectstorageTempurlV1ResourceID(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Can't find temp url resource: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("Endpoint resource ID not set")
+		}
+
+		return nil
+	}
+}
+
+/*func testAccCheckObjectstorageTempurlV1Expired(n string, ttl int) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		time.Sleep(time.Duration(ttl))
+		err := testAccCheckObjectstorageTempurlV1ResourceID(n)(s)
+		if err == nil {
+			return fmt.Errorf("The temp url didn't expire")
+		}
+		return nil
+	}
+}*/
+
+func testAccOpenStackObjectstorageTempurlV1Resource_basic(container, object string, method string, ttl int) string {
+	return fmt.Sprintf(`
+	resource "openstack_objectstorage_tempurl_v1" "tempurl_1" {
+      object = "%s"
+      container = "%s"
+      method = "%s"
+      ttl = %d
+	}
+`, object, container, method, ttl)
+}
--- a/website/docs/r/objectstorage_tempurl_v1.html.markdown
+++ b/website/docs/r/objectstorage_tempurl_v1.html.markdown
@@ -1,0 +1,46 @@
+---
+layout: "openstack"
+page_title: "OpenStack: openstack_objectstorage_tempurl_v1"
+sidebar_current: "docs-openstack-resource-objectstorage-tempurl-v1"
+description: |-
+  Generate a TempURL for a Swift container and object.
+---
+
+# openstack\_objectstorage\_tempurl_v1
+
+Use this resource to generate an OpenStack Object Storage temporary URL.
+
+## Example Usage
+
+```hcl
+resource "openstack_objectstorage_tempurl_v1" "obj_tempurl" {
+  container = "test"
+  object = "container"
+  method = "post"
+  ttl = 20
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `container` - (Required) The container name the object belongs to.
+
+* `object` - (Required) The object name the tempurl is for.
+
+* `ttl` - (Required) The TTL, in seconds, for the URL. For how long it should be valid.
+
+* `method` - (Optional) What methods are allowed for this URL. Valid values are `GET`, and `POST`. Default is `GET`.
+
+* `region` - (Optional) The region the tempurl is located in.
+
+## Attributes Reference
+
+* `Id` - Computed md5 hash based on the generated url
+* `container` - See Argument Reference above.
+* `object` - See Argument Reference above.
+* `ttl` - See Argument Reference above.
+* `method` - See Argument Reference above.
+* `url` - The URL
+* `region` - The region the endpoint is located in.
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -254,6 +254,7 @@ func Provider() terraform.ResourceProvider {
 			"openstack_networking_subnetpool_v2":           resourceNetworkingSubnetPoolV2(),
 			"openstack_objectstorage_container_v1":         resourceObjectStorageContainerV1(),
 			"openstack_objectstorage_object_v1":            resourceObjectStorageObjectV1(),
+			"openstack_objectstorage_tempurl_v1":           resourceObjectstorageTempurlV1(),
 			"openstack_vpnaas_ipsec_policy_v2":             resourceIPSecPolicyV2(),
 			"openstack_vpnaas_service_v2":                  resourceServiceV2(),
 			"openstack_vpnaas_ike_policy_v2":               resourceIKEPolicyV2(),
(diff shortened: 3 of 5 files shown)


jtopjian:
@flaper87 Thanks!

This is certainly an interesting feature. It has characteristics of both a Data Source and a Resource. At the same time, it has characteristics that make it different than other Terraform data/resources. 😄 
 
Thinking out loud:

#### Creating

Since TempURLs have an API call to create a URL, that's a good indication that this should be a resource. The reason is because Data Sources are stateless and are regenerated upon each `plan` and `apply`. Even though `d.SetId` is being called, the ID is only stored for the duration of the plan/apply. 

With that in mind, if you had resources which consumed the generated URL:

```hcl
data "openstack_objectstorage_tempurl_v1" "foo" {
  ...
}

resource "some_other_resource" "bar" {
  url = "${data.openstack_objectstorage_tempurl_v1.foo.url}"
}
```

Each time you ran plan/apply, the "some_other_resource" resource would end up having "update" called because the URL would be different each time.

So if we made this a resource, then the URL would be stored in state and would persist through multiple apply's.

#### Reading

As you mentioned, TempURLs don't exist on the server, so we can't query/read them. Two notes about this:

1. This is another good indication that this shouldn't be a data source. Data sources are best used when querying something that already exists.
2. If we can only create a resource but not read it, that's OK. We can do something like this:

https://github.com/terraform-providers/terraform-provider-random/blob/master/random/resource_pet.go#L15

#### Updating

It's OK if we can't update the resource. That's common. One example of this is [`openstack_compute_keypair_v2`](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_keypair_v2.go). Notice how there is no "Update" function and all schema arguments are "ForceNew".

#### Deleting

This is where it gets difficult. 

If we can't call a `DELETE` API action of any kind, that's still OK. We can do something like this:

https://github.com/terraform-providers/terraform-provider-random/blob/master/random/resource_pet.go#L16

It will simply remove the TempURL resource from state.

The difficult part is when the TempURL expires. Let's say we create a TempURL for 24 hours and apply our configuration. Then 48 hours later we make an update that would use the temporary URL. More than likely we would get a HTTP error back and Terraform would fail.

I wonder if we can work around this. What if we added a compute-only field to the schema called `valid_until`. We would set the value of this field to be `time.Now + TTL`. 

Next, instead of using the [`schema.Noop`](https://github.com/terraform-providers/terraform-provider-random/blob/master/random/resource_pet.go#L15) method mentioned earlier, we had a very simple `Read` function that did something like this:

```go
v := d.Get("valid_until").(string)
validUntil, err := time.Parse("something", v)
now := time.Now()
if validUntil < now {
  d.SetId("")
}

return nil
```

When the ID is unset during a "Read", Terraform will automatically re-run Create. This is normally used to handle 404s (for example, a user deleted a resource out-of-band of Terraform), but we can unset the ID for any arbitrary reason.

#### Notes

* I'm not an expert with Swift's temporary URLs so I could be wrong about a lot of this.
* If it is OK to have lots of TempURLs generated that will only ever be used once, then maybe we can keep this a data source. But my gut is saying a resource is better. I could be wrong about this.

Again, a lot of this was me thinking about loud. This feature has some unique constraints for Terraform, so I'm not exactly sure what the best solution is. If you want to try rewriting this with the above ideas, go for it, but I don't mind doing that, either, since they're my suggestions and I have no idea if they would work 😄

flaper87:
@jtopjian I think your observations are all valid and they might work. 

As far as creating many tempurls goes, I think it's fine. I'd agree with you that not triggering an update on other resources every time a tempurl is generated would be nice. However, I'm not a fan of  the idea of re-implementing the ttl management for tempurls in the provider, it belongs to the server. 

How bad would it be to implement this as a `datasource` for now and, eventually, migrate it to a `resource` if we feel it's needed?

I'm happy to give your suggestions a go, fwiw.

jtopjian:
>  I'd agree with you that not triggering an update on other resources every time a tempurl is generated would be nice.

This is my main concern. 

I'm trying to understand the valid scenarios where this feature would be used.

If we accept that a temp URL is ephemeral and can only be used a single time, then we limit this feature to only being used one time, possibly to initially deploy some infrastructure. After that, the entire Terraform configuration can no longer be trusted in case the new temp URL triggers an entire rebuild (exaggerated worst case).

_unless_ the resource consuming the temp URL is intelligent enough to download and read the content of the URL and not just base diffs on the URL itself.

For example, this data source could provide temporary access to a bootstrap script or a container image that you want to deploy in another cloud (let's say AWS). The resource was successfully created and upon the next `apply`, even though a new temp URL was generated, the AWS resource checked the bootstrap scripts contents. Then no diff would happen and you're still able to provide temporary access to your object storage account.

Perhaps the [HTTP data source](https://www.terraform.io/docs/providers/http/data_source.html) can act as a buffer here.

Are these the kinds of scenarios you have in mind or were you thinking of something different?

> However, I'm not a fan of the idea of re-implementing the ttl management for tempurls in the provider, it belongs to the server.

I agree, but to confirm: this is a limitation of the TempURL API, right? If we wanted to manage the URL's full lifecycle, we would have to do it locally.

> How bad would it be to implement this as a datasource for now and, eventually, migrate it to a resource if we feel it's needed?

There's no migration path for a data source to a resource - they would be two distinct things in Terraform. We could deprecate and remove the data source over a period of time.

My concern is that if we release this as a data source initially and then release a resource, it's going to look like the objective wasn't entirely thought out. 

There's definitely no rush or race to get this released (at least from my point of view), so we can certainly take our time walking through it. It's always possible other users will post thoughts and comments, too.

flaper87:
> Are these the kinds of scenarios you have in mind or were you thinking of something different?

My scenario may fall into the category of scenarios that would be re-triggered because of the way it's implemented. I'm creating a container and uploading some configs there. I then create another data source that creates a text blob with some other data and includes the generated URL in it. This is stored as an output, which is then consumed by another module.

IIUC, both data source would be updated and so will be the output I'm depending on, which means that anything else depending on this output would be re-triggered.

I think I'm convincing myself that having a resource is probably better. Let me work on that today and update this PR.

> I agree, but to confirm: this is a limitation of the TempURL API, right? If we wanted to manage the URL's full lifecycle, we would have to do it locally.

Actually, I take what I said back. We can call `HEAD` to verify the validity but we also have everything we need. The TempURL has a `temp_url_expires` param with the expiration time. We just need to check that to verify the validity of the URL.

https://docs.openstack.org/swift/newton/middleware.html?highlight=tempurl#tempurl

> My concern is that if we release this as a data source initially and then release a resource, it's going to look like the objective wasn't entirely thought out.

we can always point them to this PR as a proof that there was some discussion. :+1: 

I'll try to get this done today so we can, hopefully, make it part of the next release.

jtopjian:
> I'll try to get this done today so we can, hopefully, make it part of the next release.

There's no rush. I put a request in for a new release earlier today.

We don't have a set release schedule, so it's totally possible to have a quick release after this pending release. We've got some other PRs queued, so it's looking like a short release cycle will happen anyway. Definitely take your time!

> we can always point them to this PR as a proof that there was some discussion.

Yes, good point. And on the topic of the discussion itself: thank you for your input and feedback with this!

jtopjian:
Thinking out loud again: if this was a resource, that means we could store the same URL forever in Terraform's state. By nature, TempURLs expire, so eventually that URL is going to be invalid.

For some use-cases, that might be OK. Maybe the temporary URL will only be used once and because the URL never changes, it won't trigger updates for other resources.

But for situations when the URL always needs to work, that means the URL will eventually need regenerated. 

Would it be worthwhile to have an schema field of something like `regenerate_url` which can control this? This way the user can expect a possible chain of events and prepare accordingly.

openstack/resource_openstack_objectstorage_tempurl_v1_test.go
```diff
@@ -0,0 +1,94 @@
...
+			return fmt.Errorf("Endpoint resource ID not set")
+		}
+
+		return nil
+	}
+}
+
+func testAccCheckObjectstorageTempurlV1Expired(n string, ttl int) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		time.Sleep(time.Duration(ttl))
```
flaper87: @jtopjian any recommendation on how to implement this test? I want to verify that the URL expires and the ID is unset. I guess the best way to implement it is getting the id before it expires and then after since the re-generation of the URL is triggered automatically. 

Thoughts?

flaper87: @jtopjian let mw know if you have any thoughts on this. 🤗

jtopjian: Will do - I'll be able to look at this later today.

jtopjian: Unfortunately I don't think Terraform's test framework will allow this. After each step, Terraform performs a refresh and will error if the resource has changed.

Here's what I came up with, which would work if it wasn't for the above:

https://github.com/jtopjian/terraform-provider-openstack/commit/9641999891d627b450fb206a2af4c37ee70ddb96

The fact that Terraform reports a diff is in itself an indication that the url expired and a new one was regenerated. I'll ask about this tomorrow to see if there's any way of telling Terraform this is correct behavior.

But as far as functionality goes, the expiration and re-creation works. :)

jtopjian: Yep, let's skip the test for now.

flaper87:
@jtopjian looks like the current CI error is not related to my patch. Am I correct?

jtopjian:
Correct. These failures are being caused in the master branch of Glance. We're tracking/fixing them in #390

flaper87:
recheck

jtopjian:
@flaper87 Not sure what happened with testing - it doesn't look like the acceptance test phase actually started.

In addition, you might have to rebase this on top of master. I don't think the patches from this PR are overlayed on top of master automatically during testing. This is something I've been meaning to verify.

flaper87:
recheck

openstack/resource_openstack_objectstorage_tempurl_v1_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func TestAccOpenStackObjectStorageTempurlV1_basic(t *testing.T) {
+	objectName := "object"
+	containerName := "container"
+	ttl := 60
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckAdminOnly(t)
```
jtopjian: We can remove this since a normal user can create temp URLs

website/docs/r/objectstorage_tempurl_v1.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+sidebar_current: "docs-openstack-resource-objectstorage-tempurl-v1"
+description: |-
+  Generate a TempURL for a Swift container and object.
+---
+
+# openstack\_objectstorage\_tempurl_v1
+
+Use this resource to get the ID of an OpenStack tempurl.
+
+Note: This usually requires admin privileges.
```
jtopjian: We can remove this note.

website/docs/r/objectstorage_tempurl_v1.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+layout: "openstack"
+page_title: "OpenStack: openstack_objectstorage_tempurl_v1"
+sidebar_current: "docs-openstack-resource-objectstorage-tempurl-v1"
+description: |-
+  Generate a TempURL for a Swift container and object.
+---
+
+# openstack\_objectstorage\_tempurl_v1
+
+Use this resource to get the ID of an OpenStack tempurl.
```
jtopjian: Maybe:

Use this resource to generate an OpenStack Object Storage temporary URL.

?

jtopjian:
@flaper87 I apologize for the OpenLab failures. The latest test failure is a test-case that has always randomly failed and I have not been able to figure out why. I've ended up removing it (#399), so rebasing and removing the "admin" checks should cause the tests to finally trigger and work.

Also, Travis is flagging an error due to the renamed documentation.

After all of that is resolved, this looks good to me and I think we can go ahead and merge it. It works, has basic functionality, and I would be curious to hear how it's received by other users.

I've been thinking of two additional features. Depending on your opinion we can either look at adding them now or possibly later:

1. Have an option to regenerate the URL

This would control whether or not the `d.SetId("")` actually happens. The benefit of this would be that users could use this resource as a one-time generation of a temporary URL and it'll never change after that, even after expiration.

2. Have a `triggers` schema item.

This is a trick used in a few different resources, for example [here](https://github.com/terraform-providers/terraform-provider-null/blob/master/null/resource.go#L25).

The idea is that if the value of `triggers` changes, it forces a refresh. There's no internal logic for this - it's handled entirely in the schema.

So let's say a user uploads an object and that object returns an etag. You then use the etag as input for the trigger. If the object changes, the etag changes, which triggers a new temporary url to be generated.

flaper87:
@jtopjian thanks for the review and sorry for the delay.

As for the new features, they sound interesting indeed. I think I can help you work on them but, if you don't mind, I'd prefer implementing them in separate PRs. That way we can isolate them more and keep this PR specifically to add the main functionality.

I like the idea of adding a `triggers` option quite a bit. I'll take a look after I get CI to pass. :D

flaper87:
@jtopjian GH still thinks there are requests that haven't been addressed but I think this PR is complete. Let me know what you think

jtopjian:
Awesome :)

>  GH still thinks there are requests that haven't been addressed

That's just Github being Github. I have to manually change the review to Approved.

I'll pressed for time at the moment but will give this another thorough review in the next day or two.

flaper87:
@jtopjian thanks a lot! Thanks for your time, reviews and patience :+1: 

When do you think we can get a new release? :smile:

jtopjian:
> When do you think we can get a new release?

I really want to get #409 and #412 merged, so as soon as someone can review/test them, the faster we can get a release out :)

PR #381, opened 2018-08-01
Image Service v2: Changes to Image Properties

jtopjian:
New versions of OpenStack are automatically adding a property
of os_hidden with a boolean value. This is conflicting with
Terraform's TypeMap which expects all values to be a string.

This commit fixes the issue by filtering out non-string
properties.


jtopjian:
This PR will fix the OpenLab errors we've been seeing.

It should be a safe and backwards-compatible change. I considered making `properties` a non-computed field but figured that might introduce breakage.

jtopjian:
@fatmcgav pending OpenLab success, let me know what you think.

jtopjian:
@fatmcgav All tests pass! Ready for review at your convenience. We might want to get this merged quickly so future tests pass.

Issue #382, opened 2018-08-01
using terraform-provider-openstack v1.7.0 causes a panic while handling data source openstack_networking_network_v2


jrandall:
### Terraform Version
```
$ terraform -v
Terraform v0.11.7
+ provider.openstack v1.7.0
```

### Affected Resource(s)
- data "openstack_networking_network_v2"

### Terraform Configuration Files
```hcl
provider "openstack" {
  version     = "~> 1.7.0"
}

data "openstack_networking_network_v2" "external_network" {
  name = "public"
}
```

### Debug Output
Sanitized debug output with `OS_DEBUG=1` when run with 1.7.0:
https://gist.github.com/jrandall/b009674038c434891386146beef8b6c5#file-tf_os_debug-1-7-0-log

Sanitized debug output with `OS_DEBUG=1` when run (successfully) with 1.6.0 (for comparison):
https://gist.github.com/jrandall/b009674038c434891386146beef8b6c5#file-tf_os_debug-1-6-0_working-log

### Panic Output
Sanitized `crash.log` written by `terraform plan` using terraform-openstack-provider 1.7.0:
https://gist.github.com/jrandall/b009674038c434891386146beef8b6c5#file-crash-log

### Expected Behavior
No changes were expected from this plan, and pinning terraform-provider-openstack version to v1.6.0 and re-running confirms everything still works fine with 1.6.0.

### Actual Behavior
Terraform panicked and exited with a crash.log

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Red Hat OpenStack Platform 12 - based on the Pike release of OpenStack.

The panic results from Gophercloud `extractIntoPtr` function on line 93 of `results.go`: https://github.com/terraform-providers/terraform-provider-openstack/blob/v1.7.0/vendor/github.com/gophercloud/gophercloud/results.go#L93

This panics because for whatever reason `label` (i.e. "networks") is not present in `m` and so `m[label]` is nil which results in a panic when the type assertion is made that insists it must be a slice. 

The main difference between v1.7.0 and v.1.6.0 problem appears to be that for some reason v1.7.0 is setting `router:external=false` in the v2.0 networks request to the openstack API (see https://gist.github.com/jrandall/b009674038c434891386146beef8b6c5#file-tf_os_debug-1-7-0-log-L383), so it then returns no entries as the network in question is an external one (however, I would in that case still expect terraform to handle the empty list that is returned without panic). 

[...]

PR #384, opened 2018-08-02
Networks v2: Network Data Source Fixes

jtopjian:
For #382 

#358 added support for querying networks by using the "external"
argument. However, this caused results to either contain only external
networks or only non-external networks.

This fixes the behavior to only query with the "external" argument if
it is explicitly defined in the Terraform configuration.

The introduction of the "external" Gophercloud package extracted the
network results in a new way. This new method did not safely handle
results which did not contain any networks.


jtopjian:
@fatmcgav Ready for review at your convenience

PR #385, opened 2018-08-02
Add identity group data source

BugRoger:
This commit adds a datasource for Groups.

### DIFF
--- a/openstack/data_source_openstack_identity_group_v3_test.go
+++ b/openstack/data_source_openstack_identity_group_v3_test.go
@@ -1,0 +1,50 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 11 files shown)


openstack/data_source_openstack_identity_group_v3_test.go
```diff
@@ -0,0 +1,50 @@
...
+			return fmt.Errorf("Group data source ID not set")
+		}
+
+		return nil
+	}
+}
+
+const testAccOpenStackIdentityV3GroupDataSource_basic = `
+data "openstack_identity_group_v3" "group_1" {
+    name = "admin"
```
jtopjian: A standard devstack installation creates a group called "admins". Even though our CI environment doesn't run admin tests right now, let's change admin to admins so it works in the future.

BugRoger:
Thanks for the quick reply. I added the documentation and renamed the tested group in the acceptance test.

PR #386, opened 2018-08-02
Image Service v2: Make Properties non-computed

jtopjian:
This commit changes the "properties" attribute to no longer be computed.
This is because OpenStack is adding an increasing amount of properties
to the image from the server-side which is causing conflicts with
user-generated properties.

A new attribute called "all_properties" has been added which will contain
all properties, both from the user and from the server.


jtopjian:
recheck

jtopjian:
@fatmcgav ready for review at your convenience.

jtopjian:
Actually there are a few things I want to try with this first.

PR #389, opened 2018-08-03
Create an openstack_config data source

flaper87:
This data source loads an openstack config to memory in the format
expected by the provider. The use case for this is reading, creating
and uploading configs to nodes that need to access a specific OpenStack
cloud.

The data source takes as input the same schema used for the OpenStack
provider and generates a map with all the required configs to
authenticate to an OpenStack cloud. From this map, it's possible to
render a new config and upload it to the final destination.

In addition, this patch adds the ability to disable the
`Authentication` validation performed every time the provider is
instantiated. This way we'd be able to check the provider's configs
just once in an entire workflow execution. This also allows for testing
the new config data source without having to have an actual cloud in
the background. Lazy validation, FTW.

We could also remove validation entirely and let gophercloud error out
at the first interaction with the openstack cloud.

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -40,6 +40,7 @@ type Config struct {
 	Username          string
 	UserID            string
 	useOctavia        bool
+	Validate          bool
 
 	OsClient *gophercloud.ProviderClient
 }
@@ -184,7 +185,7 @@ func (c *Config) LoadAndValidate() error {
 	}
 
 	// If using Swift Authentication, there's no need to validate authentication normally.
-	if !c.Swauth {
+	if !c.Swauth && c.Validate {
 		err = openstack.Authenticate(client, *ao)
 		if err != nil {
 			return err
--- a/openstack/data_source_openstack_config.go
+++ b/openstack/data_source_openstack_config.go
@@ -5,9 +5,17 @@ import (
 )
 
 func dataSourceOpenStackConfig() *schema.Resource {
+	providerSchema := getProviderSchema()
+
+	providerSchema["validate"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  "false",
+	}
+
 	return &schema.Resource{
 		Read:   dataSourceOpenStackConfigRead,
-		Schema: getProviderSchema(),
+		Schema: providerSchema,
 	}
 }
 
--- a/openstack/data_source_openstack_config.go
+++ b/openstack/data_source_openstack_config.go
@@ -1,0 +1,24 @@
+package openstack
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceOpenStackConfig() *schema.Resource {
+	return &schema.Resource{
+		Read:   dataSourceOpenStackConfigRead,
+		Schema: getProviderSchema(),
+	}
+}
+
+// dataSourceOpenStackConfigRead performs the endpoint lookup.
+func dataSourceOpenStackConfigRead(d *schema.ResourceData, meta interface{}) error {
+	config, err := configureProvider(d)
+	if err != nil {
+		return err
+	}
+
+	(config.(*Config)).LoadAndValidate()
+	d.SetId(d.Get("auth_url").(string))
+	return nil
+}
--- a/openstack/data_source_openstack_config_test.go
+++ b/openstack/data_source_openstack_config_test.go
@@ -1,0 +1,56 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccOpenStackOpenStackConfigDataSource_basic(t *testing.T) {
+	authURL := "http://10.0.0.3:5000"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckAdminOnly(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccOpenStackOpenStackConfigDataSource_basic(authURL),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckOpenStackConfigDataSourceID("data.openstack_config.config_1"),
+					resource.TestCheckResourceAttr(
+						"data.openstack_config.config_1", "auth_url", authURL),
+					resource.TestCheckResourceAttr(
+						"data.openstack_config.config_1", "swauth", "false"),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckOpenStackConfigDataSourceID(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Can't find config data source: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("Endpoint data source ID not set")
+		}
+
+		return nil
+	}
+}
+
+func testAccOpenStackOpenStackConfigDataSource_basic(authURL string) string {
+	return fmt.Sprintf(`
+	data "openstack_config" "config_1" {
+      auth_url = "%s"
+	}
+`, authURL)
+}
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -351,7 +351,13 @@ func configureProvider(d *schema.ResourceData) (interface{}, error) {
 		Username:          d.Get("user_name").(string),
 		UserID:            d.Get("user_id").(string),
 		useOctavia:        d.Get("use_octavia").(bool),
-		Validate:          d.Get("validate").(bool),
+		Validate:          true,
+	}
+
+	if v := d.Get("validate"); v != nil {
+		if validate, ok := v.(bool); ok {
+			config.Validate = validate
+		}
 	}
 
 	v, ok := d.GetOkExists("insecure")
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -9,182 +9,187 @@ import (
 // This is a global MutexKV for use within this plugin.
 var osMutexKV = mutexkv.NewMutexKV()
 
+func getProviderSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"auth_url": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc("OS_AUTH_URL", ""),
+			Description: descriptions["auth_url"],
+		},
+
+		"region": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: descriptions["region"],
+			DefaultFunc: schema.EnvDefaultFunc("OS_REGION_NAME", ""),
+		},
+
+		"user_name": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc("OS_USERNAME", ""),
+			Description: descriptions["user_name"],
+		},
+
+		"user_id": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc("OS_USER_ID", ""),
+			Description: descriptions["user_name"],
+		},
+
+		"tenant_id": &schema.Schema{
+			Type:     schema.TypeString,
+			Optional: true,
+			DefaultFunc: schema.MultiEnvDefaultFunc([]string{
+				"OS_TENANT_ID",
+				"OS_PROJECT_ID",
+			}, ""),
+			Description: descriptions["tenant_id"],
+		},
+
+		"tenant_name": &schema.Schema{
+			Type:     schema.TypeString,
+			Optional: true,
+			DefaultFunc: schema.MultiEnvDefaultFunc([]string{
+				"OS_TENANT_NAME",
+				"OS_PROJECT_NAME",
+			}, ""),
+			Description: descriptions["tenant_name"],
+		},
+
+		"password": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			Sensitive:   true,
+			DefaultFunc: schema.EnvDefaultFunc("OS_PASSWORD", ""),
+			Description: descriptions["password"],
+		},
+
+		"token": &schema.Schema{
+			Type:     schema.TypeString,
+			Optional: true,
+			DefaultFunc: schema.MultiEnvDefaultFunc([]string{
+				"OS_TOKEN",
+				"OS_AUTH_TOKEN",
+			}, ""),
+			Description: descriptions["token"],
+		},
+
+		"user_domain_name": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc("OS_USER_DOMAIN_NAME", ""),
+			Description: descriptions["user_domain_name"],
+		},
+
+		"user_domain_id": &schema.Schema{
+			Type:        schema.TypeString,
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc("OS_USER_DOMAIN_ID", ""),
+			Description: descriptions["user_domain_id"],
+		},
+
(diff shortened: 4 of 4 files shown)


flaper87:
missing docs, will add soon :)

jtopjian:
@flaper87 Thanks. This is another interesting PR.

> Not sure if this is the best way of doing this or if there's a way to gather this data from the provider itself. 

There's not a way for a resource to access a provider's data in this fashion. So in that regard, this feature makes sense. But I'm still trying to wrap my head around how it would be used. :)

Can you show more of a full example of how you see this data source being used? 

A few comments with regard to the current code:

When running acceptance tests for your PRs, are you using the provided `make` task?

```shell
$ cd $GOPATH/src/github.com/terraform-providers/terraform-provider-openstack
$ TF_LOG=DEBUG make testacc TEST=./openstack TESTARGS="-run=TestAccOpenStackOpenStackConfigDataSource_basic -count=1"
```

This PR will show an error when doing the above:

```
2018/08/04 23:41:35 [DEBUG] Starting graph walk: walkRefresh
panic: interface conversion: interface {} is nil, not bool
```

The reason for this is because:

```go
Validate:          d.Get("validate").(bool),
```

Since "validate" doesn't yet exist in the schema, `d.Get()` is returning `nil`.

Adding "validate" to the schema will resolve this and let the test pass, but if we moved forward with this PR, we'd want to make sure "validate" isn't in the actual provider schema. If it was, then a user can choose to not validate and thus not authenticate which would make the end-user experience kind of moot.

Perhaps something like this:

```diff
diff --git a/openstack/data_source_openstack_config.go b/openstack/data_source_openstack_config.go
index 6fd9ecd..372fb66 100644
--- a/openstack/data_source_openstack_config.go
+++ b/openstack/data_source_openstack_config.go
@@ -5,9 +5,17 @@ import (
 )

 func dataSourceOpenStackConfig() *schema.Resource {
+       providerSchema := getProviderSchema()
+
+       providerSchema["validate"] = &schema.Schema{
+               Type:     schema.TypeBool,
+               Optional: true,
+               Default:  "false",
+       }
+
        return &schema.Resource{
                Read:   dataSourceOpenStackConfigRead,
-               Schema: getProviderSchema(),
+               Schema: providerSchema,
        }
 }

diff --git a/openstack/provider.go b/openstack/provider.go
index b36a2b7..9bf460b 100644
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -349,7 +349,13 @@ func configureProvider(d *schema.ResourceData) (interface{}, error) {
                Username:          d.Get("user_name").(string),
                UserID:            d.Get("user_id").(string),
                useOctavia:        d.Get("use_octavia").(bool),
-               Validate:          d.Get("validate").(bool),
+               Validate:          true,
+       }
+
+       if v := d.Get("validate"); v != nil {
+               if validate, ok := v.(bool); ok {
+                       config.Validate = validate
+               }
        }

        v, ok := d.GetOkExists("insecure")
```

With regard to the word "Validate" itself, I wonder if we should call it "Authenticate" instead?

Thoughts?

flaper87:
> Can you show more of a full example of how you see this data source being used?

Sure thing!

My current use case is being able to generate a `cloud.conf` config file for an openstack cloud provider that I'm deploying with Terraform. Now, here are some of the reasons I thought about doing it through the provider rather than adding a whole bunch of new variables:

* If we want to re-use the same authentication credential, we'd have them already
* `terraform-provider-openstack` has the ability to read from `clouds.yaml` file, so it's possible to just pass the name of the cloud
* All the options are already defined and we can just re-use them
* If you're talking to an OpenStack cloud using terraform you likely have this provider anyway.

> Since "validate" doesn't yet exist in the schema, d.Get() is returning nil.

whooops, I actually had it and then forgot to add it back after a merge conflict. I had put it in the main schema but I like your idea of just putting it in the data source schema instead to avoid issues with the provider instance creation.

> With regard to the word "Validate" itself, I wonder if we should call it "Authenticate" instead?

Yes, I like that name better. I'm still wondering if calling Authenticate on every instantiation of the provider rather than doing it right before the actual HTTP request is done to the cloud.

If we decide to make the authentication step lazy, we could do away with this option entirely.

jtopjian:
Thank you for the details!

When you say you want to "create and upload configs to nodes that need to access a specific OpenStack cloud", how do you envision using this data source to create/write the data to the remote node? Do you plan on using a [template](https://www.terraform.io/docs/providers/template/index.html)?

I apologize for asking for some many details, I just want to make sure I understand the full use-case. To be clear, I think this data source is a valid idea - I just want to make sure there isn't some other way of handling it (for example: https://www.terraform.io/docs/modules/usage.html#providers-within-modules).

With regard to Lazy Loading, that's definitely a valid thing to support, but a few notes about this:

1. Terraform is calling `openstack.Authenticate` once during an apply, plan, destroy, etc. Terraform will _always_ do this, regardless of any existing resources. Meaning: if you're deploying a new set of resources that don't yet exist, Terraform will still call `openstack.Authenticate`.

You can see this in action by doing the following

```shell
$ TF_LOG=DEBUG OS_DEBUG=1 terraform <action>
```

Note the combination of `TF_LOG` + `OS_DEBUG`. `OS_DEBUG` is a special feature of this provider that many other Terraform providers don't have. It's been invaluable for debugging the wide range of OpenStack deployments in the wild.

2. Terraform will create service clients multiple times, usually one per resource. If the resources do not yet exist in Terraform's state, no client will be created. 

With this in mind, "resources" can be thought of as already being lazy loaded, which is good. I believe what you're proposing is to make `openstack.Authenticate` only ever run when a lazy loaded resource is run, right? Just want to make sure we're on the same page here.

If we are, then this is actually a difficult thing to do. Imagine you have a simple configuration like the following:

```hcl
provider "openstack" {
  cloud = "devstack"
}

resource "openstack_compute_keypair_v2" "foo" {
  name = "foo"
}

resource "openstack_compute_keypair_v2" "bar" {
  name = "bar"
}
```

Terraform, being concurrent by nature, will create both key pairs in parallel. If the Provider is lazy loaded, which resource will trigger the actual loading of the Provider? 

In order to do this, we need to introduce a good amount of [sync](https://golang.org/pkg/sync/) activity. 

This is possible to do and I have experience [doing this](https://github.com/sl1pm4t/terraform-provider-lxd/pull/127) but it's really tricky and can lead to random crashes. Using the LXD provider as an example again, a [new bug](https://github.com/sl1pm4t/terraform-provider-lxd/issues/137) has recently started happening even though it never appeared since April.

So with that in mind, the type of authentication bypassing you're proposing with this data source is valid and possible to do. But doing true lazy loading is a whole other feature we'll want to tackle in a separate PR. And to be honest, I would be really cautious/hesitant on adding it until, say, a v2 release given the amount of new bugs it could uncover. 

I apologize for the amount of text :)

flaper87:
> When you say you want to "create and upload configs to nodes that need to access a specific OpenStack cloud", how do you envision using this data source to create/write the data to the remote node? Do you plan on using a template?

In my current tests, I've been using a template a template and then uploading this config file to the openstack instances either using a swift object or the userdata. The full workflow I have now is:

1. Create an OpenStack instance
2. Upload these configs
3. Install Kubernetes on this instance with the OpenStack provider enabled. 

All the above happens at the first boot time. It's a bit more complex than what I've detailed above but I hope that gives enough info about the usecase.

> Terraform, being concurrent by nature, will create both key pairs in parallel. If the Provider is lazy loaded, which resource will trigger the actual loading of the Provider?

This is the part I was missing about Terraform's behavior. With this in mind, I think it doesn't make much sense to actually change this code right now. The extra complexity of the sync process is not worth the benefits it'd bring to this data source. I'd rather add the `validate` option to the data source only and set it to `false` by default.

If we do so, I'd keep the name validate as it'd be specific for the data source and not a global option. I'll do these changes and push a new commit to see what it looks like and what you think about it.

> I apologize for the amount of text :)

No worries at all. Thanks a lot for taking the time to explain how Terraform and this provider work. Your patience is extremely appreciated.

kiwik:
recheck

flaper87:
recheck

jtopjian:
@flaper87 I believe the test failures are still due to the `d.Get("validate")` in the `provider.go` file. Try adding the second part of the patch I mentioned above.

> In my current tests, I've been using a template and then uploading this config file to the openstack instances either using a swift object or the userdata.

Thanks!

This makes sense to me and I think this data source is the right approach to do that. 

I think the important part is that the deployed config file is part of an entirely separate/decoupled Terraform configuration, so there's really no other way of doing variable interpolation or sharing the provider information around like you would if this was all part of one large Terraform configuration.

I'm going to ask around and see if there are any other ways of doing this just to make sure.

Side note: have you looked into [backends](https://www.terraform.io/docs/backends/types/index.html) yet? There is a Swift backend, but it needs some updates and doesn't yet support locking so I'm not sure if this would work in your situation.

But the idea is that you can publish "outputs" in your state and retrieve them in other configurations. 

For example:

```hcl
# file foo/main.tf

data "openstack_networking_network_v2" "public" {
  name = "public"
}

output "public_uuid" {
  value = "${data.openstack_networking_network_v2.public.id}"
}

# file bar/main.tf

data "terraform_remote_state" "foo" {
  backend = "local"

  config {
    path = "../foo/terraform.tfstate"
  }
}

output "public_uuid" {
  value = "${data.terraform_remote_state.foo.public_uuid}"
}
```
> If we do so, I'd keep the name validate as it'd be specific for the data source and not a global option. 

Sounds good to me.

jtopjian:
@flaper87 I had a really good chat with some other Terraform developers today about this.

First, the provider data is not available by any other means. So using a data source to obtain this information is the best way to go. There is even some precedence for this which I'll explain below.

However, there's concern about exposing _all_ of the provider data. The concern is that having the full set of credentials leads to mismanagement of the credentials, circumventing better means.

For example, let's say you also transfer a `clouds.yaml` file to the remote node. You can use this data source to self-discover the name of the cloud and then transfer the name to the remote node. Either doing this by template or by a remote/local provisioner would make sense:

```hcl
provisioner "remote-exec" {
  inline = [
    "OS_CLOUD=${data.openstack_config.foo.cloud} terraform apply",
  ]
}
```

But embedding all of the information like username, password, auth_url, etc is most likely a bad idea since you're now relying on Terraform to not only manage this information but produce further Terraform files that also manage it.

At least with using `clouds.yaml` you can leverage a number of other tools to securely manage the storage/creation of the yaml file. 

With regard to a precedence, there are a few AWS data sources which do similar things, but they only expose small portions of the provider data. For example, the [`aws_region`](https://www.terraform.io/docs/providers/aws/d/region.html) data source will return the region used in the overarching provider.

With this in mind, do you think it's possible to limit the exposed data to only the "cloud"? Or is there other information from the provider which should also be exposed - and if so, are they related or would they work best as small data sources?

Thoughts?

flaper87:
> However, there's concern about exposing all of the provider data. The concern is that having the full set of credentials leads to mismanagement of the credentials, circumventing better means.

I agree with this statement but we're not really passing all the provider's data. In fact, we're not passing any of the provider's data. What we're doing is using the provider's schema to build this data source that must receive specific arguments from the user. The user may choose the pass the same arguments but as for the provider, and that should be fine.

> With this in mind, do you think it's possible to limit the exposed data to only the "cloud"? Or is there other information from the provider which should also be exposed - and if so, are they related or would they work best as small data sources?

To be honest, in my use case, I need to be able to pass the credentials for the cloud, not just the cloud name. Let me try to explain this differently. If this data source wouldn't exist, I'd have added a new set of variables (the same as the ones for the provider) that the user would have to pass in order to create to pass this credentials to the node. Considering we already have the logic to parse a `clouds.yaml` file, we already request some of them, etc, I think it'd be better to use that and provider an easier, more familiar, way to do this. Whether we add new variables or use this data source, terraform will be writiing and managing new files and this data.

I don't really have an option as far as adding this data to the node goes. I need to create a file with this credentials in the node to be able to use the software I'm installing in the node.

I think this data source provides a better experience for consumers of this provider, it's consistent and, as it's implemented, I don't think it leaks any data. The data needs to be explicitely set/requested by the user. 

Hope this information helps, let me know if you have more questions :+1:

jtopjian:
Ah, I see. The intent is to not re-use values of the provider block. You're just looking to re-use an existing data structure rather than either having to redesign your own using ad-hoc variables or copying/pasting the existing provider schema. Correct?

With that in mind...

> we're not passing any of the provider's data

There is one case when data will be passed implicitly: when environment variables are being used for authorization.

From the output of the acceptance test, you can see something like:

```
data.openstack_config.config_1:
  ID = http://10.0.0.3:5000
  provider = provider.openstack
  auth_url = http://10.0.0.3:5000
  default_domain = default
  password = password
  project_domain_id = default
  region = RegionOne
  swauth = false
  tenant_name = demo
  use_octavia = false
  user_domain_name = Default
  user_name = demo
  validate = false
```

When environment variables are not being used, then this data source acts as (what I believe is) the intended data structure.

But now if the intent is to not actually retrieve data, but only to reduce repeated boilerplate code of OpenStack configuration settings, this might end up being confusing to users. I'm having a hard time coming up with an accurate description for this.

If I was an average OpenStack user with access to a single project and I saw this data source, how would it be useful to me? Or how can it be described in a way which would immediately tell me what use-cases it was applicable to?

Not to diminish the work done here, but if this might be a niche use-case, it's possible to create a separate Terraform Provider to support this (Terraform Provider OpenStackConfig or something like that). We would need to merge some of the changes done in this PR as well as a few other minor changes to allow Go packages outside of if this provider to import the data, but that's also doable. I've done a quick proof-of-concept and it's definitely possible.

flaper87:
> There is one case when data will be passed implicitly: when environment variables are being used for authorization.

I think we can make some of the fields in the structure required only for the data source just to force the user to pass in the parameters instead of depending on the env variables. That said, I had thought about this case when working on the data source and I didn't think it was a big deal if documented well.

If I'd like to use the same credentials that I've set in my environment and generate this data source, I should be able to do it. It'll provide a consistent experience.

(As a side note, I'm not a fan of the fact that this provider supports parsing env variables. It's dangerous and fragile.)

> But now if the intent is to not actually retrieve data, but only to reduce repeated boilerplate code of OpenStack configuration settings, this might end up being confusing to users. I'm having a hard time coming up with an accurate description for this.

The data source does retrieve and manages data. It creates a structure that is then used by other resources, just like other data sources do. The difference is that it retrieves the data from the local box and helps in managing the credentials present there. I see it as a `clouds.yaml` template utility, if that means anything :smile: 

> If I was an average OpenStack user with access to a single project and I saw this data source, how would it be useful to me?

It doesn't have to be useful to everyone. Admittedly, it has a more specific use case but, to be fair, every data source and resource have a specific use case.

> Or how can it be described in a way which would immediately tell me what use-cases it was applicable to?

It's a data source to retrieve OpenStack credentials from the local box that may be required by other terraform resources. It reads `clouds.yaml` files or environment variables.

> Not to diminish the work done here, but if this might be a niche use-case, it's possible to create a separate Terraform Provider to support this (Terraform Provider OpenStackConfig or something like that). We would need to merge some of the changes done in this PR as well as a few other minor changes to allow Go packages outside of if this provider to import the data, but that's also doable. I've done a quick proof-of-concept and it's definitely possible.

To be honest, I'd rather not. I don't think we need a separate provider. Having one would be an overkill for such a simple feature. I don't find this data source to be confusing or dangerous. I see it as a utility that is part of the provider.

Another option would be to merge it as is, let it evolve and see if it requires its own provider in the future. By creating a new provider we're not really solving much but rather moving the maintenance burden somewhere else, which is not to underestimate. I'm a huge fan of maintaining the least amount of code necessary :smiley:

jtopjian:
> It's a data source to retrieve OpenStack credentials from the local box that may be required by other terraform resources. It reads clouds.yaml files or environment variables.

Except for the environment variable piece, this isn't the behavior I'm seeing. This is why I'm so confused about this data source -- it's not actually pulling from a source of data.

If I create a Terraform file like the following:

```hcl
data "openstack_config" "foo" {
  auth_url = "https://example.com:5000/v3"
  cloud    = "some-cloud"
}
```

This is what I get back:

```shell
$ terraform init
$ terraform apply
$ terraform show

data.openstack_config.foo:
  id = https://example.com:5000/v3
  auth_url = https://example.com:5000/v3
  cloud = some-cloud
  default_domain = default
  swauth = false
  use_octavia = false
  validate = false
```

The `id` is set to the `auth_url`. This makes `auth_url` required, but that's not important right now. `auth_url` and `cloud` are set because they were both explicitly defined in the data source. The rest of the output is from the default values of the schema.

Nothing else was populated. The `some-cloud` entry in `clouds.yaml` wasn't read or its corresponding values in this data source populated.

Is there missing functionality that still needs to be added?

PR #393, opened 2018-08-06
Allow openstack_lb_member_v2 to have a weight of 0

### LINKED ISSUE #392: LBaaS Pool Member should allow a weight of 0
The terraform openstack provider does not allow LBaaS pool members with a weight of 0.

A weight of 0 is a valid value and should be supported by the terraform openstack provider.

See the OpenStack documentation, which specifies that 0 is a valid value:
* [create member](https://developer.openstack.org/api-ref/load-balancer/v2/index.html#create-member)
* [update member](https://developer.openstack.org/api-ref/load-balancer/v2/index.html#update-a-member)

> The weight of a member determines the portion of requests or connections it services compared to the other members of the pool. For example, a member with a weight of 10 receives five times as many requests as a member with a weight of 2. _A value of 0 means the member does not receive new connections but continues to service existing connections_. A valid value is from 0 to 256. Default is 1.

> Setting the member weight to 0 means that the member will not receive new requests but will finish any existing connections. This “drains” the backend member of active connections.

### Terraform Version
terraform 0.11.7 
terraform-provider-openstack 1.7.0

### Affected Resource(s)
`openstack_lb_member_v2`

```
resource "openstack_lb_member_v2" "member" {
  pool_id       = "${var.pool_id}"
  subnet_id     = "${var.subnet_id}"
  address       = "${var.address}"
  protocol_port = "${var.port}"
  weight        = "0"
}
```

### Expected Behavior

The terraform openstack provider should allow setting a pool member's weight to 0

### Actual Behavior

The terraform openstack provider returns the following error:

```
Error: openstack_lb_member_v2.member: Only numbers greater than 0 are supported values for 'weight'
```

The error comes from [here](https://github.com/terraform-providers/terraform-provider-openstack/blob/5e6c44989b04278c1ee24c50fbadd044de9d8175/openstack/resource_openstack_lb_member_v2.go#L63-L70)

### Steps to Reproduce

Attempt to plan or apply a `openstack_lb_member_v2` resource with a weight of 0

philsttr:
Fixes #392

### DIFF
--- a/openstack/resource_openstack_lb_member_v2.go
+++ b/openstack/resource_openstack_lb_member_v2.go
@@ -62,9 +62,9 @@ func resourceMemberV2() *schema.Resource {
 				Computed: true,
 				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 					value := v.(int)
-					if value < 1 {
+					if value < 0 || value > 256 {
 						errors = append(errors, fmt.Errorf(
-							"Only numbers greater than 0 are supported values for 'weight'"))
+							"Only numbers between 0 and 256 (inclusive) are supported values for 'weight'"))
 					}
 					return
 				},
--- a/website/docs/r/lb_member_v2.html.markdown
+++ b/website/docs/r/lb_member_v2.html.markdown
@@ -45,10 +45,12 @@ The following arguments are supported:
 * `protocol_port` - (Required) The port on which to listen for client traffic.
     Changing this creates a new member.
 
-* `weight` - (Optional)  A positive integer value that indicates the relative
-    portion of traffic that this member should receive from the pool. For
-    example, a member with a weight of 10 receives five times as much traffic
-    as a member with a weight of 2.
+* `weight` - (Optional)  An integer value between 0 and 256 (inclusive) that
+    indicates the relative portion of traffic that this member should receive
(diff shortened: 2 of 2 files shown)


jtopjian:
@philsttr Thank you for reporting this issue and working on a fix.

Are you sure this patch actually resolves the issue? `CreateMemberOpts` and `UpdateMemberOpts` is defined with `omitempty` which will cause zero-values to be ignored in the request:

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/pools/requests.go#L277

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/pools/requests.go#L326

This is most likely why the valid values are only restricted to greater than 0 right now.

Because of this, we're going to have to first patch Gophercloud and then do a vendor update as part of this PR.

philsttr:
Hey  @jtopjian 

Nope.  This looked like a simple fix.  I didn't notice the gophercloud config.

Thanks for giving this some attention.

jtopjian:
@philsttr No problem. I can get Gophercloud patched, but this PR might be delayed as we're trying to sort out some other vendoring cleanup right now.

jtopjian:
@philsttr This was implemented in #451. Please let me know if you have any questions :)

PR #396, opened 2018-08-07
Updating Developer Documentation

jtopjian:


website/docs/index.html.markdown
```diff
@@ -189,32 +199,208 @@ add any additional comments, please let us know.
...
+$ make testacc TEST=./openstack TESTARGS="-run=TestAccComputeV2Keypair_basic -count=1"
+```
+
+We recommend running tests with logging set to `DEBUG`:
+
+```shell
+$ TF_LOG=DEBUG make testacc TEST=./openstack TESTARGS="-run=TestAccComputeV2Keypair_basic -count=1"
+```
+
+And you can even enable OpenStack debugging to see the API traffic:
```
ozerovandrei: Maybe we can use different word instead of`traffic`? It sounds like we're looking straight into some tcpdump output 😅 
I guess that `requests` or `HTTP requests` is better.

jtopjian: Good idea.

jtopjian:
@ozerovandrei Thanks - let me know how this update looks.

jtopjian:
recheck

PR #399, opened 2018-08-08
Compute v2: Remove Attach to Second Network Test

jtopjian:
This test regularly fails in our CI environment for unknown reasons.
Testing locally has always been successful.


jtopjian:
recheck

PR #401, opened 2018-08-10
Add networking router v2 data source

rasta-rocket:


openstack/data_source_openstack_networking_router_v2.go
```diff
@@ -0,0 +1,156 @@
...
+	if v, ok := d.GetOk("router_id"); ok {
+		listOpts.ID = v.(string)
+	}
+
+	if v, ok := d.GetOk("name"); ok {
+		listOpts.Name = v.(string)
+	}
+
+	if v, ok := d.GetOk("admin_state_up"); ok {
+		listOpts.AdminStateUp = v.(*bool)
```
jtopjian: This causes a crash. You can verify this by creating a test which sets `admin_state_up` to true.

I suggest doing the following:

```go
if v, ok := d.GetOkExists("admin_state_up"); ok {
  asu := v.(bool)
  listOpts.AdminStateUp := &asu
}
```

openstack/data_source_openstack_networking_router_v2.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	if v, ok := d.GetOk("distributed"); ok {
+		listOpts.Distributed = v.(*bool)
```
jtopjian: same here

openstack/data_source_openstack_networking_router_v2_test.go
```diff
@@ -0,0 +1,62 @@
...
+resource "openstack_networking_router_v2" "router" {
+  name = "router_tf"
+}
+`
+
+var testAccOpenStackNetworkingRouterV2DataSource_name = fmt.Sprintf(`
+%s
+
+data "openstack_networking_router_v2" "router" {
+   name = "${openstack_networking_router_v2.router.name}"
```
jtopjian: Let's add:

```
admin_state_up = true
distributed = false
```

here to ensure the test does not crash.

jtopjian:
@rasta-rocket oops - this is my mistake. It looks like simply specifying `distributed`, even if the value is `false` will trigger a policy error:

```
Error creating OpenStack Neutron router: Request forbidden: [POST http://10.3.0.27:9696/v2.0/routers], error message: {"NeutronError": {"message": "(rule:create_router and rule:create_router:distributed) is disallowed by policy", "type": "PolicyNotAuthorized", "detail": ""}}
```

rasta-rocket:
@jtopjian yeah I saw that, but it is not related to the READ (GET) request but with the CREATE (POST) since I had that field in the terraform **resource**, so I just remove it from the resource in the test.
Waiting the CI ;)

jtopjian:
Perfect - good call 😄

rasta-rocket:
\o/

openstack/data_source_openstack_networking_router_v2.go
jtopjian: @rasta-rocket I apologize - I didn't catch this earlier today when you updated the PR.

Instead of `GetOk`, we want to use `GetOkExists`. `GetOk` doesn't work very well for boolean fields as it doesn't register if the user specified "false" as a value. You can see this if you run the acceptance tests with `OS_DEBUG=1` set.

`GetOk`:

```
2018/08/14 03:49:42 [DEBUG] OpenStack Request URL: GET http://neutron:9696/v2.0/routers?admin_state_up=true&name=router_tf
```

`GetOkExists`:

```
2018/08/14 03:48:16 [DEBUG] OpenStack Request URL: GET http://neutron:9696/v2.0/routers?admin_state_up=true&distributed=false&name=router_tf
```

rasta-rocket: Good catch indeed

rasta-rocket:
@jtopjian I think the CI get stuck on my side :confused:

jtopjian:
@rasta-rocket OpenLab can take 2-3 hours to run sometimes.

rasta-rocket:
@jtopjian I think the current job (and only job that I saw) which is running is regarding PR #407 and not the current one.
Am I wrong ? :thinking:

jtopjian:
recheck

jtopjian:
It's running now.

rasta-rocket:
Thx :+1:

jtopjian:
recheck

openstack/data_source_openstack_networking_router_v2.go
```diff
@@ -0,0 +1,158 @@
...
+		listOpts.Status = v.(string)
+	}
+
+	if v, ok := d.GetOk("tenant_id"); ok {
+		listOpts.TenantID = v.(string)
+	}
+
+	pages, err := routers.List(networkingClient, listOpts).AllPages()
+	if err != nil {
+		fmt.Errorf("Unable to list Routers: %s", err)
```
jtopjian: This should be:

```go
return fmt.Errorf("Unable to list Routers: %s", err)
```

It took me a while to spot this 😄 

The OpenLab test failure is actually legitimate. All tests were passing for me locally, so that usually means something has changed in the upcoming release of OpenStack. 

I created a devstack environment and was then able to reproduce the test failure:

```
2018/08/15 04:12:32 [DEBUG] OpenStack Response Body: {
  "NeutronError": {
    "detail": "",
    "message": "[u'distributed'] is invalid attribute for filtering",
    "type": "HTTPBadRequest"
  }
}
```

So two things:

1. Looks like we have to remove the previously added `distributed = "false"` from the test fixture. While it's a legitimate option for most versions of OpenStack, something must have changed in the upcoming release to make it an invalid query parameter. We'll still keep `distributed` a valid schema field, though. And at least we know the `GetOkExists` stuff works now :)
2. Add `return` to the  `fmt.Errorf...` line above to properly catch the error.

rasta-rocket: :+1:

Issue #402, opened 2018-08-12
OpenStack Container Infrastructure Management (Magnum) support


ozerovandrei:
Terraform-provider-openstack should support OpenStack Magnum as discussed in the upstream library project:
[Gophercloud#1003](https://github.com/gophercloud/gophercloud/issues/1003)

Progress on the upstream Magnum API implementation is tracked in:
[Gophercloud#1062](https://github.com/gophercloud/gophercloud/issues/1062)

This issue tracks Magnum resources and data sources implementation in the `terraform-provider-openstack`.

- [ ] Certificates resource
- [ ] Cluster template resource
- [ ] Cluster template datasource
- [ ] Cluster resource
- [ ] Cluster datasource

ozerovandrei:
@jtopjian I've been thinking of how to implement a `resource_openstack_containerinfra_certificate_v1` and it seems that it can't be implemented at the current state of the Magnum API (at least in the way that it can be useful).

Each Magnum cluster has the root CA (that is created by Magnum at the initial cluster creation).
Then each user (if he owns this cluster) can create its own certificate with the provided CSR but Magnum API doesn't track their uuids or their count whatsoever.
User only can read current root CA of a cluster but he can't see how many custom certificates were issued. He can only rotate the root CA so all created custom certificates will be disabled.

This logic is implemented in the [via following requests](https://developer.openstack.org/api-ref/container-infrastructure-management/?expanded=#manage-certificates-for-bay-cluster):
 - GET `/v1/certificates/cluster_uuid` - shows root CA of a cluster
 - POST `/v1/certificates` - creates user certificate with provided custom CSR
 - PATCH `/v1/certificates/cluster_uuid` - rotates root CA of a cluster

So, If we create a resource that issues new custom certificates we won't be able do implement a delete method that won't rotate all custom certificates of a cluster. We also can't implement a proper read method (we will only get root CA data). And I don't know what we can use for the ID of that resource.

Also I checked remaining Magnum API handlers in the [Gophercloud#1062](
https://github.com/gophercloud/gophercloud/issues/1062) and I see that there is no other stuff that we can implement in the Terraform provider.
First, there are `Bay` and `Baymodels` but we don't really need them because they're were renaimed into `Cluster` and `Clustertemplate` and they shouldn't work after `Neutron` release.
[...]

PR #403, opened 2018-08-12
ContainerInfraV1: add Cluster template resource

ozerovandrei:
Add resource to manage OpenStack Magnum cluster templates.
Add common resourceLinks() function to unmarhal slice of gophercloud.Links to a slice of maps.

For #402

Some links:
[Constraints for the `apiserver_port`](https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L82)
[Requirement of the `coe`](https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L50)
[Requirement of the `image`](https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L53)

### DIFF
--- a/openstack/resource_openstack_containerinfra_clustertemplate_v1.go
+++ b/openstack/resource_openstack_containerinfra_clustertemplate_v1.go
@@ -313,8 +313,13 @@ func resourceContainerInfraClusterTemplateV1Read(d *schema.ResourceData, meta in
 	d.Set("name", s.Name)
 	d.Set("project_id", s.ProjectID)
 	d.Set("user_id", s.UserID)
-	d.Set("created_at", s.CreatedAt)
-	d.Set("updated_at", s.UpdatedAt)
+
+	if err := d.Set("created_at", s.CreatedAt); err != nil {
+		log.Printf("[DEBUG] created_at: %s", err)
+	}
+	if err := d.Set("updated_at", s.CreatedAt); err != nil {
+		log.Printf("[DEBUG] updated_at: %s", err)
+	}
 
 	return nil
 }
--- a/openstack/resource_openstack_containerinfra_clustertemplate_v1.go
+++ b/openstack/resource_openstack_containerinfra_clustertemplate_v1.go
@@ -312,6 +312,7 @@ func resourceContainerInfraClusterTemplateV1Read(d *schema.ResourceData, meta in
 
 	log.Printf("[DEBUG] Retrieved Clustertemplate %s: %#v", d.Id(), s)
 
+	d.Set("apiserver_port", s.APIServerPort)
 	d.Set("coe", s.COE)
 	d.Set("cluster_distro", s.ClusterDistro)
 	d.Set("dns_nameserver", s.DNSNameServer)
@@ -344,15 +345,6 @@ func resourceContainerInfraClusterTemplateV1Read(d *schema.ResourceData, meta in
 	d.Set("user_id", s.UserID)
 	d.Set("created_at", s.CreatedAt)
 	d.Set("updated_at", s.UpdatedAt)
-
-	// Set apiserver_port.
-	if s.APIServerPort != "" {
-		apiServerPort, err := strconv.Atoi(s.APIServerPort)
-		if err != nil {
-			return fmt.Errorf("Error setting Cluster template API server port: %v", s.APIServerPort)
-		}
-		d.Set("apiserver_port", apiServerPort)
-	}
 
 	return nil
 }
@@ -474,8 +466,8 @@ func resourceContainerInfraClusterTemplateV1Delete(d *schema.ResourceData, meta 
 
 func validateClusterTemplateAPIServerPortV1(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(int)
-	if value < 1024 {
-		err := fmt.Errorf("%s should be greater or equal to 1024", k)
+	if value < 1024 || value > 65535 {
+		err := fmt.Errorf("%s should be between 1024 and 65535", k)
 		errors = append(errors, err)
 	}
 	return
--- a/openstack/resource_openstack_containerinfra_clustertemplate_v1.go
+++ b/openstack/resource_openstack_containerinfra_clustertemplate_v1.go
@@ -152,25 +152,6 @@ func resourceContainerInfraClusterTemplateV1() *schema.Resource {
 				Optional: true,
 				ForceNew: false,
 			},
-			"links": &schema.Schema{
-				Type: schema.TypeList,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"href": &schema.Schema{
-							Type:     schema.TypeString,
-							ForceNew: false,
-							Computed: true,
-						},
-						"rel": &schema.Schema{
-							Type:     schema.TypeString,
-							ForceNew: false,
-							Computed: true,
-						},
-					},
-				},
-				ForceNew: false,
-				Computed: true,
-			},
 			"master_lb_enabled": &schema.Schema{
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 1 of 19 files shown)


ozerovandrei:
It's almost ready, I just want to resolve to more issues:

1) Right now the cluster template resource gets the `labels` attribute as a string of key-value attributes (as we can do via `openstack coe cluster template create --labels kube_tag =1.11.1, prometheus_monitoring =true,...`). I think that this will be uncomfortable for users to see such long lines in their hcl manifests.
I want to make an input as a map:
```hcl
labels = {
    kube_tag = "1.11.1"
    prometheus_monitoring = "true"
    influx_grafana_dashboard_enabled = "true"
    kube_dashboard_enabled = "true"
}
```
And then convert it to the string inside a provider resource.

2) Second issue I can't fix right now and I need some help.
So, Magnum API gets UpdateOpts as a slice of add/replace/delete values: [doc.go#L71](https://github.com/gophercloud/gophercloud/blob/master/openstack/containerinfra/v1/clustertemplates/doc.go#L71).
I tried to use the replace action on every attribute that is changed (checking that with the usual `d.HasChange()`) and use the delete action if new value of the attribute is an empty string.
But for some reason `d.HasChange` ignores new empty value. (So i just use the replace action for now).
I will do some digging about it but maybe someone can point me in the right direction? @jtopjian do you know some workarounds for it?

jtopjian:
> But for some reason d.HasChange ignores new empty value.

It shouldn't. `HasChange` should notify whenever any sort of change to a previously set argument happens.

Can you provide an example that reproduces this behavior?

ozerovandrei:
For example, we created the following cluster template:
```hcl
resource "openstack_containerinfra_clustertemplate_v1" "clustertemplate_1" {
  name = "clustertemplate_1"
  image = "Fedora-Atomic-27"
  coe = "kubernetes"
  http_proxy = "localhost:8800"
}
```

And after that we want to set the `http_proxy` attribute to `""`.

jtopjian:
The reason is because `http_proxy` is set to `Computed`. This example is timely because I've recently begun questioning and being strict in PRs about whether an argument really needs to be Computed or not. There have been too many bugs introduced because of this.

Unless the Magnum API sets a default proxy, it does not need to be computed.

The reason `HasChange` isn't triggering a change is because although you have removed the value, Terraform thinks Magnum is going to send a computed value.

Try this out by making `http_proxy` not computed.

A better example might be `project_id` and `user_id`. Let's assume an admin user can set both of these to assign a template to a different project. So an admin would run Terraform with these values set:

```hcl
resource "openstack_containerinfra_clustertemplate_v1" "clustertemplate_1" {
  name = "clustertemplate_1"
  image = "Fedora-Atomic-27"
  coe = "kubernetes"
  http_proxy = "localhost:8800"

  project_id = "some-other-project"
  user_id = "some-other-user"
}
```

But if these arguments are not set, then the API will return the project and user who created the template. 

A good rule of thumb for determining if an argument should be computed is: who is authoritative of the data: the user or API service? If a user can set, change, and remove a value, then _usually_ (there are, of course, some exceptions) the argument should not be computed. If a user can set and change a value, but the API can set a default value, then computed should be fine. If the user _cannot_ set a value and only the API sets it, then it is a "read-only" argument (see the data sources for examples of this).

Make sense? If so, you can see how misuse of "computed" can lead to bugs. :(

One common misconception about "computed" arguments is that you cannot use `d.Set()` unless the value is computed. This isn't true. I think this dates back to before Terraform had "import" support. Back then, you would only run `d.Set` on computed fields because the user defined everything else, so why do `d.Set` again? But with importing, you are emulating an existing tf configuration, so you must set non-computed fields, too.

ozerovandrei:
@jtopjian this is awesome explanation, thank you!

ozerovandrei:
This is ready.
Although, we need the `OpenStack Magnum` installed in the OpenLab acceptance tests runner.

jtopjian:
@ozerovandrei Cool! I hope to be able to review this in the next few days. My schedule has been really busy lately.

openstack/resource_openstack_containerinfra_clustertemplate_v1.go
```diff
@@ -0,0 +1,491 @@
...
+	d.Set("registry_enabled", s.RegistryEnabled)
+	d.Set("server_type", s.ServerType)
+	d.Set("tls_disabled", s.TLSDisabled)
+	d.Set("volume_driver", s.VolumeDriver)
+	d.Set("region", GetRegion(d, config))
+	d.Set("name", s.Name)
+	d.Set("project_id", s.ProjectID)
+	d.Set("user_id", s.UserID)
+	d.Set("created_at", s.CreatedAt)
+	d.Set("updated_at", s.UpdatedAt)
```
jtopjian: For both of these, you'll need to format them into a string. If you do:

```go
if err := d.Set("created_at", s.CreatedAt); err != nil {
  log.Printf("[DEBUG] created_at: %s", err)
}
```

You'll see an error message in the logs. In addition, you'll notice that these are not being set in state, either.

Alternatively, we can just remove these fields unless you see a need to have them.

jtopjian:
@ozerovandrei also, with regard to the test failure, we'll want to add something similar to this:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider_test.go#L114

Then when OpenLab adds Magnum support, they can set the environment variable on their side.

ozerovandrei:
@jtopjian Thank you for the review.

I've tried the [IntBetween](https://github.com/hashicorp/terraform/blob/v0.11.7/helper/validation/validation.go#L18) validation function and got the following problem:
 - we need to vendor `github.com/hashicorp/terraform/helper/validation`
 - if we add it to vendor, terraform-provider-openstack will be broken, it will throw panic:
```sh
panic: gob: registering duplicate types for "github.com/zclconf/go-cty/cty.primitiveType": cty.primitiveType != cty.primitiveType

goroutine 1 [running]:
encoding/gob.RegisterName(0xc42003c7b0, 0x2b, 0x12c7d20, 0xc420290937)
	/usr/lib/go-1.10/src/encoding/gob/type.go:820 +0x5d4
github.com/hashicorp/terraform/vendor/github.com/zclconf/go-cty/cty.init.3()
	/opt/stack/go/src/github.com/hashicorp/terraform/vendor/github.com/zclconf/go-cty/cty/types_to_register.go:52 +0x5bf
exit status 2
FAIL	github.com/terraform-providers/terraform-provider-openstack/openstack	0.010s
```
 - we can fix this, by fetching the latest full `github.com/hashicorp/terraform` with all of its dependencies.

Can I do such big vendoring fetching inside this PR?

jtopjian:
Did you do the vendoring with a v0.11.7 tag?

If so, yuck. You're right - we definitely don't want to do this large of an update in this PR. We're going to have to do a vendor update when Terraform 0.12 comes out, so let's hold off until then.

Thank you for looking into this.

ozerovandrei:
Of course not 😄 
Yeah, it's working with the `v0.11.7` tag, all good.

ozerovandrei:
This can be reviewed again.
Hope that I resolved all issues.

openstack/resource_openstack_containerinfra_clustertemplate_v1.go
```diff
@@ -0,0 +1,522 @@
...
+	d.Set("name", s.Name)
+	d.Set("project_id", s.ProjectID)
+	d.Set("user_id", s.UserID)
+
+	if err := d.Set("created_at", s.CreatedAt); err != nil {
+		log.Printf("[DEBUG] created_at: %s", err)
+	}
+	if err := d.Set("updated_at", s.CreatedAt); err != nil {
+		log.Printf("[DEBUG] updated_at: %s", err)
+	}
```
jtopjian: If you look in the debug logs, you'll see something similar to:

```
2018/08/30 15:57:42 [DEBUG] created_at: created_at: '' expected type 'string', got unconvertible type 'time.Time'
```

So we need to do:

```
d.Set("created_at", s.CreatedAt.Format(time.RFC3339))
```

ozerovandrei:
recheck

jtopjian:
@ozerovandrei LGTM!

I want to get #409 merged first since it's a bug fix and has priority. Once that's merged, feel free to merge this one when you're ready :)

ozerovandrei:
@jtopjian got it.

jtopjian:
@ozerovandrei looks like there's a vendor conflict - I figured that might happen.

Once it's resolved, feel free to merge whenever you're ready.

After it's merged, you can edit the CHANGELOG file directly through Github to add the release note. Let me know if you have any questions.

Issue #301, opened 2018-04-27
openstack_images_image_v2 indicates changes every time due to read-only computed property


jrandall:
### Terraform Version
terraform 0.11.2, terraform-provider-openstack 1.3.0

### Affected Resource(s)
- openstack_images_image_v2

### Terraform Configuration Files
```hcl
resource "openstack_images_image_v2" "hgi-image" {
  name             = "${var.env}-${var.region}-${var.setup}-${var.image_name}"
  image_source_url = "${var.image_url_base}/${var.image_name}"
  container_format = "${var.image_container_format}"
  disk_format      = "${var.image_disk_format}"

  properties {
    user = "beastie"
  }
}
```

### Expected Behavior
Image property `user` should be set and on subsequent runs, nothing should be changed.

### Actual Behavior
Property is set, but resource attempts to change every time to remove a read-only property `direct_url`:
```
Terraform will perform the following actions:

  ~ module.hgi-openstack-image-hgi-base-freebsd11-4cb02ffa.openstack_images_image_v2.hgi-image
      properties.direct_url:      "rbd://049fc780-8998-45a8-be12-d3b8b6f30e69/zeta-images/803a17db-8023-460f-9ea0-8f778b0f66d5/snap" => ""
      properties.user:            "" => "beastie"
```

### Important Factoids
Our openstack environment has glance backed by Ceph, and the images v2 schema (see https://developer.openstack.org/api-ref/image/v2/index.html#image-schemas) includes a read only property called `direct_url` that represents the backend storage location of the image data. 

terraform-provider-openstack does not seem to take the schema into account when determining which properties need updating, and as a result it is not possible to set any properties without causing terraform to indicate a change on every invocation. 

terraform-provider-openstack could use the image schemas API to check which properties are read-only, or perhaps we could add a provider or resource configuration option that would provide a list of read-only image properties that should be ignored when updating images?

jtopjian:
Thanks for reporting this.

> terraform-provider-openstack could use the image schemas API to check which properties are read-only, or perhaps we could add a provider or resource configuration option that would provide a list of read-only image properties that should be ignored when updating images?

Probably the latter, most likely in the form of an `all_properties` attribute that would list combined properties from the Terraform user and from other sources.

In order to support the former, the schema part of the Glance API would need added to Gophercloud first.

jrandall:
[...]

Issue #289, opened 2018-04-17
openstack_images_image_v2 does not set property


moio:
### Terraform Version
Latest 0.11.7, with latest openstack provider 1.3.

### Affected Resource(s)

- openstack_images_image_v2

### Terraform Configuration Files

Relevant parts pasted below:

```hcl
resource "openstack_images_image_v2" "images" {
  name = "${element(var.images, count.index)}"
  image_source_url = "${var.image_locations["${element(var.images, count.index)}"]}"
  count = "${length(var.images)}"
  container_format = "bare"
  disk_format = "qcow2"
  properties {
    hw_rng_model = "virtio"
  }
}```

Variables are defined as follows:

```hcl
variable "images" {
  description = "list of images to be uploaded to Glance, leave default for all"
  default = ["centos7", "opensuse423"]
  type = "list"
}

variable "image_locations" {
  description = "list of locations to download images, override to add custom ones"
  default = {
    centos7 = "http://w3.nue.suse.com/~smoioli/sumaform-images/openstack/centos7.qcow2"
    opensuse423 = "https://download.opensuse.org/repositories/systemsmanagement:/sumaform:/images:/openstack/images/opensuse423.x86_64.qcow2"
  }
  type = "map"
}
```

### Debug Output
https://gist.github.com/moio/cf242b87b84ca2aa186ec08b1fb96f80

### Expected Behavior

Image property `hw_rng_model` should be set to `virtio`, at next terraform plan nothing should be done.

### Actual Behavior

Image property `hw_rng_model` is not set and next terraform plan signals the change again.

Note that setting the property via the `openstack` client works, provided it is specified in single quotes:

```
openstack image set --property hw_rng_model='virtio' <UUID>
```

### Steps to Reproduce
1. `terraform apply`

jtopjian:
@moio Thank you for reporting this. I _think_ I fixed this issue upstream a few days ago, coincidentally. I should have a fix for this shortly.

moio:
If there is any patch I could test I am happy to volunteer verifying them! Thanks!

jtopjian:
@moio I was unable to reproduce the issue you're seeing with the config you provided, but I've opened #292 which _might_ fix the problem. 

If you're able to test it out, that would definitely help! Let me know if you need help compiling it.

jtopjian:
@moio Version 1.4 of this provider was just released. Let me know if it helps at all or if you're still seeing this issue. :)

[...]

PR #409, opened 2018-08-17
Image Service v2: Support Updating Properties

jtopjian:
This commit enables full addition, modification, and removal
of properties during an image update.

In addition, the "direct_url" property will automatically be
added to the list of all properties if it is detected.

For #289 
For #301

### DIFF
--- a/openstack/resource_openstack_images_image_v2.go
+++ b/openstack/resource_openstack_images_image_v2.go
@@ -348,39 +348,33 @@ func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if d.HasChange("properties") {
-		old, new := d.GetChange("properties")
-		oldProperties := resourceImagesImageV2ExpandProperties(old.(map[string]interface{}))
-		newProperties := resourceImagesImageV2ExpandProperties(new.(map[string]interface{}))
+		o, n := d.GetChange("properties")
+		oldProperties := resourceImagesImageV2ExpandProperties(o.(map[string]interface{}))
+		newProperties := resourceImagesImageV2ExpandProperties(n.(map[string]interface{}))
 
 		// Check for new and changed properties
 		for newKey, newValue := range newProperties {
-			var found bool
 			var changed bool
-			for oldKey, oldValue := range oldProperties {
-				if newKey == oldKey {
-					found = true
-				}
 
-				if newValue != oldValue {
-					changed = true
-				}
+			oldValue, found := oldProperties[newKey]
+			if found && (newValue != oldValue) {
+				changed = true
+			}
 
-				// os_ keys are provided by the OpenStack Image service.
-				// These are read-only properties that cannot be modified.
-				// Ignore them here and let CustomizeDiff handle them.
-				if strings.HasPrefix(newKey, "os_") {
-					found = true
-					changed = false
-				}
+			// os_ keys are provided by the OpenStack Image service.
+			// These are read-only properties that cannot be modified.
+			// Ignore them here and let CustomizeDiff handle them.
+			if strings.HasPrefix(newKey, "os_") {
+				found = true
+				changed = false
+			}
 
-				// direct_url is provided by some storage drivers.
-				// This is a read-only property that cannot be modified.
-				// Ignore it here and let CustomizeDiff handle it.
-				if newKey == "direct_url" {
-					found = true
-					changed = false
-				}
-
+			// direct_url is provided by some storage drivers.
+			// This is a read-only property that cannot be modified.
+			// Ignore it here and let CustomizeDiff handle it.
+			if newKey == "direct_url" {
+				found = true
+				changed = false
 			}
 
 			if !found {
@@ -405,13 +399,8 @@ func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error
 		}
 
 		// Check for removed properties
-		for oldKey, _ := range oldProperties {
-			var found bool
-			for newKey, _ := range newProperties {
-				if oldKey == newKey {
-					found = true
-				}
-			}
+		for oldKey := range oldProperties {
+			_, found := newProperties[oldKey]
 
 			if !found {
 				v := images.UpdateImageProperty{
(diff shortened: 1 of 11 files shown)


jtopjian:
Some notes about this for posterity:

It was discovered that `properties` was originally implemented as updateable argument, but the update implementation was never done. Fixing this was not possible at the time because Gophercloud did not support updating properties.

To resolve this error, `properties` was modified to force a new resource when there was an update to properties. This happened in the 1.8.0 release. `ForceNew` should have been set initially on this resource, but that was a bug which slipped by.

The intention was to eventually get support into Gophercloud to update properties. This happened sooner than expected and `ForceNew` will go away for the 1.9.0 release.

In addition to all of this, our testing service caught a change in the upcoming version of OpenStack which was adding some read-only properties to each image. This was causing perpetual diffs, which meant our CI was failing, and was causing important PRs to become queued and delayed.

One solution would have been adding an `all_properties` argument which contained both the user-provided and server-provided properties while keeping `properties` strictly reserved for user-provided properties. This pattern has worked in the past, but it has always felt like a workaround more than a proper solution. Since we know what properties the server will be providing, we should be able to reconcile them in the `properties` argument and have everything work out. This was accomplished using a `CustomizeDiff` function and released in 1.8.0.

This PR retains this functionality.

With the merge of this PR, the `openstack_images_image_v2` resource should provide the user with the original intended behavior of being able to fully manage properties throughout the lifecycle of the image. In addition, it should safely handle integrating certain server-provided properties.

moio:
Time for me to try this PR for #289?

jtopjian:
> Time for me to try this PR for #289?

If you have the time, I would totally appreciate that 😄

openstack/resource_openstack_images_image_v2.go
```diff
@@ -348,6 +347,83 @@ func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error
 		updateOpts = append(updateOpts, v)
 	}
 
+	if d.HasChange("properties") {
+		old, new := d.GetChange("properties")
+		oldProperties := resourceImagesImageV2ExpandProperties(old.(map[string]interface{}))
+		newProperties := resourceImagesImageV2ExpandProperties(new.(map[string]interface{}))
```
ozerovandrei: It may sounds strange but I won't recommend using Go's `new` keyword: [effective_go.html#allocation_new](https://golang.org/doc/effective_go.html#allocation_new)

What about using `o, n := d.GetChange("properties")` like in `terraform` code: [resource_data.go#L82](https://github.com/hashicorp/terraform/blob/master/helper/schema/resource_data.go#L82)?

openstack/resource_openstack_images_image_v2.go
```diff
@@ -348,6 +347,83 @@ func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error
...
+					Name:  newKey,
+					Value: newValue,
+				}
+
+				updateOpts = append(updateOpts, v)
+			}
+		}
+
+		// Check for removed properties
+		for oldKey, _ := range oldProperties {
```
ozerovandrei: You don't need to use `oldKey, _` if you're ignoring map values, you can just do `for oldKey := range oldProperties`.

openstack/resource_openstack_images_image_v2.go
```diff
@@ -348,6 +347,83 @@ func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error
...
+			var found bool
+			for newKey, _ := range newProperties {
```
ozerovandrei: Same as L#408: `for newKey := range newProperties`.

openstack/resource_openstack_images_image_v2.go
```diff
@@ -348,6 +347,83 @@ func resourceImagesImageV2Update(d *schema.ResourceData, meta interface{}) error
...
+				if oldKey == newKey {
+					found = true
+				}
+			}
+
```
ozerovandrei: I think we don't need to run over `newProperties` for every `oldKey`. It will be `O(n^2)` after all.

Can't we just use something like this:
```
for oldKey, _ := range oldProperties {
  _, found := newProperties[oldKey]
  if !found {
    v := images.UpdateImageProperty{
      Op:   images.RemoveOp,
      Name: oldKey,
    }

    updateOpts = append(updateOpts, v)
  }
}
```

I guess that we can even improve code in L#359 using that trick.

openstack/resource_openstack_images_image_v2.go
```diff
@@ -573,34 +649,18 @@ func resourceImagesImageV2UpdateComputedAttributes(diff *schema.ResourceDiff, me
 			// new = user properties only
 			old, new := diff.GetChange("properties")
 
-			oldProperties := resourceImagesImageV2ExpandProperties(old.(map[string]interface{}))
 			newProperties := resourceImagesImageV2ExpandProperties(new.(map[string]interface{}))
```
ozerovandrei: Same question about using `new` here.

ozerovandrei: Same question about using `new`.

website/docs/r/images_image_v2.html.markdown
```diff
@@ -57,7 +57,8 @@ The following arguments are supported:
 * `name` - (Required) The name of the image.
 
 * `properties` - (Optional) A map of key/value pairs to set freeform
-    information about an image.
+    information about an image. See the "Notes" section for further
```
ozerovandrei: Extra spaces at the beginning break indentation.

jtopjian: Not sure I understand. The indentation isn't consistent with some of the other blocks, but that's normal. Most other provider docs have a mixture of indentations.

jtopjian:
@ozerovandrei Another really good review - thank you! I have one question about the indentation, but I think I've addressed everything else.

website/docs/r/images_image_v2.html.markdown
ozerovandrei: Other attributes with multi-line description indent in the following way:
```
* `image_cache_path` - (Optional) This is the directory where the images will
   be downloaded. Images will be stored with a filename corresponding to
   the url's md5 hash. Defaults to "$HOME/.terraform/image_cache"
```

But here you have extra space before line:
```
* `properties` - (Optional) A map of key/value pairs to set freeform
    information about an image. See the "Notes" section for further
    information about properties.
```

But yeah, I checked other attributes and they're using different type of indentaion. So I guess it doesn't matter and final html page will be the same.

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
recheck

PR #412, opened 2018-08-21
 Object Storage v1: Add regenerate to temp url resource 

jtopjian:
This commit adds the `regenerate` argument to the
`openstack_objectstorage_tempurl_v1` resource. This allows users to
opt-in to having the resource automatically regenerate when the
temporary URL has expired.

### DIFF
--- a/openstack/resource_openstack_objectstorage_tempurl_v1.go
+++ b/openstack/resource_openstack_objectstorage_tempurl_v1.go
@@ -67,6 +67,12 @@ func resourceObjectstorageTempurlV1() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"regenerate": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+			},
+
 			"url": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -134,8 +140,10 @@ func resourceObjectstorageTempurlV1Read(d *schema.ResourceData, meta interface{}
 		return fmt.Errorf("Failed to parse the temp url expiration time: %s", qp.Get("temp_url_expires"))
 	}
 
+	// Regenerate the URL if it has expired and if the user requested it to be.
+	regen := d.Get("regenerate").(bool)
 	now := time.Now().Unix()
-	if expiry < now {
+	if expiry < now && regen {
 		log.Printf("[DEBUG] URL expired, generating a new one")
 		d.SetId("")
 	}
--- a/openstack/resource_openstack_objectstorage_tempurl_v1.go
+++ b/openstack/resource_openstack_objectstorage_tempurl_v1.go
@@ -108,11 +108,11 @@ func resourceObjectstorageTempurlV1Create(d *schema.ResourceData, meta interface
 	containerName := d.Get("container").(string)
 	objectName := d.Get("object").(string)
 
-	log.Printf("[DEBUG] Create TempURL Options: %#v", turlOptions)
+	log.Printf("[DEBUG] Create temporary url Options: %#v", turlOptions)
 
 	url, err := objects.CreateTempURL(objectStorageClient, containerName, objectName, turlOptions)
 	if err != nil {
-		return fmt.Errorf("Unable to generate a TempURL for the object %s in container %s: %s",
+		return fmt.Errorf("Unable to generate a temporary url for the object %s in container %s: %s",
 			objectName, containerName, err)
 	}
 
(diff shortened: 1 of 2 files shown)


website/docs/r/objectstorage_tempurl_v1.html.markdown
```diff
@@ -25,19 +32,25 @@ resource "openstack_objectstorage_tempurl_v1" "obj_tempurl" {
...
 * `container` - (Required) The container name the object belongs to.
 
 * `object` - (Required) The object name the tempurl is for.
 
-* `ttl` - (Required) The TTL, in seconds, for the URL. For how long it should be valid.
+* `ttl` - (Required) The TTL, in seconds, for the URL. For how long it should
+  be valid.
 
-* `method` - (Optional) What methods are allowed for this URL. Valid values are `GET`, and `POST`. Default is `GET`.
+* `method` - (Optional) What methods are allowed for this URL.
```
ozerovandrei: Shouldn't we use `What method is allow for this URL`? We're passing a single method here, right?

website/docs/r/objectstorage_tempurl_v1.html.markdown
```diff
@@ -10,6 +10,13 @@ description: |-
 
 Use this resource to generate an OpenStack Object Storage temporary URL.
 
+The temporary URL will be valid for as long as TTL is set to (in seconds).
+Once the URL has expired, it will no longer be valid, but the resource
+will remain in place. If you wish to automatically regenerate a URL, set
+the `regenerate` argument to `true`. This will create a new resource with
+a new ID and URL.
+
+
```
ozerovandrei: There was a single blank line before. This extra line is a mistake, right?

openstack/resource_openstack_objectstorage_tempurl_v1.go
```diff
@@ -140,3 +125,28 @@ func resourceObjectstorageTempurlV1Create(d *schema.ResourceData, meta interface
 	d.Set("url", url)
 	return nil
 }
+
+// resourceObjectstorageTempurlV1Read performs the image lookup.
+func resourceObjectstorageTempurlV1Read(d *schema.ResourceData, meta interface{}) error {
+	turl := d.Get("url").(string)
+	u, err := url.Parse(turl)
+	if err != nil {
+		return fmt.Errorf("Failed to read the temp url: %s", turl)
```
ozerovandrei: Don't we need to use an error string here instead of provided url? Like that:
```
return fmt.Errorf("Failed to read the temp url: %v", err)
```

Also, I think that using `temporary URL` instead of `temp url` is better in log messages because we're using that term in the resource documentation.

openstack/resource_openstack_objectstorage_tempurl_v1.go
```diff
@@ -140,3 +125,28 @@ func resourceObjectstorageTempurlV1Create(d *schema.ResourceData, meta interface
...
+	}
+
+	qp, _ := url.ParseQuery(u.RawQuery)
+	expiry, err := strconv.ParseInt(qp.Get("temp_url_expires"), 10, 64)
+	if err != nil {
+		return fmt.Errorf("Failed to parse the temp url expiration time: %s", qp.Get("temp_url_expires"))
```
ozerovandrei: Same questions as in L#134.
Also, I think that it's better to do `qp.Get("temp_url_expires")` only a single time.
And why are we ignoring the `url.ParseQuery` error?

openstack/resource_openstack_objectstorage_tempurl_v1.go
```diff
@@ -140,3 +125,28 @@ func resourceObjectstorageTempurlV1Create(d *schema.ResourceData, meta interface
...
+	}
+
+	// Regenerate the URL if it has expired and if the user requested it to be.
+	regen := d.Get("regenerate").(bool)
+	now := time.Now().Unix()
```
ozerovandrei: I think it should be in UTC: [temporary_url_middleware.html#temporary-url-format](https://docs.openstack.org/swift/latest/api/temporary_url_middleware.html#temporary-url-format)

jtopjian: Good catch, but epoch time is timezone agnostic:

https://play.golang.org/p/Crgq5U5x0bJ

I think the docs can be interpreted this way:

"An expiration date as a UNIX Epoch timestamp _or_ ISO 8601 UTC timestamp. "

jtopjian:
@ozerovandrei Nice review. I think I've addressed all comments.

jtopjian:
recheck

openstack/resource_openstack_objectstorage_tempurl_v1.go
ozerovandrei: Oh, right.

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
recheck

PR #413, opened 2018-08-23
When calling API endpoints, allow a certain amount of retries

PedroSantosDataminded:
Hello! 

We are working with a cloud provided whose API endpoints sometimes timeout.
The current behavior is for the terraform-provider-openstack to return an error and abort terraform.

This pull request adds 5 retries to the API calls. After, it fails as before.


jtopjian:
@PedroSantosDataminded Thank you for submitting this. I understand the use-case and the code looks good. 

I wonder if this should be something that's more opt-in? Such as a provider-level option called `enable_retries` or something like `request_retries` where the user could specify the number of retries to make. By default, if the user didn't provide an option, then the behavior would act as it does now: no retrying and Terraform will error.

Thoughts?

PedroSantosDataminded:
@jtopjian Thank you for your feedback. I agree with you, the best way is to make it a `request_retries` configuration option. By default, the value is zero, meaning the current behavior. I'll get back to you as soon as possible with the updated code.

PedroSantosDataminded:
Took a bit longer that in should, but now this pull has `max_retries` as a configuration parameter.

openstack/provider.go
```diff
@@ -329,6 +336,8 @@ func init() {
 			"service (Octavia) instead of the Networking service (Neutron).",
 
 		"cloud": "An entry in a `clouds.yaml` file to use.",
+
+		"max_retries": "How many times an http connection should be retried until giving up",
```
ozerovandrei: How many times HTTP connection should be retried until giving up

openstack/types.go
```diff
@@ -64,8 +65,23 @@ func (lrt *LogRoundTripper) RoundTrip(request *http.Request) (*http.Response, er
 	}
 
 	response, err := lrt.Rt.RoundTrip(request)
-	if response == nil {
-		return nil, err
+
+	//retrying connection
```
ozerovandrei: // Retrying connection.

openstack/types.go
```diff
@@ -64,8 +65,23 @@ func (lrt *LogRoundTripper) RoundTrip(request *http.Request) (*http.Response, er
...
+	retry := 1
+	for response == nil {
+
+		if retry > lrt.MaxRetries {
+			if lrt.OsDebug {
+				log.Printf("[DEBUG] Openstack connection error, retries exhausted. Aborting")
```
ozerovandrei: log.Printf("[DEBUG] Openstack connection error, after %d retries last error was: %s", retry, err)

openstack/types.go
```diff
@@ -64,8 +65,23 @@ func (lrt *LogRoundTripper) RoundTrip(request *http.Request) (*http.Response, er
...
+			}
+			return nil, err
+		}
+
+		if lrt.OsDebug {
+			log.Printf("[DEBUG] Openstack connection error, retry number %d", retry)
```
ozerovandrei: log.Printf("[DEBUG] Openstack connection error: %s, retry number %d", err, retry)

domdom82:
bump this. would be really nice to get this merged. we are constantly running into keystone errors when the system comes under load. being a little more tolerant like the AWS provider would be great.
@PedroSantosDataminded thanks for the PR, I was about to write the same :)

openstack/types.go
jtopjian: I've gone back and forth on this both ways.

I'll issue a tie breaker and say let's not have the duplicate error logged. If I'm reading this correctly, the log entries would look like:

```
Openstack connection error, retry number 1: err
Openstack connection error, retry number 2: err
Openstack connection error, retry number 3: err
Openstack connection error, retries exhausted. Aborting
```

Which I think is fine.

But while we're on the topic, these debug logs are great, but users don't normally with debug logging enabled. Is it possible to modify the error being returned to something like:

```
err = fmt.Errorf("Openstack connection error, retries exhausted. Aborting. Last error was: %s", err)
```

This way, the user knows both that the max retries were attempted and that there was an error being returned.

openstack/types.go
jtopjian: I agree with this, but slight change of formatting:

```go
log.Printf("[DEBUG] OpenStack connection error, retry number %d: ", retry, err)
```

The error message should always be the last argument so it doesn't get muddled with the rest of the log template.

PR #415, opened 2018-09-09
ContainerInfraV1: add Cluster template datasource

ozerovandrei:
Add new datasource for the OpenStack Magnum cluster template.
Provide acceptance test and website documentation.

For #402


ozerovandrei:
@jtopjian thanks for catching that.

PR #416, opened 2018-09-10
Add description field to openstack_networking_secgroup_rule_v2

jsm222:
For #419

### DIFF
--- a/openstack/resource_openstack_networking_secgroup_rule_v2.go
+++ b/openstack/resource_openstack_networking_secgroup_rule_v2.go
@@ -34,6 +34,12 @@ func resourceNetworkingSecGroupRuleV2() *schema.Resource {
 				Computed: true,
 				ForceNew: true,
 			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: false,
+				ForceNew: true,
+			},
 			"direction": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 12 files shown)


jsm222:
gophercloud/gophercloud#1230

ozerovandrei:
@jsm222 Hello.

I will add a [WIP] to a title since those changes can't be added as-is because they're depend on upstream library implementation.

We need to add [#1231](https://github.com/gophercloud/gophercloud/pull/1231) into Gophercloud first.
And then the latest Gophercloud code must be added into `vendor` via the `govendor` tool. It can't be just edited inside vendor directory.

Also it will be better if you create an issue for this PR.

You can read more about development process there: [testing-and-development](https://www.terraform.io/docs/providers/openstack/index.html#testing-and-development)

jsm222:
[#1231](gophercloud/gophercloud#1231) is merged. How do I update the gophercloud vendor correctly with govendor? I'm quite new to  golang.. Thanks

jsm222:
Gave it a shot and missed. 

> govendor sync; 
> make test

works locally..

jtopjian:
@jsm222 No problem at all - we need to update the developer documentation on how to use `govendor`.

Try this command:

```
govendor fetch github.com/gophercloud/gophercloud/...@a5e4374c99156e0f98429d363ace306d8f9fff6d
```

What this will do is bring all of Gophercloud up-to-date with the latest commit. You'll see that there are other files in `vendor` which are updated and that's OK. It's usually best to have the vendor update in a separate commit from the actual Terraform changes.

Let me know if this helps or if you run into any problems.

jtopjian:
@jsm222 Perfect :smile: 

The last part is to update the documentation with the new field: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/docs/r/networking_secgroup_rule_v2.html.markdown

ozerovandrei:
@jsm222 thank you for adding this feature!

PR #421, opened 2018-09-15
ContainerInfra V1: add Cluster resource

ozerovandrei:
Add the new "openstack_containerinfra_cluster_v1" resource with basic acceptance test.

For #402 

References: 
[Cluster statuses](https://github.com/openstack/magnum/blob/stable/rocky/magnum/objects/fields.py#L18)
[Cluster creation timeout](https://github.com/openstack/magnum/blob/stable/rocky/magnum/conf/cluster_heat.py#L34)
[Only nodes_count can be updated](https://developer.openstack.org/api-ref/container-infrastructure-management/#update-information-of-cluster)


ozerovandrei:
recheck

liu-sheng:
recheck

openstack/resource_openstack_containerinfra_cluster_v1_test.go
```diff
@@ -0,0 +1,200 @@
...
+  properties {
+    os_distro = "fedora-atomic"
+  }
+}
+
+resource "openstack_compute_keypair_v2" "keypair_1" {
+  name = "%s"
+}
+
+resource "openstack_compute_flavor_v2" "flavor_1" {
```
jtopjian: Let's have the flavor be defined with `OS_MAGNUM_FLAVOR` and `OS_MAGNUM_MASTER_FLAVOR`. This way, non-admin users can run the tests.

mnaser:
@jtopjian I think this might be related to our new region, let me verify.

mnaser:
@jtopjian please give it another shot, it *should* work now

jtopjian:
@mnaser sorry for the late reply - busy day! Still seeing the same error.

ozerovandrei:
@jtopjian Thank you for a thorough review!

mnaser:
Better late than ever, but Magnum should be working now @jtopjian

jtopjian:
@mnaser Nice! Thank you for following up!

Issue #420, opened 2018-09-14
Compute V2: cloud-side resize verifying


ozerovandrei:
Greetings.

Right now `terraform-provider-openstack` always waits for 10 seconds to get the `VERIFY_RESIZE` status of a Compute instance and then sends verify request: [resource_openstack_compute_instance_v2.go#L810](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_instance_v2.go#L810)

This behavior doesn't work well in cases where OpenStack cloud provider automatically sets the `ACTIVE` state for the Compute instance.
Cloud provider can do that by setting the [resize_confirm_window](https://docs.openstack.org/nova/rocky/configuration/config.html#DEFAULT.resize_confirm_window) value to some integer. In that case Compute service will automatically confirm resize of the instance after configured interval.
Therewith `terraform-provider-openstack` will throw an error and won't complete the resize operation: [resource_openstack_compute_instance_v2.go#L812](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_instance_v2.go#L812).

I think of two ways of how to deal with it:
- allow to configure `resource.StateChangeConf.Delay` by some env variables instead of using hardcoded values;
- don't interrupt resize action on verify errors, just warn in log (or even quietly continue if instance is already in `ACTIVE` state).

@jtopjian thoughts?

References:
https://docs.openstack.org/nova/rocky/user/resize.html
https://developer.openstack.org/api-ref/compute/#resize-server-resize-action

jtopjian:
Good question... I'm not very familiar with resizing.

Option 1 might cause issues depending on what the cloud provider has set the resize window to. No matter what value the user sets, in some situations Nova might always confirm first.

Instead of an environment variable to set the delay, what about an argument+env var that skips/ignores resize errors? Maybe we could model this as a [`vendor_option`](https://github.com/terraform-providers/terraform-provider-openstack/blob/e9fe36a257561b7b0a7e29ff409e504c8305facf/openstack/resource_openstack_networking_router_v2.go#L111-L126) of `ignore_resize_confirmation` or something like that?

ozerovandrei:
@jtopjian sounds great.

I will create a PR with `ignore_resize_confirmation` option then.

ozerovandrei:
Implemented in #422

PR #422, opened 2018-09-17
ComputeV2: add instance vendor_options

ozerovandrei:
Add "vendor_options" map attribute for instance with the
"ignore_resize_confirmation" option.

New option can be used to ignore manual confirmation of the Compute
instance resizing. Some OpenStack cloud providers automatically confirm
resizing of instances after configured timeout.

For #420

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -360,6 +360,21 @@ func resourceComputeInstanceV2() *schema.Resource {
 					return
 				},
 			},
+			"vendor_options": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				MinItems: 1,
+				MaxItems: 1,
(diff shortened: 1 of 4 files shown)


ozerovandrei:
@jtopjian I added it.

`I don't have an easy way of checking this.`
Well, I checked it in on the [Selectel VPC](https://selectel.ru/en/services/cloud/vpc/), it's OpenStack-based cloud and it has the [resize_confirm_window](
https://docs.openstack.org/nova/rocky/configuration/config.html#DEFAULT.resize_confirm_window) configured to 10 seconds. So it's always causes errors for the `terraform-provider-openstack` users 😄

jtopjian:
Awesome - sounds good.

jtopjian:
Oh duh - merged this by habit. Sorry about that.

ozerovandrei:
@jtopjian whatever, it was ready anyway 😄

PR #426, opened 2018-09-20
Handle rackspace server addresses without type

kouk:
This change fixes the server network's fixed IP addresses appearing blank in the resource output because the [rackspace servers api](https://developer.rackspace.com/docs/cloud-servers/v2/api-reference/svr-basic-operations/#show-server-details) doesn't return the `OS-EXT-IPS:type` field in the `addresses` list items, e.g.:

    {
        "server": {
            "OS-DCF:diskConfig": "AUTO",
            "OS-EXT-STS:power_state": 1,
            "OS-EXT-STS:task_state": null,
            "OS-EXT-STS:vm_state": "active",
            "accessIPv4": "198.101.241.238",
            "accessIPv6": "2001:4800:780e:0510:d87b:9cbc:ff04:513a",
            "addresses": {
                "private": [
                    {
                        "addr": "10.180.3.171",
                        "version": 4
                    }
                ],
                "public": [
                    {
                        "addr": "198.101.241.238",
                        "version": 4
                    },
                    {
                        "addr": "2001:4800:780e:0510:d87b:9cbc:ff04:513a",
                        "version": 6
                    }
                ],
                "production": [
                    {
                        "addr": "192.168.21.30",
                        "version": 4
                    }
                ]
            },
           ...

Output from `terraform refresh` before the change:

        {
            access_network = 1,
            fixed_ip_v4 = ,
            fixed_ip_v6 = ,
            floating_ip = ,
            mac = ,
            name = production,
            port = ,
            uuid = deadbeef-67bb-4ca1-9ec4-deadbeef
        }

output after the change:

        {
            access_network = 1,
            fixed_ip_v4 = 192.168.21.30,
            fixed_ip_v6 = ,
            floating_ip = ,
            mac = ,
            name = production,
            port = ,
            uuid = deadbeef-67bb-4ca1-9ec4-deadbeef
        }

Co-Authored-By: Konstantinos Koukopoulos <kouk@transifex.com>
Co-Authored-By: Nestoras Stefanou <nestoras@transifex.com>

### DIFF
--- a/openstack/compute_instance_v2_networking.go
+++ b/openstack/compute_instance_v2_networking.go
@@ -324,7 +324,7 @@ func getInstanceAddresses(addresses map[string]interface{}) []InstanceAddresses 
 				instanceNIC.MAC = v
 			}
 
-			if v["OS-EXT-IPS:type"] == "fixed" {
+			if v["OS-EXT-IPS:type"] == "fixed" || v["OS-EXT-IPS:type"] == nil {
 				switch v["version"].(float64) {
 				case 6:
 					instanceNIC.FixedIPv6 = fmt.Sprintf("[%s]", v["addr"].(string))


jtopjian:
@kouk Thanks for submitting this. 

It looks like the JSON sample you gave is from the Rackspace documentation. Would you be able to provide actual output by doing:

```shell
$ TF_LOG=DEBUG OS_DEBUG=1 terraform apply 2>&1 | tee ~/log.txt
```

Then sanitize the output and copy it to a Github gist?

The reason is because if this patch is working for you, then the `OS-EXT-IPS:type` key is set or else Terraform would crash. I just want to confirm if this is true.

kouk:
@jtopjian hi, here's the gist: https://gist.github.com/kouk/41aac55a8b854359146e543805e2b5c8

I included the debug output without the patch as well and the terraform config that produced the logs.

As you can see the OS-EXT-IPS:type key is not present at all, however this doesn't lead to a crash (I'm no golang expert but i've read the value of a missing key is `nil`)

jtopjian:
> As you can see the OS-EXT-IPS:type key is not present at all, however this doesn't lead to a crash (I'm no golang expert but i've read the value of a missing key is nil)

Yes, this is correct :)

Thank you for adding the API output. 

Thank you for you work on this. This looks good to me and I'm OK with merging it. However, I'm always cautious about making changes to the server/instance network configuration since changes in behavior can affect a large number of users. If this ends up causing issues, I might have to revert this since this provider does not officially support Rackspace.

jtopjian:
I'm going to omit this feature from the CHANGELOG because I don't want it to be interpreted that we're now supporting Rackspace. I do appreciate the work you and @nestoras did here. If omitting this from the CHANGELOG is an issue, please let me know.

jtopjian:
Actually, nevermind - a CHANGELOG entry should be made in case this causes problems for other users.

kouk:
Thanks!

Just a thought that it could be that this affects some other versions of
openstack, aside from rackspace, ( before Mitaka):
https://specs.openstack.org/openstack/nova-specs/specs/mitaka/approved/ips-add-mac-and-type.html

>

jtopjian:
Good call. 

By default, all communication with this provider should be done over the v2 API. If the user is able to access the v2.1 API, then, yes, there looks like there is a small window in microversions where IP addresses would not have been parsed and this patch will help with that.

PR #427, opened 2018-09-20
[WIP] Allow delete for ERROR-state instances

### LINKED ISSUE #404: Allow to delete Error-state instances
Hi there, I am running OpenStack Ocata version and deploying infrastructure via Terraform.
When I create instances and they got into the error state for some reasons - I cannot destroy them with terraform. 
I guess there are some issues in handling instance's power state. At Error message, Terraform complains about 
```Invalid current power_state:ERROR ```
but actual power_state of instance in reproduce steps - is RUNNING.

 

### Terraform Version
Terraform v0.11.7
+ provider.openstack v1.8.0

### Affected Resource(s)
- instance resource

### Expected Behavior
delete all error state instances

### Steps to Reproduce

1. create an instance via terraform any way you like, apply it

2. set instance state into "ERROR" using 
```openstack server set --state error  ee21f319-eab1-403a-ae33-5c88bd0a787a```

3.  terraform destroy will fail with an error

```Error: Error refreshing state: 1 error(s) occurred:

* openstack_compute_instance_v2.instance: 1 error(s) occurred:

* openstack_compute_instance_v2.instance[0]: openstack_compute_instance_v2.instance.0: Error changing the instance(ee21f319-eab1-403a-ae33-5c88bd0a787a) power_state. Invalid current power_state:ERROR 




Error: Error refreshing state: 1 error(s) occurred:

* openstack_compute_instance_v2.instance: 1 error(s) occurred:

* openstack_compute_instance_v2.instance[0]: openstack_compute_instance_v2.instance.0: Error changing the instance(ee21f319-eab1-403a-ae33-5c88bd0a787a) power_state. Invalid current power_state:ERROR ```

ChiefAlexander:
Fixes #404

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -45,7 +45,6 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Computed: true,
 				ForceNew: true,
 			},
-
 			"name": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
@@ -342,6 +341,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"destroy_on_error": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true, // i'm lazy and set this to true while testing so i wouldn't have to update my terraform code
+			},
 			"all_metadata": &schema.Schema{
 				Type:     schema.TypeMap,
 				Computed: true,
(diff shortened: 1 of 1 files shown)


jtopjian:
@ChiefAlexander It looks like this change was trickier than I imagined :)

Your code was working fine -- if you add some `log.Printf` entries, you'll see that the areas of code are being triggered.

The problem is that Terraform will do a "read" before it looks at any schema changes. So in the situation where a user created an instance without the new `destroy_on_error` flag, they would be unable to modify their configuration to allow an error state.

In my opinion, digging out of this situation is a bug, so allowing a state of "error" without a new flag is the right way to go.

The next problem was that we wanted to allow an error state but we don't want a user to be able to define `power_state` of "error" because there's nothing in the code that would do the transition. So we need a way of allowing an "error" state but still only allowing "active" and "shutoff" as the only two inputs. We need to do a `DiffSuppressFunc` to allow this.

I've submitted #428. Let me know what you think.

ChiefAlexander:
Awesome thank you for taking a look! With your changes in #428 do we need this pr still?

jtopjian:
Nope, this is closable. :)

PR #428, opened 2018-09-21
Compute v2: Allow power_state of error

jtopjian:
This commit will allow an instance to cleanly transition into a
status of "error". This will allow users to destroy instances
which have changed into an error state either during creation
or by some event outside of Terraform.

For #404 

Additional comments in #427

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -23,6 +23,7 @@ import (
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceComputeInstanceV2() *schema.Resource {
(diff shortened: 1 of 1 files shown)


openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -628,13 +625,20 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
...
-		d.Set("power_state", strings.ToLower(server.Status))
+	// Set the current power_state
+	var validStatus bool
+	currentStatus := strings.ToLower(server.Status)
+	switch currentStatus {
+	case "active", "shutoff", "error":
+		validStatus = true
 	}
 
+	if !validStatus {
```
ChiefAlexander: Why wouldn't you just have this as the default action of your switch if we don't meet any of the cases?

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -628,13 +625,20 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
...
+		return fmt.Errorf("Invalid power_state for instance %s: %s", d.Id(), server.Status)
+	}
+
+	d.Set("power_state", currentStatus)
```
ChiefAlexander: Move this into your switch statement instead of having a var validStatus so that we can join both the if statement for !validStatus and the switch

openstack/resource_openstack_compute_instance_v2.go
jtopjian: The switch case is only being used as a lazy if/else statement. Generally I try to avoid multi-clauses when I can since it prevents code from being read from top-to-bottom without interruption.

But totally valid point and I'll try this out.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -628,11 +625,13 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
...
-		return fmt.Errorf("Error changing the instance(%s) power_state. Invalid current power_state:%s ", d.Id(), server.Status)
-	} else {
-		d.Set("power_state", strings.ToLower(server.Status))
+	// Set the current power_state
+	currentStatus := strings.ToLower(server.Status)
+	switch currentStatus {
+	case "active", "shutoff", "error":
+		d.Set("power_state", currentStatus)
+	default:
+		return fmt.Errorf("Invalid power_state for instance %s: %s", d.Id(), server.Status)
```
jtopjian: @ChiefAlexander Good call. That turned out much cleaner.

ChiefAlexander: Looks great!

jtopjian:
Thanks, all

Issue #429, opened 2018-09-21
Define 'ip_address = "" ' parameter in "fixed_ip"  for openstack_networking_port_v2 ,but the ip will be assigned a value.


xudehua:
Hi there,

Terraform v0.11.7
+ provider.openstack v1.9.0

### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_compute_instance_v2
- openstack_networking_port_v2

According the terraform docs https://www.terraform.io/docs/providers/openstack/r/networking_port_v2.html

The fixed_ip block supports:
    subnet_id - (Required) Subnet in which to allocate IP address for this port.
    ip_address - (Optional) IP address desired in the subnet for this port. If you don't specify ip_address, an available IP address from the specified subnet will be allocated to this port. **This field will not be populated if it is left blank.** To retrieve the assigned IP address, use the all_fixed_ips attribute.

    I create port and instance
resource "openstack_networking_port_v2" "port_net-03_v4" {
    name = "test"
    network_id = "${var.uuid}"
    admin_state_up = "true"
    security_group_ids = [ "${var.security_group_id}" ]
    fixed_ip {
        "subnet_id" = "${var.sub_uuid}"
        "ip_address" = ""
    }
}

 Output
   I just want to create a port in the instance ,no IP required. But after creating instance , the instance's port had been assigned IP value( may be by DHCP).

Thanks

ozerovandrei:
@xudehua Hello.

This is expected behaviour according to [OpenStack Neutron Documentation](https://developer.openstack.org/api-ref/network/v2/#create-port):

> If you specify only a subnet ID, OpenStack Networking allocates an available IP from that subnet to the port.

And we have the following note in the [Terraform Provider Documentation](https://www.terraform.io/docs/providers/openstack/r/networking_port_v2.html):

> If you don't specify ip_address, an available IP address from the specified subnet will be allocated to this port.

Setting this value to blank is the same as skipping this flag. 
Let me know if you have any questions.

xudehua:
@ozerovandrei  Hi,

Thank you for you answer.
By terraform *.tf file,I created a port from the specified subnet and attached to one VM on openstack platformthe .I hope the port's IP address is blank when the VM was deployed.
Maybe I misunderstood the note “**This field will not be populated if it is left blank.**”. I thought that  the port's IP address would not be assigned when the "ip_address" was set blank in the fixed_ip.

Thanks again

Edward

[...]

PR #433, opened 2018-09-26
Networking v2: Add no_fixed_ip to openstack_networking_port_v2

jtopjian:
This commit adds the `no_fixed_ip` argument to the
openstack_networking_port_v2 resource which will enable a port to
be created without a Fixed IP address set.

For #429

### DIFF
--- a/openstack/resource_openstack_networking_port_v2.go
+++ b/openstack/resource_openstack_networking_port_v2.go
@@ -167,20 +167,6 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
 		return fmt.Errorf("Cannot have both no_security_groups and security_group_ids set")
 	}
 
-	// Parse the fixed_ips
-	fixedIPs := resourcePortFixedIpsV2(d)
-
-	// Check if the user specified no_fixed_ip.
-	//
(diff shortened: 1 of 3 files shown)


openstack/resource_openstack_networking_port_v2.go
```diff
@@ -160,6 +167,20 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
...
+	fixedIPs := resourcePortFixedIpsV2(d)
+
+	// Check if the user specified no_fixed_ip.
+	//
+	// Since fixed_ip and no_fixed_ip are mutually exclusive,
+	// we can safely overwrite the previous fixedIPs variable.
+	//
+	// Since we're only concerned about a value of "true",
+	// GetOk is used.
+	if _, ok := d.GetOk("no_fixed_ip"); ok {
```
ozerovandrei: Maybe it's better to hide this logic in the `resourcePortFixedIpsV2`? Because it doesn't make sense to run `resourcePortFixedIpsV2` if `no_fixed_ip` was set.

jtopjian: Sounds good to me. See the latest commit.

Issue #434, opened 2018-09-28
provider configuration is ignored


xmik:
### Terraform Version
* terraform v0.11.8
* terraform-provider-openstack v1.9.0

### Affected Resource(s)
Provider configuration.

### Terraform Configuration Files
```hcl
provider "openstack" {
  auth_url    = "http://<openstack_host>:<openstack_port>/v3"
  domain_name = "Default"
  tenant_name = "my_tenant"
}
resource "openstack_networking_port_v2" "port" {
  count       = "1"
  name        = "terraform-openstack-provider-bug"
  network_id     = "some-long-id"
  no_security_groups = "true"
}
```

### Actual Behavior
The tenant_name is taken from `OS_PROJECT_NAME` variable.

### Expected Behavior
I expected the tenant_name to be taken from hcl files (here: "my_tenant").

### Steps to Reproduce
Having set environment variables to:
```
$ env | grep OS_
OS_REGION_NAME=RegionOne
OS_USER_DOMAIN_NAME=Default
OS_PROJECT_DOMAIN_NAME=Default
OS_IDENTITY_API_VERSION=3
OS_AUTH_URL=http://<openstack_host>:<openstack_port>/v3
OS_USERNAME=<some_username>
OS_PASSWORD=<some_password>
OS_PROJECT_NAME=testing
OS_TENANT_NAME=testing
OS_INTERFACE=internal
```
I run:
```
OS_DEBUG=1 TF_LOG=debug terraform plan
```
the output contains (look at project name):
```
2018-09-28T15:37:50.571Z [DEBUG] plugin.terraform-provider-openstack_v1.9.0_x4: 2018/09/28 15:37:50 [DEBUG] OpenStack Request Body: {
2018-09-28T15:37:50.571Z [DEBUG] plugin.terraform-provider-openstack_v1.9.0_x4:   "auth": {
2018-09-28T15:37:50.571Z [DEBUG] plugin.terraform-provider-openstack_v1.9.0_x4:     "identity": {
2018-09-28T15:37:50.571Z [DEBUG] plugin.terraform-provider-openstack_v1.9.0_x4:       "methods": [
2018-09-28T15:37:50.571Z [DEBUG] plugin.terraform-provider-openstack_v1.9.0_x4:         "password"
2018-09-28T15:37:50.571Z [DEBUG] plugin.terraform-provider-openstack_v1.9.0_x4:       ],
[...]

PR #436, opened 2018-10-04
vendor: Updating gophercloud/utils

jtopjian:
This commit updates the gophercloud/utils vendor packages to resolve a
bug where the gophercloud/utils clientconfig package was overriding
settings with environment variables.

For #434


jtopjian:
recheck

jtopjian:
recheck

PR #441, opened 2018-10-11
Acc Tests: Updating LBaaS v2 Member test

jtopjian:


jtopjian:
I'm looking into another LBaaS thing and had to make these changes for the test to consistently pass.

jtopjian:
The test failure is due to an OpenLab issue which is being worked on.

Issue #432, opened 2018-09-25
Feature request: Update Openstack Provider multiattach functionality


jamesmmccarthy:
The gophercloud sdk has been updated to facilitate multiattach enabled volumes in openstack.

- [x] New `multiattach` argument and attribute for the `openstack_blockstorage_volume_v3` resource #431 
- [x] improve work with attachment of multiattach-enabled volumes

ozerovandrei:
Implemented in #431.

@jamesmmccarthy thank you!

jamesmmccarthy:
@jtopjian @ozerovandrei Joe and myself have been chatting some more about this offline. While the work that has been done so far is fine in and of itself, there looks to be some more work needed around attaching/dealing with multiattach enabled volumes. Is it ok to re-open this issue, or should a new one be created ?

jtopjian:
I think this is closable again - but we can re-open if there's still anything missing :)

PR #442, opened 2018-10-13
Compute v2: Enable Multiattach Volumes

jtopjian:
This commit allows instances to attach multiattach volumes
by adding a new argument called "multiattach". When set, the
compute client will use Microversion 2.60, which enables multiattach
support in the Compute/Nova API.

For #432 

@jamesmmccarthy I'm unable to test multiattach volumes at the moment, so I'm not sure if this works. Would you be able to give it a try? I've updated the documentation to include an example implementation, though I'm not sure if it's correct.

### DIFF
--- a/openstack/resource_openstack_compute_volume_attach_v2.go
+++ b/openstack/resource_openstack_compute_volume_attach_v2.go
@@ -52,6 +52,12 @@ func resourceComputeVolumeAttachV2() *schema.Resource {
 				Computed: true,
 				Optional: true,
 			},
+
+			"multiattach": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+			},
 		},
 	}
 }
@@ -77,6 +83,10 @@ func resourceComputeVolumeAttachV2Create(d *schema.ResourceData, meta interface{
 	}
 
 	log.Printf("[DEBUG] Creating volume attachment: %#v", attachOpts)
+
+	if v := d.Get("multiattach").(bool); v {
+		computeClient.Microversion = "2.60"
+	}
 
 	attachment, err := volumeattach.Create(computeClient, instanceId, attachOpts).Extract()
 	if err != nil {
--- a/website/docs/r/compute_volume_attach_v2.html.markdown
+++ b/website/docs/r/compute_volume_attach_v2.html.markdown
@@ -93,6 +93,45 @@ output "volume devices" {
 }
 ```
 
+### Using Multiattach-enabled volumes
+
+Multiattach Volumes are dependent upon your OpenStack cloud and not all
+clouds support multiattach.
+
+```hcl
+resource "openstack_blockstorage_volume_v3" "volume_1" {
+  name        = "volume_1"
+  size        = 1
+  multiattach = true
+}
+
+resource "openstack_compute_instance_v2" "instance_1" {
+  name            = "instance_1"
+  security_groups = ["default"]
+}
+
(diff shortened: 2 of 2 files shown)


jamesmmccarthy:
Initial attempts at testing this PR seem to show it's not having the desired effect - will look into this further. (Also small typo in the doc section)

jamesmmccarthy:
I have re-tested and found that this PR actually does work - it requires adding multiattach = true property to the attachment definition, looks like the doc typo is the only outstanding issue. Probably with terraform/IaC this is a normal requirement.

jamesmmccarthy:
For completeness I also did a test where a single volume is attached to two instances. It required the use of depends_on for sanity, otherwise it worked well, like so:

+--------------------------------------+-------------+--------+----------------------+--------------+--------+
| ID                                   | Name        | Status | Networks             | Image        | Flavor |
+--------------------------------------+-------------+--------+----------------------+--------------+--------+
| 60c521a6-837c-4e03-a81e-01d11662ca2e | myinstance2 | ACTIVE | public=10.196.223.11 | cirros-0.4.0 |        |
| 7deb75a3-14a1-4ac6-943b-931f6c8f1161 | myinstance  | ACTIVE | public=10.196.223.23 | cirros-0.4.0 |        |
+--------------------------------------+-------------+--------+----------------------+--------------+--------+
[kolla@operator2 boot-from-vol]$ openstack volume list
+--------------------------------------+-------+--------+------+-------------------------------------------------------------------------+
| ID                                   | Name  | Status | Size | Attached to                                                             |
+--------------------------------------+-------+--------+------+-------------------------------------------------------------------------+
| b35f8871-2392-4dc4-af78-0f86ddcebd41 | myvol | in-use |    1 | Attached to myinstance on /dev/vdb Attached to myinstance2 on /dev/vdb  |
+--------------------------------------+-------+--------+------+-------------------------------------------------------------------------+

jtopjian:
> it requires adding multiattach = true property to the attachment definition, looks like the doc typo is the only outstanding issue

Where is the doc typo? The example in this PR has the following:

```hcl
resource "openstack_compute_volume_attach_v2" "va_1" {
  instance_id = "${openstack_compute_instance_v2.instance_1.id}"
  volume_id   = "${openstack_blockstorage_volume_v2.volume_1.id}"
  multiattach = true
}

resource "openstack_compute_volume_attach_v2" "va_2" {
  instance_id = "${openstack_compute_instance_v2.instance_2.id}"
  volume_id   = "${openstack_blockstorage_volume_v2.volume_1.id}"
  multiattach = true
}
```

> It required the use of depends_on for sanity

This makes sense. I'll update the PR to include this.

Thank you for testing this :smile:

jamesmmccarthy:
Line 151 * `multiattach` - (Optional) Enable attachment of multiattach-capabile volumes.

jamesmmccarthy:
re: depends_on: the "va_2" resource would want:
depends_on = ["openstack_compute_volume_attach_v2.va_1"]

Thanks for the updates !

jamesmmccarthy:
re: depends_on: the "va_2" resource would want something like:
depends_on = ["openstack_compute_volume_attach_v2.va_1"]

Thanks for the updates ! :smile:

jtopjian:
> Probably there are other terraform techniques to achieve similar, this was what I used for this testing.

I think for this situation, `depends_on` is the most appropriate solution. `depends_on` is the cleanest way of "serializing" Terraform.

jtopjian:
recheck

jtopjian:
recheck

Issue #337, opened 2018-06-04
openstack_compute_instance_v2 access_ip_v6 returns addresses in literal syntax


Lasering:
### Terraform Version
```
Terraform v0.11.7
+ provider.openstack v1.5.0
```
### Affected Resource(s)
- openstack_compute_instance_v2
- openstack_dns_recordset_v2

### Terraform Configuration Files
```hcl
resource "openstack_compute_instance_v2" "servers" {
  count = "${var.number-of-servers}"

  name        = "server${count.index}-${var.suffix}"
  flavor_name = "${var.flavor-name}"
  image_name  = "${var.image-name}"

  key_pair        = "${local.keypair}"
  security_groups = ["${openstack_networking_secgroup_v2.consul.name}"]
}

resource "openstack_dns_recordset_v2" "server-ipv6" {
  zone_id = "${data.openstack_dns_zone_v2.default.id}"
  name    = "${local.address}."
  ttl     = 1
  type    = "AAAA"
  records = ["${openstack_compute_instance_v2.servers.*.access_ip_v6}"]
}
```
### Actual Behavior
`openstack_dns_recordset_v2.server-ipv6: Error creating OpenStack DNS record set: Invalid request due to incorrect syntax or missing required parameters.`

### Expected Behavior
Either:

  1. The field `access_ip_v6` of `openstack_compute_instance_v2.servers` should return the IPv6 addresses without any brackets, aka, not in literal syntax.
  2. `openstack_dns_recordset_v2` should accept the value returned by `${openstack_compute_instance_v2.servers.*.access_ip_v6}`

### Important Factoids
This hack solves the problem while its not fixed:
```hcl
resource "openstack_dns_recordset_v2" "server-ipv6" {
  zone_id = "${data.openstack_dns_zone_v2.default.id}"
  name    = "${local.address}."
  ttl     = 1
  type    = "AAAA"

  # Records must be an array of IPv6 address without brackets. So this is valid: ["8afe:affb:9d95:d381:40c:7d10:c714:339e", "e163:648f:7797:3548:9827:7799:cbd3:e5fa"]
  # However this is not: ["[8afe:affb:9d95:d381:40c:7d10:c714:339e]", "[e163:648f:7797:3548:9827:7799:cbd3:e5fa]"]
  # Unfortunately openstack_compute_instance_v2.servers.*.access_ip_v6 returns an array of IPv6 addresses in literal syntax
[...]

PR #443, opened 2018-10-13
DNSv2: Format IPv6 Records

jtopjian:
This commit will format IPv6 records in a way that the DNSaaS /
Designate service expects them to be. Notably, it strips the brackets
from the addresses.

Passing IPv6 records with brackets already removed still works.

For #337


jtopjian:
I went ahead and removed the "timeout" acceptance test since it's no longer needed. Timeouts have been a core Terraform resource long enough that we can consider them stable.

Issue #345, opened 2018-06-19
Does terraform-provider-openstack supports trunk ports?


jnunyez:
Hi,
does terraform-provider-openstack supports the feature of adding/setting network trunk ports?

Thanks,
Jose

jtopjian:
Hello,

Unfortunately the Neutron trunk extension is not supported at this time. I'm going to leave this issue open to track it as a feature request.

Please let me know if you have any questions :)

celebdor:
@jtopjian I guess you can close it now

jtopjian:
@celebdor Yes, good point :)

@jnunyez This has been implemented in #446 and will be available in the next release.

Issue #174, opened 2017-11-30
allowed_address_pairs does not accept list


michalmedvecky:
`allowed_address_pairs` does not accept (correct?) interpolated parameter.

### Terraform Version

0.11

### Affected Resource(s)
- openstack_networking_port_v2

### Terraform Configuration Files

First declare 10 ports:

```hcl
resource "openstack_networking_port_v2" "vip" {
  name = "mytest-${count.index}"
  network_id = "${openstack_networking_network_v2.net-backend.id}"
  admin_state_up = "true"
  count = "10"
}
```

Next, create port where those 10 addresses are allowed:

```hcl
resource "openstack_networking_port_v2" "port" {
  name = "mytest"
  admin_state_up = "true"
  network_id = "${openstack_networking_network_v2.net-backend.id}"
  allowed_address_pairs = "${formatlist("{ip_address=\"%v\"}", 
  openstack_networking_port_v2.vip.*.all_fixed_ips.0)}"
}
```

### Actual result

```
Error: openstack_networking_port_v2.port: allowed_address_pairs: should be a list
```

If you output the interpolation:

```hcl
output "mylist" {
  value="${formatlist("{ ip_address=\"%v\" }", openstack_networking_port_v2.vip.*.all_fixed_ips.0)}"
}
```

you get:

```
mylist = [
    { ip_address="100.127.172.186" },
    { ip_address="100.127.172.19" },
    { ip_address="100.127.172.180" },
    { ip_address="100.127.172.18" },
    { ip_address="100.127.172.182" },
    { ip_address="100.127.172.184" },
    { ip_address="100.127.172.189" },
    { ip_address="100.127.172.183" },
    { ip_address="100.127.172.179" },
    { ip_address="100.127.172.181" }
]
```

And if you copypaste this output to the `.tf` file, it works. But that's not what you want to do :-)

anta5010:
Firstly, try to put square brackets:

```
allowed_address_pairs = [ "${formatlist("{ip_address=\"%v\"}", openstack_networking_port_v2.vip.*.all_fixed_ips.0)}" ]

```

If this doesn't work then I think this issue has the same root cause as the one I raised
https://github.com/terraform-providers/terraform-provider-openstack/issues/170

michalmedvecky:
I tried. The difference is `Error: openstack_networking_port_v2.ldap16_port: "allowed_address_pairs.0.ip_address": required field is not set`

[...]

Issue #170, opened 2017-11-26
I can't define "fixed_ip" parameter for openstack_networking_port_v2 using a list based on splat syntax list of subnets IDs


anta5010:
### Terraform Version
$ terraform -v
Terraform v0.11.0
+ provider.null v1.0.0
+ provider.openstack v1.0.0
+ provider.template v1.0.0

### Affected Resource(s)
openstack_networking_port_v2 but it might be a common issue for other openstack provider resources

### Terraform Configuration Files

I would like to define fixed IPs dynamically for openstack_networking_port_v2 definitions using splat syntax list of subnets IDs. This way I can deploy IPv4-only  or IPv4+IPv6 networks using the same terraform definitions.

```hcl

variable "is_ipv6" {}

resource "openstack_networking_subnet_v2" "s1_public" {
  count      = "${var.is_ipv6+1}"
  name       = "${var.cdn_name}s1.public_v${count.index == 0 ? 4 : 6}-${var.user_name}"
  network_id = "${openstack_networking_network_v2.s1_public.id}"
  cidr       = "${element(var.cdn_ipnets["s1_public"], count.index)}"
  ip_version = "${ count.index == 0 ? 4 : 6 }"
}

data "null_data_source" "s1_public_gw_ips" {
  count = "${var.is_ipv6+1}"
  depends_on = [ "openstack_networking_subnet_v2.s1_public" ]
  inputs = {
    subnet_id = "${element(openstack_networking_subnet_v2.s1_public.*.id,count.index)}"
    ip_address = "${cidrhost(element(var.cdn_ipnets["s1_public"], count.index), var.public_offsets["gw1"])}"
  }
}

resource "openstack_networking_port_v2" "rtr_s1_public_port" {
  name = "gw1.${var.cdn_name}s1.public-${var.user_name}"
  network_id = "${openstack_networking_network_v2.s1_public.id}"
  admin_state_up = "true"
  depends_on = [ "data.null_data_source.s1_public_gw_ips" ]

  fixed_ip = [ "${data.null_data_source.s1_public_gw_ips.*.inputs}" ]
}
```

### Expected Behavior

  A port should be created with either one (ipv4) or two (ipv4 and ipv6) IP addresses depending on is_ipv6 variable.

### Actual Behavior
```
Error: openstack_networking_port_v2.rtr_s1_public_port: "fixed_ip.0.subnet_id": required field is not set
```

### Steps to Reproduce

[...]

PR #446, opened 2018-10-16
Network resource - Trunk

celebdor:
This patch support for the OpenStack networking trunk extension.

For #345 

Signed-off-by: Antoni Segura Puimedon <celebdor@gmail.com>

### DIFF
--- a/openstack/resource_openstack_networking_trunk_v2.go
+++ b/openstack/resource_openstack_networking_trunk_v2.go
@@ -1,0 +1,334 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
+)
+
+func resourceNetworkingTrunkV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceNetworkingTrunkV2Create,
+		Read:   resourceNetworkingTrunkV2Read,
+		Update: resourceNetworkingTrunkV2Update,
+		Delete: resourceNetworkingTrunkV2Delete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
+			"port_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"admin_state_up": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"tenant_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"sub_port": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: false,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"port_id": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"segmentation_type": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"segmentation_id": &schema.Schema{
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+					},
+				},
+			},
+			"tags": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: false,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func resourceNetworkingTrunkV2Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	client, err := config.networkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	createOpts := trunks.CreateOpts{
+		Name:         d.Get("name").(string),
+		AdminStateUp: resourceNetworkingTrunkV2AdminStateUp(d),
+		PortID:       d.Get("port_id").(string),
+		TenantID:     d.Get("tenant_id").(string),
+		Subports:     resourceNetworkingTrunkV2Subports(d.Get("sub_port").(*schema.Set)),
+	}
+
+	log.Printf("[DEBUG] Create Options: %#v", createOpts)
+	trunk, err := trunks.Create(client, createOpts).Extract()
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack Neutron trunk: %s", err)
+	}
+
+	log.Printf("[DEBUG] Waiting for OpenStack Neutron trunk (%s) to become available.", trunk.ID)
+
+	stateConf := &resource.StateChangeConf{
+		Target:     []string{"ACTIVE"},
+		Refresh:    waitForNetworkTrunkActive(client, trunk.ID),
+		Timeout:    d.Timeout(schema.TimeoutCreate),
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error waiting for trunk to become active: %s", err)
+	}
+
+	tags := networkV2AttributesTags(d)
+	if len(tags) > 0 {
+		tagOpts := attributestags.ReplaceAllOpts{Tags: tags}
+		tags, err := attributestags.ReplaceAll(client, "trunks", trunk.ID, tagOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Unable to set trunk tags: %v", err)
+		}
+		log.Printf("[DEBUG] Set Tags = %+v on trunk %s", tags, trunk.ID)
+	}
+
+	d.SetId(trunk.ID)
+	return resourceNetworkingTrunkV2Read(d, meta)
+}
+
+func resourceNetworkingTrunkV2Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	client, err := config.networkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	trunk, err := trunks.Get(client, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "trunk")
+	}
+	log.Printf("[DEBUG] Retrieved trunk %s: %+v", d.Id(), trunk)
+
+	d.Set("region", GetRegion(d, config))
+	d.Set("name", trunk.Name)
+	d.Set("port_id", trunk.PortID)
+	d.Set("admin_state_up", trunk.AdminStateUp)
+	d.Set("tenant_id", trunk.TenantID)
+	d.Set("tags", trunk.Tags)
+
+	subports := make([]map[string]interface{}, len(trunk.Subports))
+	for i, trunkSubport := range trunk.Subports {
+		subports[i] = make(map[string]interface{})
+		subports[i]["port_id"] = trunkSubport.PortID
+		subports[i]["segmentation_type"] = trunkSubport.SegmentationType
+		subports[i]["segmentation_id"] = trunkSubport.SegmentationID
+	}
+	if err = d.Set("sub_port", subports); err != nil {
+		return fmt.Errorf("Unable to set sub_port for trunk %s: %s", d.Id(), err)
+	}
+
+	return nil
+}
+
+func resourceNetworkingTrunkV2Update(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	client, err := config.networkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	var updateTrunk bool
+	var updateOpts trunks.UpdateOpts
+
+	if d.HasChange("name") {
+		updateTrunk = true
+		updateOpts.Name = d.Get("name").(string)
+	}
+	if d.HasChange("admin_state_up") {
+		updateTrunk = true
+		updateOpts.AdminStateUp = resourceNetworkingTrunkV2AdminStateUp(d)
+	}
+
+	if updateTrunk {
+		log.Printf("[DEBUG] Updating trunk %s with options: %+v", d.Id(), updateOpts)
+		_, err = trunks.Update(client, d.Id(), updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating OpenStack Neutron trunk: %s", err)
+		}
+	}
+
+	if d.HasChange("sub_port") {
+		old, new := d.GetChange("sub_port")
+
+		toRemove := old.(*schema.Set).Difference(new.(*schema.Set))
+		toAdd := new.(*schema.Set).Difference(old.(*schema.Set))
+
+		if len(toRemove.List()) > 0 {
+			var subportsToRemove trunks.RemoveSubportsOpts
+
+			subports := resourceNetworkingTrunkV2Subports(toRemove)
+			log.Printf("[DEBUG] Removing subports from trunk %s: %#v", d.Id(), subports)
+			for _, subport := range subports {
+				v := trunks.RemoveSubport{PortID: subport.PortID}
+				subportsToRemove.Subports = append(subportsToRemove.Subports, v)
+			}
+			_, err := trunks.RemoveSubports(client, d.Id(), subportsToRemove).Extract()
+			if err != nil {
+				return fmt.Errorf("Error removing subports when updating OpenStack Neutron trunk: %s", err)
+			}
+		}
+
+		if len(toAdd.List()) > 0 {
+			var subportsToAdd trunks.AddSubportsOpts
+			subportsToAdd.Subports = resourceNetworkingTrunkV2Subports(toAdd)
+
+			log.Printf("[DEBUG] Adding subports to trunk %s: %#v", d.Id(), subportsToAdd.Subports)
+
+			_, err := trunks.AddSubports(client, d.Id(), subportsToAdd).Extract()
+			if err != nil {
+				return fmt.Errorf("Error adding subports when updating OpenStack Neutron trunk: %s", err)
+			}
+		}
+	}
+
+	if d.HasChange("tags") {
+		tags := networkV2AttributesTags(d)
+		tagOpts := attributestags.ReplaceAllOpts{Tags: tags}
+		tags, err := attributestags.ReplaceAll(client, "trunks", d.Id(), tagOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating tags on trunk: %s", err)
+		}
+		log.Printf("[DEBUG] Updated tags = %+v on trunk %s", tags, d.Id())
+	}
+
+	return resourceNetworkingTrunkV2Read(d, meta)
+}
+
+func resourceNetworkingTrunkV2Delete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	client, err := config.networkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"ACTIVE"},
+		Target:     []string{"DELETED"},
+		Refresh:    waitForNetworkTrunkDelete(client, d.Id()),
+		Timeout:    d.Timeout(schema.TimeoutDelete),
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error deleting OpenStack Neutron trunk: %s", err)
+	}
+
+	return nil
+}
+
+func resourceNetworkingTrunkV2Subports(d *schema.Set) (subports []trunks.Subport) {
+	rawSubports := d.List()
+	subports = make([]trunks.Subport, len(rawSubports))
+
+	for i, raw := range rawSubports {
+		rawMap := raw.(map[string]interface{})
+		subports[i] = trunks.Subport{
+			PortID:           rawMap["port_id"].(string),
+			SegmentationType: rawMap["segmentation_type"].(string),
+			SegmentationID:   rawMap["segmentation_id"].(int),
+		}
+	}
+	return
+}
+
+func resourceNetworkingTrunkV2AdminStateUp(d *schema.ResourceData) *bool {
+	value := false
+
+	if v, ok := d.GetOkExists("admin_state_up"); ok {
+		value = v.(bool)
+	}
+
+	return &value
+}
+
+func waitForNetworkTrunkActive(client *gophercloud.ServiceClient, trunkID string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		trunk, err := trunks.Get(client, trunkID).Extract()
+		if err != nil {
+			return nil, "", err
+		}
+
+		log.Printf("[DEBUG] OpenStack Neutron trunk: %+v", trunk)
+		if trunk.Status == "DOWN" || trunk.Status == "ACTIVE" {
+			return trunk, "ACTIVE", nil
+		}
+
+		return trunk, trunk.Status, nil
+	}
+}
+
+func waitForNetworkTrunkDelete(client *gophercloud.ServiceClient, trunkID string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		log.Printf("[DEBUG] Attempting to delete OpenStack Neutron trunk %s", trunkID)
+
+		trunk, err := trunks.Get(client, trunkID).Extract()
+		if err != nil {
+			if _, ok := err.(gophercloud.ErrDefault404); ok {
+				log.Printf("[DEBUG] Successfully deleted OpenStack trunk %s", trunkID)
+				return trunk, "DELETED", nil
+			}
+			return trunk, "ACTIVE", err
+		}
+
+		err = trunks.Delete(client, trunkID).ExtractErr()
+		if err != nil {
+			if _, ok := err.(gophercloud.ErrDefault404); ok {
+				log.Printf("[DEBUG] Successfully deleted OpenStack trunk %s", trunkID)
+				return trunk, "DELETED", nil
+			}
+			return trunk, "ACTIVE", err
+		}
+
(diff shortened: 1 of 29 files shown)


openstack/resource_openstack_compute_instance_v2_test.go
```diff
@@ -724,6 +725,36 @@ func testAccCheckComputeV2InstanceState(
 	}
 }
 
+func TestAccComputeV2Instance_trunk(t *testing.T) {
```
jtopjian: Since the "trunk" resource is the core focus of this test, let's move this to a dedicated `resource_openstack_networking_trunk_v2_test.go` file.

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
+			"port_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"admin_state_up": &schema.Schema{
```
jtopjian: Let's change this to look like: https://github.com/terraform-providers/terraform-provider-openstack/blob/e8fa32f8549b144625f0a4b011be5f83eb0aae2f/openstack/data_source_openstack_networking_router_v2.go#L29-L32

And then access it with: https://github.com/terraform-providers/terraform-provider-openstack/blob/e8fa32f8549b144625f0a4b011be5f83eb0aae2f/openstack/data_source_openstack_networking_router_v2.go#L93-L96

There are some `admin_state_up` arguments in other resources that are `TypeString`. Those were some of the first resources before Terraform was able to handle booleans better.

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+						},
+						"segmentation_id": &schema.Schema{
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+					},
+				},
+			},
+			"tags": &schema.Schema{
+				Type:     schema.TypeSet,
```
jtopjian: This might work better as a `TypeMap`?

https://github.com/terraform-providers/terraform-provider-openstack/blob/e9fe36a257561b7b0a7e29ff409e504c8305facf/openstack/resource_openstack_compute_flavor_v2.go#L69-L73

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+
+	stateConf := &resource.StateChangeConf{
+		Target:     []string{"ACTIVE"},
+		Refresh:    waitForNetworkTrunkActive(client, trunk.ID),
+		Timeout:    d.Timeout(schema.TimeoutCreate),
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
```
jtopjian: Let's make sure we check for an error here:

```go
if err != nil {
  return fmt.Errorf("Error waiting for trunk to become active: %s", err)
}
```

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+	d.Set("tags", trunk.Tags)
+
+	subports := make([]map[string]interface{}, len(trunk.Subports))
+	for i, trunkSubport := range trunk.Subports {
+		subports[i] = make(map[string]interface{})
+		subports[i]["port_id"] = trunkSubport.PortID
+		subports[i]["segmentation_type"] = trunkSubport.SegmentationType
+		subports[i]["segmentation_id"] = trunkSubport.SegmentationID
+	}
+	d.Set("sub_ports", subports)
```
jtopjian: For "non-scalar" types, it's best to check for an error:

```go
if err := d.Set("sub_ports", subports); err != nil {
  return fmt.Errorf("Unable to set sub_port for trunk %s: %s", d.Id(), err)
}
```

website/docs/r/networking_trunk_v2.html.markdown
```diff
@@ -0,0 +1,117 @@
...
+    (must be "true" or "false" if provided). Changing this updates the
+    `admin_state_up` of an existing trunk.
+
+* `tenant_id` - (Optional) The owner of the Trunk. Required if admin wants
+    to create a trunk on behalf of another tenant. Changing this creates a new trunk.
+
+* `sub_ports` - (Optional) The set of ports that will be made subports of the trunk.
+    The structure of each subport is described below.
+
+* `value_specs` - (Optional) Map of additional options.
```
jtopjian: `value_specs` isn't defined here, so let's remove this entry

website/docs/r/networking_trunk_v2.html.markdown
```diff
@@ -0,0 +1,117 @@
...
+
+The `subports` block supports:
```
jtopjian: `sub_port`

website/docs/r/networking_trunk_v2.html.markdown
```diff
@@ -0,0 +1,117 @@
...
+## Attributes Reference
+
+The following attributes are exported:
+
+* `region` - See Argument Reference above.
+* `name` - See Argument Reference above.
+* `port_id` - See Argument Reference above.
+* `admin_state_up` - See Argument Reference above.
+* `tenant_id` - See Argument Reference above.
+* `subports` - See Argument Reference above
```
jtopjian: `sub_port`

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+				ForceNew: false,
+				Computed: true,
+			},
+			"tenant_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"sub_ports": &schema.Schema{
```
jtopjian: This might read better if this was called `sub_port` instead of `sub_ports`:

```hcl
resource "openstack_networking_trunk_v2" "foo" {
  ...
  sub_port {
    segmentation_id = 1
  }

  sub_port {
    segmentation_id = 2
  }
}
```

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error deleting OpenStack Neutron trunk: %s", err)
+	}
+
+	d.SetId("")
```
jtopjian: nit: This is no longer needed. It used to be in older versions of Terraform and might be hanging around some resources.

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,341 @@
...
+	return nil
+}
+
+func resourceSubportsMap(d *schema.Set) (subports map[trunks.Subport]bool) {
```
jtopjian: nit: let's make this function name more specific to the resource. Something like:

`resourceNetworkingV2TrunkSubportsMap`

Same for the other function names, too. It helps prevent collisions under a single namespace.

openstack/resource_openstack_compute_instance_v2_test.go
celebdor: Done. Thanks.

openstack/resource_openstack_networking_trunk_v2.go
celebdor: Done

openstack/resource_openstack_networking_trunk_v2.go
celebdor: Done, thanks.

openstack/resource_openstack_networking_trunk_v2.go
celebdor: Tags are not key value like:
```  labels {
    kube_tag                         = "1.11.1"
    prometheus_monitoring            = "true"
    influx_grafana_dashboard_enabled = "true"
    kube_dashboard_enabled           = "true"
  }
```

Rather, they are like:
```
{
   "port":{
      "status":"DOWN",
      "created_at":"2018-10-16T20:25:12Z",
      "description":"",
      "allowed_address_pairs":[

      ],
      "tags":[
         "bar",
         "foo"
      ],
      "network_id":"360451b6-d2d2-4b9d-b1cb-f6fe8254103c",
      "tenant_id":"e153f3f9082240a5974f667cfe1036e3",
      "extra_dhcp_opts":[

      ],
      "admin_state_up":true,
      "updated_at":"2018-10-17T09:11:26Z",
      "name":"subport_3",
      "device_owner":"",
      "revision_number":3,
      "mac_address":"fa:16:3e:29:2b:be",
      "binding:vnic_type":"normal",
      "port_security_enabled":true,
      "project_id":"e153f3f9082240a5974f667cfe1036e3",
      "fixed_ips":[

      ],
      "id":"f1d082f2-aa3d-4955-a400-ceea7c5ffc28",
      "security_groups":[
         "ca55e00b-75dd-4292-b76d-977e133f9689"
      ],
      "device_id":""
   }
}```

File: openstack/resource_openstack_networking_trunk_v2.go (hunk 5 above)
Comment (celebdor) [reply openstack/resource_openstack_networking_trunk_v2.go]: Great idea!

File: openstack/resource_openstack_networking_trunk_v2.go (hunk 10 above)
Comment (celebdor) [reply openstack/resource_openstack_networking_trunk_v2.go]: Great. Thanks

File: website/docs/r/networking_trunk_v2.html.markdown (hunk 7 above)
Comment (celebdor) [reply website/docs/r/networking_trunk_v2.html.markdown]: Done!

File: website/docs/r/networking_trunk_v2.html.markdown (hunk 8 above)
Comment (celebdor) [reply website/docs/r/networking_trunk_v2.html.markdown]: Done!

File: website/docs/r/networking_trunk_v2.html.markdown (hunk 6 above)
Comment (celebdor) [reply website/docs/r/networking_trunk_v2.html.markdown]: Done!

**Comment by jtopjian** [comment]
> Acceptance testing should catch any side effects.

And sure enough, unit and acceptance testing caught a problem. I recently merged https://github.com/gophercloud/gophercloud/pull/1294 so I can fix a Terraform issue. We'll have to get that issue fixed before this PR can be merged. We should be able to have that done in the next day or two.

**Comment by celebdor** [comment]
I saw compilation errors that needed this unrelated change:

```
--- a/openstack/resource_openstack_lb_member_v2.go
+++ b/openstack/resource_openstack_lb_member_v2.go
@@ -99,12 +99,13 @@ func resourceMemberV2Create(d *schema.ResourceData, meta interface{}) error {
        }
 
        adminStateUp := d.Get("admin_state_up").(bool)
+        weight := d.Get("weight").(int)
        createOpts := pools.CreateMemberOpts{
                Name:         d.Get("name").(string),
                TenantID:     d.Get("tenant_id").(string),
                Address:      d.Get("address").(string),
                ProtocolPort: d.Get("protocol_port").(int),
-               Weight:       d.Get("weight").(int),
+               Weight:       &weight,
                AdminStateUp: &adminStateUp,
        }
 
@@ -186,7 +187,8 @@ func resourceMemberV2Update(d *schema.ResourceData, meta interface{}) error {
                updateOpts.Name = d.Get("name").(string)
        }
        if d.HasChange("weight") {
-               updateOpts.Weight = d.Get("weight").(int)
+               weight := d.Get("weight").(int)
+               updateOpts.Weight = &weight
        }
        if d.HasChange("admin_state_up") {
                asu := d.Get("admin_state_up").(bool)
```

**Comment by jtopjian** [comment]
> I saw compilation errors that needed this unrelated change:

Right. It's best to not bundle this change into this PR. I'll have a separate PR shortly.

The reason for the upstream change is for #392

**Comment by celebdor** [comment]
Cool. Thanks!

**Comment by celebdor** [comment]
@jtopjian In the meantime, should I add support for setting tags for the trunk resource, now that gophercloud supports it, or is it better to make a second PR after this one is merged?

**Comment by jtopjian** [comment]
@celebdor sorry for the delay. Since this is a brand new resource that hasn't been merged, sure, bring in the new Gophercloud code and add Tags :)

If this was an existing resource and you were adding two features, then 2 PRs would be better. But that's not the case here.

File: openstack/resource_openstack_networking_trunk_v2.go (hunk 3 above)
Comment (jtopjian) [reply openstack/resource_openstack_networking_trunk_v2.go]: Oh duh. I'm sorry - I was mistaking tags for metadata.

File: openstack/resource_openstack_networking_trunk_v2.go [hunk 12]
```diff
@@ -0,0 +1,342 @@
...
+	value := false
+
+	if v, ok := d.GetOkExists("admin_state_up"); ok {
+		value = v.(bool)
+	}
+
+	return &value
+}
+
+func resourceTags(d *schema.ResourceData) (tags []string) {
```
jtopjian: It's very possible this function can be used for any of the future network tags field, so how about we move it to the `util.go` file? Let's call it `networkV2AttributesTags` since there could be another tag implementation in another OpenStack project that does tags differently.

hardys: Agreed this will be useful for other resources - I have a WIP branch at https://github.com/hardys/terraform-provider-openstack/tree/neutron_tags which does as suggested by @jtopjian but since @celebdor beat me to it, I'll rebase on this PR :)

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,342 @@
...
+		return CheckDeleted(d, err, "trunk")
+	}
+	log.Printf("[DEBUG] Retrieved trunk %s: %+v", d.Id(), trunk)
+
+	d.Set("region", GetRegion(d, config))
+	d.Set("name", trunk.Name)
+	d.Set("port_id", trunk.PortID)
+	d.Set("admin_state_up", trunk.AdminStateUp)
+	d.Set("tenant_id", trunk.TenantID)
+	d.Set("tags", trunk.Tags)
```
hardys: Perhaps I missed it, but did you add any test proving the set/get/update of tags?

If we're adding tags in this commit it'd be nice to test them (particularly since I've run into a few issues with the acceptance test locally and I'd like to confirm if it'd my code or a problem that will also occur with this patch).

celebdor:
I'm pushing the test for flags now

celebdor:
@jtopjian @hardys Note that I also added an acceptance check in utils.go that I think could be used for the other resources.

jtopjian:
@celebdor Separate from the review I just submitted, after reading through this resource and running the acceptance tests, I'm wondering if we should break this into two separate resources:

* `openstack_networking_trunk_v2`
* `openstack_networking_trunk_subport_v2`

So an implementation would look like:

```hcl
resource "openstack_networking_trunk_v2" "trunk_1" {          
  name = "trunk_1"                                            
  port_id = "${openstack_networking_port_v2.parent_port_1.id}"
  admin_state_up = "true"                                     
}                                                             

resource "openstack_networking_trunk_subport_v2" "subport_1" {
  trunk_id = "${openstack_networking_trunk_v2.trunk_1.id}"
  port_id = "${openstack_networking_port_v2.subport_1.id}"
  segmentation_id = 1
  segmentation_type = "vlan"
}

resource "openstack_networking_trunk_subport_v2" "subport_2" {
  trunk_id = "${openstack_networking_trunk_v2.trunk_1.id}"
  port_id = "${openstack_networking_port_v2.subport_2.id}"
  segmentation_id = 2
  segmentation_type = "vlan"
}
```

The disadvantage is that it's more verbose and that you won't be able to add subports during the initial creation of the trunk.

However, the advantage is that a separate resource is more flexible: users can add and remove subports using Terraform's native HCL logic and users can more easily reference individual subports. For example, here's what 3 subports look like in the Terraform state:

```
sub_port.# = 3
sub_port.1580179709.port_id = b5689463-1426-44d6-a8fc-1efe56ac389b
sub_port.1580179709.segmentation_id = 1
sub_port.1580179709.segmentation_type = vlan
sub_port.272728530.port_id = 9852a0a1-b47a-4730-9d23-0d0df1fba9a7
sub_port.272728530.segmentation_id = 3
sub_port.272728530.segmentation_type = vlan
sub_port.504626688.port_id = f304bf3b-175d-4d5a-b68a-e873136b96b2
sub_port.504626688.segmentation_id = 4
sub_port.504626688.segmentation_type = vlan
```

The index is a computed hash that is not predictable and makes it near impossible for a user to reference it with `${openstack_networking_trunk_v2.sub_ports.123456.segmentation_type}`.

Historically, any time we have had a "sub-resource" like this in a parent resource, users end up using all sorts of hacks to better manage the sub-resource. Two good examples are:

* https://github.com/terraform-providers/terraform-provider-openstack/issues/170
* https://github.com/terraform-providers/terraform-provider-openstack/issues/174

Terraform 0.12 _should_ provide a way to more easily handle these situations, but if the resources were designed to be more "composable" from the start, there wouldn't be an issue. As an aside, doing a _very_ large refactor of the `openstack_networking_port_v2` resource is on our TODO list to remedy this exact situation :)

Thoughts?

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -0,0 +1,354 @@
...
+		log.Printf("[DEBUG] Updating trunk %s with options: %+v", d.Id(), updateOpts)
+		_, err = trunks.Update(client, d.Id(), updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating OpenStack Neutron trunk: %s", err)
+		}
+	}
+
+	var subportsToAdd trunks.AddSubportsOpts
+	var subportsToRemove trunks.RemoveSubportsOpts
+	if d.HasChange("sub_port") {
```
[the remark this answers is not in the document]
celebdor: That's great! I was trying to find set operations and implemented this cause I didn't find them. Mush better this way!

openstack/resource_openstack_networking_trunk_v2.go
celebdor: There wasn't at the time, now it is:
TestAccNetworkingV2Trunk_tags

celebdor:
> @celebdor Separate from the review I just submitted, after reading through this resource and running the acceptance tests, I'm wondering if we should break this into two separate resources:
> 
>     * `openstack_networking_trunk_v2`
> 
>     * `openstack_networking_trunk_subport_v2`
> 
> 
> So an implementation would look like:
> 
> ```terraform
> resource "openstack_networking_trunk_v2" "trunk_1" {          
>   name = "trunk_1"                                            
>   port_id = "${openstack_networking_port_v2.parent_port_1.id}"
>   admin_state_up = "true"                                     
> }                                                             
> 
> resource "openstack_networking_trunk_subport_v2" "subport_1" {
>   trunk_id = "${openstack_networking_trunk_v2.trunk_1.id}"
>   port_id = "${openstack_networking_port_v2.subport_1.id}"
>   segmentation_id = 1
>   segmentation_type = "vlan"
> }
> 
> resource "openstack_networking_trunk_subport_v2" "subport_2" {
>   trunk_id = "${openstack_networking_trunk_v2.trunk_1.id}"
>   port_id = "${openstack_networking_port_v2.subport_2.id}"
>   segmentation_id = 2
>   segmentation_type = "vlan"
> }
> ```
> 
> The disadvantage is that it's more verbose and that you won't be able to add subports during the initial creation of the trunk.
> 
> However, the advantage is that a separate resource is more flexible: users can add and remove subports using Terraform's native HCL logic. For example:
> 
> ```terraform
> resource "openstack_networking_trunk_v2" "trunk_1" {          
>   name           = "trunk_1"                                            
>   port_id        = "${openstack_networking_port_v2.parent_port_1.id}"
>   admin_state_up = "true"                                     
> }                                                             
> 
> resource "openstack_networking_trunk_subport_v2" "subports" {
>   count             = 3
>   trunk_id          = "${openstack_networking_trunk_v2.trunk_1.id}"
>   port_id           = "${openstack_networking_port_v2.ports.*.id[count.index]}"
>   segmentation_id   = "${4000 + count.index}"
>   segmentation_type = "vlan"
> }
> ```
> 
> and users can more easily reference individual subports. For example, here's what 3 subports look like in the Terraform state:
> 
> ```
> sub_port.# = 3
> sub_port.1580179709.port_id = b5689463-1426-44d6-a8fc-1efe56ac389b
> sub_port.1580179709.segmentation_id = 1
> sub_port.1580179709.segmentation_type = vlan
> sub_port.272728530.port_id = 9852a0a1-b47a-4730-9d23-0d0df1fba9a7
> sub_port.272728530.segmentation_id = 3
> sub_port.272728530.segmentation_type = vlan
> sub_port.504626688.port_id = f304bf3b-175d-4d5a-b68a-e873136b96b2
> sub_port.504626688.segmentation_id = 4
> sub_port.504626688.segmentation_type = vlan
> ```
> 
> The index is a computed hash that is not predictable and makes it near impossible for a user to reference it with `${openstack_networking_trunk_v2.sub_ports.123456.segmentation_type}`.
> 
> Historically, any time we have had a "sub-resource" like this in a parent resource, users end up using all sorts of hacks to better manage the sub-resource. Two good examples are:
> 
>     * #170
> 
>     * #174
> 
> 
> Terraform 0.12 _should_ provide a way to more easily handle these situations, but if the resources were designed to be more "composable" from the start, there wouldn't be an issue. As an aside, doing a _very_ large refactor of the `openstack_networking_port_v2` resource is on our TODO list to remedy this exact situation :)
> 
> Thoughts?

The HCL argument for using count on ports convinced me to give it a try. I'll try to get it done over the weekend and see how it looks like.

celebdor:
> The HCL argument for using count on ports convinced me to give it a try. I'll try to get it done over the weekend and see how it looks like.

I'm worried about having a separate resource meaning that we'll need to add each subport one by one instead of being able to do it in a single add_subports remove_subports operation. This would add a significant load on Neutron.

> 
> Thanks a lot for the background info and all the improvement proposals. They are much, much appreciated!

openstack/resource_openstack_networking_trunk_v2.go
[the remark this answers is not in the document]
jtopjian: I just realized that `len(toAdd.List())` can be changed to `toAdd.Len()`

jtopjian:
> I'm worried about having a separate resource meaning that we'll need to add each subport one by one instead of being able to do it in a single add_subports remove_subports operation. This would add a significant load on Neutron.

Good point - this is something to consider.

There are 3 possible execution methods in play here:

1. One single resource bundles all subport changes into 1 API call

2. Two separate resources would cause all subport changes to happen _in parallel_ in separate API calls, each call performing an action on 1 subport. This is because Terraform is parallel in nature.  
  
    For the example with `count = 3`, 3 parallel `add_subports` calls would happen, one for each subport. In my opinion, this execution method will cause the most possible load/problems with Neutron since parallel requests hitting the ML2 driver to do the actual physical work on the NIC/vNIC might cause concurrency issues. 

    A similar existing example might be the `openstack_networking_secgroup_v2` and `openstack_networking_secgroup_rule_v2` resources. They execute in the same, parallel fashion and we haven't had any reported problems. That's not to say problems can't happen, though.

3. We can place a [mutex](https://golang.org/pkg/sync/) on the `openstack_networking_trunk_subport_v2` resource to _enforce_ it to run serial. This would be equivalent to a user doing `openstack network trunk set --subport` one by one.

The main reason I proposed a separate resource is to enable easier composition of trunks and subports using HCL. But if that is going to cause problems with the Neutron service itself, it might be a bad idea. As I mentioned, Terraform 0.12 should be able to handle [composition within a resource](https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each), so perhaps we can consider this a non-issue.

The second benefit of a separate resource is to get around the inability to reference a subport using the auto-generated index. Again, maybe Terraform 0.12 can [help with this](https://www.hashicorp.com/blog/terraform-0-12-generalized-splat-operator), but I'm not 100% sure yet. I'm OK with considering this a non-blocking issue at the moment.

so tl;dr: If there is serious concern about overloading the API and serializing the requests (method 3) isn't good, let's proceed with a single resource. The `sub_port` argument in `openstack_networking_trunk_v2` isn't "computed" and that will allow us an easy path of introducing a `openstack_networking_trunk_subport_v2` resource in the future if there's enough user feedback for it.

celebdor:
I addressed the comments and fixed the build:

```
[stack@kuryr-a32c-02 openstack]$ TF_ACC=1 go test . -v --run=TestAccNetworkingV2Trunk_ -count=1 -timeout 120m
=== RUN   TestAccNetworkingV2Trunk_nosubports
--- PASS: TestAccNetworkingV2Trunk_nosubports (49.46s)
=== RUN   TestAccNetworkingV2Trunk_subports
--- PASS: TestAccNetworkingV2Trunk_subports (54.50s)
=== RUN   TestAccNetworkingV2Trunk_tags
--- PASS: TestAccNetworkingV2Trunk_tags (51.55s)
=== RUN   TestAccNetworkingV2Trunk_trunkUpdateSubports
--- PASS: TestAccNetworkingV2Trunk_trunkUpdateSubports (66.52s)
=== RUN   TestAccNetworkingV2Trunk_computeInstance
--- PASS: TestAccNetworkingV2Trunk_computeInstance (99.29s)
PASS
```

In regards to the subport options. After discussing it a bit with my team, we feel like:

* Option 2 is going to cause Neutron server load problems indeed, so we'd rather avoid that
* Option 3 with serialization via mutex would make creating big pools slow.
* Option 1 is missing count to be able to create trunk pools, but we hope that Terraform 0.12 will come soon and it will make the issue go away.

I suppose there is really no way to bundle resource creation for bulk creation/deletion in terraform (note that this would also be useful for the port resource, that supports bulk operations in the neutron api), is there?

website/openstack.erb
```diff
@@ -67,6 +67,9 @@
             <li<%= sidebar_current("docs-openstack-datasource-networking-subnetpool-v2") %>>
               <a href="/docs/providers/openstack/d/networking_subnetpool_v2.html">openstack_networking_subnetpool_v2</a>
             </li>
+            <li<%= sidebar_current("docs-openstack-datasource-networking-trunk-v2") %>>
+              <a href="/docs/providers/openstack/d/networking_trunk_v2.html">openstack_networking_trunk_v2</a>
```
jtopjian: This is linking to `/d/` which is for data sources. It'll need moved further below to the `/r/` sections. Travis will flag an error if the link is incorrect: https://travis-ci.org/terraform-providers/terraform-provider-openstack/builds/444878986#L554-L555

openstack/resource_openstack_networking_trunk_v2_test.go
```diff
@@ -0,0 +1,690 @@
...
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+)
+
+func TestAccNetworkingV2Trunk_nosubports(t *testing.T) {
+	var port_1 ports.Port
+	var trunk_1 trunks.Trunk
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
```
jtopjian: Similar with Gophercloud, it looks like OpenLab's devstack configuration does not enable the trunk extension at this time, so acceptance tests are always going to fail. Normally in this situation, we add a new environment variable [here](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider_test.go#L15) to toggle the valid type of configuration, but this seems excessive for the trunk extension. 

Instead, let's change this to an admin-only test (which won't run under Terraform / OpenLab at this time) and I will ask the OpenLab team to enable the trunk extension.

Let's change all `PreCheck` entries to:

```go
PreCheck: func() {           
  testAccPreCheck(t)         
  testAccPreCheckAdminOnly(t)
},                           
```

openstack/util.go
```diff
@@ -187,3 +191,47 @@ func containerInfraLabelsStringV1(d *schema.ResourceData) (string, error) {
...
+	rawTags := d.Get("tags").(*schema.Set).List()
+	tags = make([]string, len(rawTags))
+
+	for i, raw := range rawTags {
+		tags[i] = raw.(string)
+	}
+	return
+}
+
+func testAccCheckNetworkingV2Tags(name string, key string, tags []string) resource.TestCheckFunc {
```
hardys: Small nit - we could probably remove the key parameter here, since it's always going to be "tags"?

celebdor: Sure, why not :+1:

openstack/resource_openstack_networking_trunk_v2_test.go
celebdor: Got it! Will do!

celebdor:
We can merge. Thanks!

PR #447, opened 2018-10-16
Allow to specify flavor access for tenants

kirecek:
This PR adds `access` attribute to flavor resource. `access` contains list of tenant ids that are allowed to use flavor.

EDIT: create `openstack_compute_flavor_access_v2` instead of access list in flavor resource.

### DIFF
--- a/openstack/import_openstack_compute_flavor_access_v2_test.go
+++ b/openstack/import_openstack_compute_flavor_access_v2_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
-func TestAccComputeV2FloatingIPAssociate_importBasic(t *testing.T) {
+func TestAccComputeV2FlavorAccess_importBasic(t *testing.T) {
 	resourceName := "openstack_compute_flavor_access_v2.access_1"
 
 	flavorName := fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
--- a/openstack/import_openstack_compute_flavor_access_v2_test.go
+++ b/openstack/import_openstack_compute_flavor_access_v2_test.go
@@ -1,0 +1,33 @@
+package openstack
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccComputeV2FloatingIPAssociate_importBasic(t *testing.T) {
+	resourceName := "openstack_compute_flavor_access_v2.access_1"
+
+	flavorName := fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
+	projectName := fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeV2FlavorAccessDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeV2FlavorAccess_basic(flavorName, projectName),
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
--- a/openstack/import_openstack_compute_flavor_access_v2_test.go
+++ b/openstack/import_openstack_compute_flavor_access_v2_test.go
@@ -15,7 +15,10 @@ func TestAccComputeV2FlavorAccess_importBasic(t *testing.T) {
 	projectName := fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
 
 	resource.Test(t, resource.TestCase{
-		PreCheck:     func() { testAccPreCheck(t) },
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckAdminOnly(t)
+		},
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckComputeV2FlavorAccessDestroy,
 		Steps: []resource.TestStep{
--- a/openstack/resource_openstack_compute_flavor_access_v2.go
+++ b/openstack/resource_openstack_compute_flavor_access_v2.go
@@ -77,7 +77,7 @@ func resourceComputeFlavorAccessV2Read(d *schema.ResourceData, meta interface{})
 
 	flavorAccess, err := getFlavorAccess(computeClient, d)
 	if err != nil {
-		return fmt.Errorf("Error getting flavor access: %s", err)
+		return CheckDeleted(d, err, "Error getting flavor access")
 	}
 
 	d.Set("region", GetRegion(d, config))
--- a/openstack/resource_openstack_compute_flavor_access_v2.go
+++ b/openstack/resource_openstack_compute_flavor_access_v2.go
@@ -33,7 +33,6 @@ func resourceComputeFlavorAccessV2() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-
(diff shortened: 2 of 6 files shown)


jtopjian:
@kirecek Thank you for submitting this.

What are your thoughts on creating a dedicated resource for this? Something like:

```hcl
resource "openstack_compute_flavor_access_v2" "foo" {
  flavor_id  = "${openstack_compute_flavor_v2.flavor_1.id}"
  project_id = "${openstack_identity_project_v3.project_1.id}"
}
```

It would be a small and simple resource, but when joining two resources together, this pattern has proven to be more flexible than adding a new argument to one resource. 

The following similar resources are a good template to use:

* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_floatingip_associate_v2.go
* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_volume_attach_v2.go
* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_networking_floatingip_associate_v2.go

kirecek:
@jtopjian yeah, thought also about this approach. Decided to go this way just because I wanted to reduce amount of recources. Now when I think about this, currently there is no way how to add access to already existing flavors.

So yeah, I'll create a new `openstack_compute_flavor_access_v2` resource. Thanks.

kirecek:
@ozerovandrei thank you for the review. I've updated the code and added some docs. Please check it out if it's alright :)

openstack/resource_openstack_compute_flavor_access_v2.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+func resourceComputeFlavorAccessV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceComputeFlavorAccessV2Create,
+		Read:   resourceComputeFlavorAccessV2Read,
+		Delete: resourceComputeFlavorAccessV2Delete,
+		Importer: &schema.ResourceImporter{
```
jtopjian: This will allow users the ability to import this resource - which is good! But we'll want to include a test for this in a dedicated file like this:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/import_openstack_compute_floatingip_associate_v2_test.go

And then amend the docs with this section: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/docs/r/compute_floatingip_associate_v2.html.markdown#import

openstack/resource_openstack_compute_flavor_access_v2.go
```diff
@@ -0,0 +1,149 @@
...
+	flavorAccess, err := getFlavorAccess(computeClient, d)
+	if err != nil {
+		return fmt.Errorf("Error getting flavor access: %s", err)
+	}
+
+	removeAccessOpts := flavors.RemoveAccessOpts{Tenant: flavorAccess.TenantID}
+	log.Printf("[DEBUG] RemoveAccess Options: %#v", removeAccessOpts)
+
+	if _, err := flavors.RemoveAccess(computeClient, flavorAccess.FlavorID, removeAccessOpts).Extract(); err != nil {
+		return fmt.Errorf("Error remove tenant %s access from flavor %s: %s", flavorAccess.TenantID, flavorAccess.FlavorID, err)
```
jtopjian: "Error removing"

openstack/import_openstack_compute_flavor_access_v2_test.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func TestAccComputeV2FlavorAccess_importBasic(t *testing.T) {
+	resourceName := "openstack_compute_flavor_access_v2.access_1"
+
+	flavorName := fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
+	projectName := fmt.Sprintf("ACCPTTEST-%s", acctest.RandString(5))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
```
ozerovandrei: This test needs `testAccPreCheckAdminOnly(t)` too.

openstack/resource_openstack_compute_flavor_access_v2_test.go
```diff
@@ -0,0 +1,152 @@
...
+			for _, a := range accessList {
+				if a.TenantID == tid {
+					return false, fmt.Errorf("Flavor Access still exists")
+				}
+			}
+
+			return true, nil
+		})
+
+		if err != nil {
```
jtopjian: This should be:

```go
if err != nil {                                    
    return nil                                     
  }                                                
  return err                                       
}                                                  
```

Because a 404 can be returned when there are no access entries associated with a flavor.

openstack/resource_openstack_compute_flavor_access_v2_test.go
```diff
@@ -0,0 +1,152 @@
...
+    resource "openstack_compute_flavor_v2" "flavor_1" {
+      name = "%s"
+      ram = 512
+      vcpus = 1
+      disk = 5
+
+      is_public = false
+    }
+
+    resource "openstack_identity_project_v3" "tenant_1" {
```
jtopjian: `project_1` instead of `tenant_1`

openstack/resource_openstack_compute_flavor_access_v2_test.go
```diff
@@ -0,0 +1,152 @@
...
+      name = "%s"
+    }
+
+    resource "openstack_compute_flavor_access_v2" "access_1" {
+      flavor_id = "${openstack_compute_flavor_v2.flavor_1.id}"
+      tenant_id = "${openstack_identity_project_v3.tenant_1.id}"
```
jtopjian: `project_1` instead of `tenant_1`

openstack/resource_openstack_compute_flavor_access_v2.go
```diff
@@ -0,0 +1,149 @@
...
+func resourceComputeFlavorAccessV2Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	computeClient, err := config.computeV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
+	}
+
+	flavorAccess, err := getFlavorAccess(computeClient, d)
+	if err != nil {
+		return fmt.Errorf("Error getting flavor access: %s", err)
```
jtopjian: Instead do:

```go
return CheckDeleted(d, err, "Error getting flavor access")
```

kirecek:
@jtopjian  thanks. I've rebased the repo and made changes you have requested. Hope it will pass. Sorry for messy PR :disappointed: 

hmmm, I don't understand what does the latest message from openlab bot mean. `Please rebase the change and upload a new patchset`

jtopjian:
> @jtopjian thanks. I've rebased the repo and made changes you have requested. Hope it will pass. Sorry for messy PR

Absolutely no problem at all. Working through this pointed out some errors in other tests, so it helped. Thank you for your patience with this.

> hmmm, I don't understand what does the latest message from openlab bot mean. Please rebase the change and upload a new patchset

That's ignorable. OpenLab might have been having some problems.

openstack/resource_openstack_compute_flavor_access_v2.go
```diff
@@ -0,0 +1,149 @@
...
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+			"flavor_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
```
ozerovandrei: Can you please remove this blank line (or add a blank line between `region` and `flavor_id`)?
So we will have unified style in the `Schema`.

kirecek: done

ozerovandrei:
recheck

PR #451, opened 2018-10-17
 LBaaS v2: Allow member weights of 0

jtopjian:
For #392 
Supersedes #393


jtopjian:
@ozerovandrei Thanks!

Issue #453, opened 2018-10-22
Add support for neutron tags


hardys:
Recently added gophercloud support for the neutron attributetags extension means we should be able to specify tags for neutron resources that support attributetags, so the created resources can be optionally tagged during resource create/update via terraform.

For example:

resource "openstack_networking_network_v2" "terraform" {
  name           = "terraform_test"
  admin_state_up = "true"
  tags = "${var.tags}"
}

tags should be optional and accept a list of strings - this is useful when you want to group neutron resources then later query for them by tag using the neutron API and server-side filtering by tags.

### Affected Resource(s)

The affected resources will be the same as added to gophercloud via https://github.com/gophercloud/gophercloud/pull/1286, e.g:

Router: https://github.com/openstack/neutron/blob/master/neutron/db/models/l3.py#L67
FloatingIP: https://github.com/openstack/neutron/blob/master/neutron/db/models/l3.py#L112
SecurityGroup: https://github.com/openstack/neutron/blob/master/neutron/db/models/securitygroup.py#L32
Port: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L115
Subnet: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L210
SubnetPool: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L250
Network: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L278
~~QoS/RBAC Policy: https://github.com/openstack/neutron/blob/master/neutron/db/qos/models.py#L36~~
Trunk: https://github.com/openstack/neutron/blob/master/neutron/services/trunk/models.py#L49

Note that Trunk tag support is being added via https://github.com/terraform-providers/terraform-provider-openstack/pull/446 so this issue can track adding tag support to the remaining resources.

### Checklist

- [x] Router (PR #467)
- [x] FloatingIP (PR #466)
- [x] SecurityGroup (PR #463)
- [x] Port (PR #461)
- [x] Subnet (PR #459)
- [x] SubnetPool (PR #460)
- [x] Network (PR #454)
- [x] ~~QoS/RBAC Policy~~ (Not currently supported in terraform-provider-openstack AFAICS)
- [x] Trunk (PR #446)

### References
https://github.com/gophercloud/gophercloud/pull/1286
https://docs.openstack.org/neutron/pike/contributor/internals/tag.html
https://github.com/gophercloud/gophercloud/issues/1285
[...]

PR #454, opened 2018-10-22
Add tags support for Neutron Networks

hardys:
This adds support for tags to networks, note this is branched from PR #446 since that already adds initial support for tags for the trunk resource, and a couple of functions are reused here.

Subsequent PRs will add support for the other neutron resources which support standard attribute tags as described in Issue #453 

Issue #453

### DIFF
--- a/website/docs/r/networking_network_v2.html.markdown
+++ b/website/docs/r/networking_network_v2.html.markdown
@@ -95,6 +95,8 @@ The following arguments are supported:
     so that they are scheduled on different availability zones. Changing this 
     creates a new network.
 
+* `tags` - (Optional) A set of string tags for the network. 
+
 The `segments` block supports:
 
 * `physical_network` - The phisical network where this network is implemented.
@@ -112,6 +114,7 @@ The following attributes are exported:
 * `tenant_id` - See Argument Reference above.
 * `admin_state_up` - See Argument Reference above.
 * `availability_zone_hints` - See Argument Reference above.
+* `tags` - See Argument Reference above.
 
 ## Import
 
--- a/openstack/resource_openstack_networking_network_v2.go
+++ b/openstack/resource_openstack_networking_network_v2.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/external"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/provider"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
@@ -94,6 +95,11 @@ func resourceNetworkingNetworkV2() *schema.Resource {
 				Type:     schema.TypeMap,
 				Optional: true,
 				ForceNew: true,
+			},
+			"tags": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"availability_zone_hints": &schema.Schema{
 				Type:     schema.TypeList,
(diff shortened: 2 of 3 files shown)


hardys:
Marking this WIP because the trunk patch from @celebdor has some build failures breaking the tests - when that's resolved I'll rebase this but it's otherwise ready for review.

If the approach taken here is OK I'll go ahead and push PRs/patches for the remaining resources using a similar pattern @jtopjian would you prefer one PR per resource or adding commits to this PR for all resources?  I assume the former similar to gophercloud but wanted to check.

openstack/util.go
```diff
@@ -187,3 +191,47 @@ func containerInfraLabelsStringV1(d *schema.ResourceData) (string, error) {
...
+	rawTags := d.Get("tags").(*schema.Set).List()
+	tags = make([]string, len(rawTags))
+
+	for i, raw := range rawTags {
+		tags[i] = raw.(string)
+	}
+	return
+}
+
+func testAccCheckNetworkingV2Tags(name string, key string, tags []string) resource.TestCheckFunc {
```
hardys: Small nit - now this function has been renamed we could probably drop the key parameter, as it'll always be "tags" I think?

hardys:
Note I've added the test for network to the per-resource test, but that will end up a little inefficient as we end up creating a network for other tests e.g subnet/port etc as well.

This is kind of a side-effect of the one PR per resource requirement I think, but I could probably refactor the test at the end of the series to a single template which creates all-the-resources that support tags?

hardys:
Rebased on master now the trunk patch landed, also fixed a minor issue in the docs markdown.

jtopjian:
@hardys I'm going to be away for the next few days and might not get a chance to review this in detail. I'll do it as soon as I can, though.

> This is kind of a side-effect of the one PR per resource requirement I think, but I could probably refactor the test at the end of the series to a single template which creates all-the-resources that support tags?

Good idea :)

hardys:
> Good idea :)

Ok I'll look into it as the current tests are a bit slow:

```
[shardy@localhost terraform-provider-openstack]$ TF_ACC=1 go test -v -run "TestAccNetworkingV2.*_tags" github.com/terraform-providers/terraform-provider-openstack/openstack -count=1
=== RUN   TestAccNetworkingV2FloatingIP_tags
--- PASS: TestAccNetworkingV2FloatingIP_tags (34.73s)
=== RUN   TestAccNetworkingV2Network_tags
--- PASS: TestAccNetworkingV2Network_tags (28.52s)
=== RUN   TestAccNetworkingV2Port_tags
--- PASS: TestAccNetworkingV2Port_tags (65.28s)
=== RUN   TestAccNetworkingV2Router_tags
--- PASS: TestAccNetworkingV2Router_tags (30.82s)
=== RUN   TestAccNetworkingV2SecGroup_tags
--- PASS: TestAccNetworkingV2SecGroup_tags (21.24s)
=== RUN   TestAccNetworkingV2Subnet_tags
--- PASS: TestAccNetworkingV2Subnet_tags (44.90s)
=== RUN   TestAccNetworkingV2SubnetPool_tags
--- PASS: TestAccNetworkingV2SubnetPool_tags (24.06s)
=== RUN   TestAccNetworkingV2Trunk_tags
--- PASS: TestAccNetworkingV2Trunk_tags (63.71s)
PASS
ok  	github.com/terraform-providers/terraform-provider-openstack/openstack	313.268s
```

I guess this duplication isn't unique to the tags test, but if I refactor things to have a new e.g TestAccNetworkingV2_tags common test we can at least remove the Network/Subnet test overhead (since those are already created for the Port test) which is over 1min and I suspect using a single terraform invocation/template will reduce the overall test overhead quite a bit as well.

hardys:
Ok I refactored the tests for the entire series and it's a lot faster:

```
[shardy@localhost openstack]$ TF_ACC=1 go test -v -run "TestAccNetworkingV2_tags" github.com/terraform-providers/terraform-provider-openstack/openstack -count=1
=== RUN   TestAccNetworkingV2_tags
--- PASS: TestAccNetworkingV2_tags (94.14s)
PASS
ok  	github.com/terraform-providers/terraform-provider-openstack/openstack	94.154s
```

hardys:
Ok the latest revision has a separate test for tags, and the series at https://github.com/hardys/terraform-provider-openstack/commits/neutron_tags_wip adds to the test which speeds things up considerably.

When this passes CI I'll push a couple more of the pending patches as dependent PRs as I think that neutron_tags_wip branch is all ready for review now.

hardys:
recheck

website/docs/r/networking_network_v2.html.markdown
```diff
@@ -112,6 +114,7 @@ The following attributes are exported:
 * `tenant_id` - See Argument Reference above.
 * `admin_state_up` - See Argument Reference above.
 * `availability_zone_hints` - See Argument Reference above.
+* `tags` - Argument Reference above.
```
ozerovandrei: Let's keep it similar to other attributes: "See Argument Reference above.".

hardys:
@ozerovandrei thanks for the review - I fixed the docs nit in the latest revision.

PR #459, opened 2018-10-29
Add tags support for Neutron Subnets

hardys:
This adds support for tags to subnets note this is branched from PR #454 

Subsequent PRs will add support for the other neutron resources which support standard attribute tags as described in Issue #453

Issue #453

### DIFF
--- a/openstack/resource_openstack_networking_tags_v2_test.go
+++ b/openstack/resource_openstack_networking_tags_v2_test.go
@@ -1,0 +1,49 @@
+package openstack
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccNetworkingV2_tags(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNetworkingV2NetworkDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccNetworkingV2_tags,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNetworkingV2Tags(
+						"openstack_networking_network_v2.network_1",
+						[]string{"a", "b", "c"}),
+				),
+			},
+			resource.TestStep{
+				Config: testAccNetworkingV2_tags_update,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNetworkingV2Tags(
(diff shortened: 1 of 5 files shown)


openstack/resource_openstack_networking_tags_v2_test.go
```diff
@@ -13,37 +14,59 @@ func TestAccNetworkingV2_tags(t *testing.T) {
...
+
+const testAccNetworkingV2_tags_create = `["a", "b", "c"]`
+
+const testAccNetworkingV2_tags_update = `["a", "b", "c", "d"]`
+
+var testAccNetworkingV2_config_create = strings.Replace(
+	testAccNetworkingV2_config, "__TAGS__", testAccNetworkingV2_tags_create, -1)
+
+var testAccNetworkingV2_config_update = strings.Replace(
+	testAccNetworkingV2_config, "__TAGS__", testAccNetworkingV2_tags_update, -1)
```
jtopjian: Is it possible to use the `fmt.Sprintf` method commonly use in other tests? I understand interpolation like this is never a fun thing, but it'll keep consistency with everything else.

hardys: Sure I can, but FWIW I did it this way deliberately, for two reasons:

1. I wanted to avoid duplicating sprintf value arguments, but looking again at docs/google I see I can do "%[1]s" for the format placeholder to avoid that.

2. Replace is faster (about twice as fast), I suppose that's not a huge deal for this one test but I knew I'd be doing multiple replacements so wanted to avoid using the slowest possible approach ;)

[shardy@localhost join_test]$ go test --bench=.
goos: linux
goarch: amd64
BenchmarkReplace-4   	 5000000	       291 ns/op
BenchmarkSprintf-4   	 3000000	       522 ns/op

(benchmark at https://paste.fedoraproject.org/paste/mxXLGBLbVBeM3s3~U4O7fQ/)

I don't have a particularly strong opinion either way, but wanted to confirm that I had given this some thought already - I'll update it to Sprintf if that's what you prefer.

hardys:
@jtopjian thanks for the review!  Now adjusted to use fmt.Sprintf instead

openstack/resource_openstack_networking_tags_v2_test.go
jtopjian: > 1. I wanted to avoid duplicating sprintf value arguments, but looking again at docs/google I see I can do "%[1]s" for the format placeholder to avoid that.

TIL :)

Thanks!

jtopjian:
I did a normal (non-squash) merge here which automatically merged #454

PR #462, opened 2018-10-29
Move Neutron ports trunk test to common tags test

hardys:
This converts the existing trunk tags test to the new common test, which saves some time as we can reuse the existing network/subnet/port resources from other tests.

Issue #453


hardys:
Ah this is failing because testAccPreCheckAdminOnly(t) is used in the existing test to disable the acceptance test.

I'll abandon this and rebase the remaining patches, perhaps this can be revived when openlab trunk support is enabled.

PR #465, opened 2018-10-30
Enable object versioning

cryptobioz:
To enable versioning, we must set a HTTP header `X-History-Location` or `X-Versions-Location` while creating or updating the container (cf: [Openstack documentation](https://docs.openstack.org/swift/latest/overview_object_versioning.html)).
This PR adds support of both headers.

### DIFF
--- a/openstack/resource_openstack_objectstorage_container_v1.go
+++ b/openstack/resource_openstack_objectstorage_container_v1.go
@@ -113,6 +113,8 @@ func resourceObjectStorageContainerV1Create(d *schema.ResourceData, meta interfa
 		switch vParams["type"].(string) {
 		case "versions":
 			createOpts.VersionsLocation = vParams["location"].(string)
+		case "history":
+			createOpts.HistoryLocation = vParams["location"].(string)
 		}
 	}
 
@@ -166,11 +168,14 @@ func resourceObjectStorageContainerV1Update(d *schema.ResourceData, meta interfa
 	if versioning, ok := d.Get("versioning").(*schema.Set); ok && d.HasChange("versioning") {
 		if versioning.Len() == 0 {
 			updateOpts.RemoveVersionsLocation = "true"
+			updateOpts.RemoveHistoryLocation = "true"
 		} else {
 			vParams := versioning.List()[0].(map[string]interface{})
 			switch vParams["type"].(string) {
 			case "versions":
 				updateOpts.VersionsLocation = vParams["location"].(string)
+			case "history":
+				updateOpts.HistoryLocation = vParams["location"].(string)
 			}
 		}
 	}
--- a/openstack/resource_openstack_objectstorage_container_v1.go
+++ b/openstack/resource_openstack_objectstorage_container_v1.go
@@ -55,6 +55,11 @@ func resourceObjectStorageContainerV1() *schema.Resource {
 				Optional: true,
 				ForceNew: false,
 			},
+			"versions_location": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
 			"metadata": &schema.Schema{
 				Type:     schema.TypeMap,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


jtopjian:
@cryptobioz Thank you for working on this. Please let us know if you need any help. :)

cryptobioz:
@jtopjian I'd also like to add the support of `X-History-Location` but I guess we will need this patch: https://github.com/gophercloud/gophercloud/pull/1318

cryptobioz:
@jtopjian Do you think we could merge this?

openstack/resource_openstack_objectstorage_container_v1.go
```diff
@@ -134,6 +165,21 @@ func resourceObjectStorageContainerV1Update(d *schema.ResourceData, meta interfa
...
+		} else {
+			vParams := versioning.List()[0].(map[string]interface{})
+			switch vParams["type"].(string) {
+			case "versions":
+				updateOpts.VersionsLocation = vParams["location"].(string)
+			case "history":
+				updateOpts.HistoryLocation = vParams["location"].(string)
+			}
+		}
+	}
```
jtopjian: `HasChange` should be the first thing checked, just for consistency with other resources. And similar changes to the above:

```go
if d.HasChange("versioning") {
  versioning := d.Get("versioning").(*schema.Set)
  if versioning.Len() == 0 {
    updateOpts.RemoveVersionsLocation = "true"
    updateOpts.RemoveHistoryLocation = "true"
  } else {
    vParams := versioning.List()[0]
    if vRaw, ok := vParams.(map[string]interface{}); ok {
      switch vRaw["type"].(string) {
      case "versions":
        updateOpts.VersionsLocation = vRaw["location"].(string)
      case "history":
        updateOpts.HistoryLocation = vRaw["location"].(string)
      }
    }
  }
}
```

Also, what happens if the user changes their values to empty string (`""`)? Should the `else` statement have a check for empty strings and set `Remove*Location`?

website/docs/r/objectstorage_container_v1.html.markdown
```diff
@@ -50,6 +55,10 @@ The following arguments are supported:
 * `container_write` - (Optional) Sets an ACL that grants write access.
     Changing this updates the access control list write access.
 
+* `versioning` - (Optional) Enable object versioning. Two parameters are
+    required: `type` is the type of versioning and only accepts the values `versions` and `history`,
+    `location` is the container in which the versions will be stored.
```
jtopjian: This is good, but let's change the structure to how other docs are formatted. Take a look at how `fixed_ip` is described here:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/docs/r/networking_port_v2.html.markdown

cryptobioz:
@jtopjian Thanks for the review!

I remember why I didn't update vendoring as you requested, the build is now failing...

jtopjian:
@cryptobioz OK, as luck would have it, there was a breaking change. I'll get that sorted out shortly in a separate PR.

openstack/resource_openstack_objectstorage_container_v1.go
```diff
@@ -87,6 +108,16 @@ func resourceObjectStorageContainerV1Create(d *schema.ResourceData, meta interfa
...
 
+	if versioning, ok := d.Get("versioning").(*schema.Set); ok && versioning.Len() > 0 {
+		vParams := versioning.List()[0].(map[string]interface{})
+		switch vParams["type"].(string) {
+		case "versions":
+			createOpts.VersionsLocation = vParams["location"].(string)
+		case "history":
+			createOpts.HistoryLocation = vParams["location"].(string)
+		}
+	}
```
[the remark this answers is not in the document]
cryptobioz: That's much better!

openstack/resource_openstack_objectstorage_container_v1.go
cryptobioz: Now, an empty value for `type` or `location` will disable versioning.

jtopjian:
@cryptobioz I've merged #489 which will take care of the failing tests. You'll also no longer need to do any versioning in this commit.

Can you rebase off of master?

cryptobioz:
Thanks @jtopjian! I just rebased the PR.

cryptobioz:
@jtopjian Thank you very much for your help!

Issue #456, opened 2018-10-23
Internal: Code Cleanup


jtopjian:
This issue will track some internal code cleanup we plan to do over the next while. All PRs related to code cleanup should link to this issue.

At the moment, we plan to:

1. Replace provider-specific validation code with core Terraform helpers (where applicable).

2. Rename functions to use the "expand/flatten" pattern seen in other providers ([example](https://github.com/terraform-providers/terraform-provider-aws/blob/f0359b9048d0bfa6f5a6268d8d9ce071bbe54477/aws/resource_aws_eks_cluster.go#L251-L288)).

3. Attempt to standardize on log and error message formats.

4. Possibly move non-CRUD functions to a shared resource file. For example, functions in the `openstack_dns_*` resources that are not named `*Create`, `*Read`, `*Update`, or `*Delete` will be moved to a `dns_v2_shared.go` or `shared_dns_v2.go` file.

This list will be updated with any other cleanup items.

These changes _should not_ alter the resource's behavior. If we come across refactoring that might alter the behavior, it will be done in a separate issue. If a change introduced a bug, we fix it and create a new release as soon as possible.

jtopjian:
@ozerovandrei I've updated this issue with some of the methods I used in the first few PRs. These aren't written in stone, though, so let me know if you want to see something changed.

ozerovandrei:
@jtopjian rules looks good for now.

I think we will extend them with some additional cases about performing resource updates in several steps.
#258 is a good example of such complex updates.

I checked current PRs and I want to ask - can we also add simple unit tests for all new helpers in the `compute_flavor_v2.go`, `compute_keypair_v2.go` and `compute_secgroup_v2.go`? I think they will be helpful.

kayrus:
According to [this comment](https://github.com/terraform-providers/terraform-provider-openstack/pull/515/files/9ac77a391917ba7341f8470cde7f24b5f5b7452e#r241261199) we can use `validation.StringInSlice` for limited input entries. Is a new ticket required for this task?

kayrus:
Another [cleanup case](https://github.com/terraform-providers/terraform-provider-openstack/pull/515/files/9ac77a391917ba7341f8470cde7f24b5f5b7452e#r241262712):

remove `d.SetId("")` at the end of `Delete` functions.

jtopjian:
> According to this comment we can use validation.StringInSlice for limited input entries. Is a new ticket required for this task?

[...]

PR #478, opened 2018-11-04
Internal: compute v2 secgroup cleanup

jtopjian:
For #456 

This one can probably use more refactoring, but I didn't want to make too many changes for fear of breaking something. 

For example, `computeV2SecGroupRulesCheckForErrors` _might_ work as a validation function in the schema. As well, `flattenComputeV2SecGroupRules` might have room for changes.

### DIFF
--- a/openstack/compute_secgroup_v2.go
+++ b/openstack/compute_secgroup_v2.go
@@ -51,10 +51,10 @@ func computeSecGroupV2RulesCheckForErrors(d *schema.ResourceData) error {
 }
 
 func expandComputeSecGroupV2CreateRules(d *schema.ResourceData) []secgroups.CreateRuleOpts {
-	rawRules := d.Get("rule").(*schema.Set)
-	createRuleOptsList := make([]secgroups.CreateRuleOpts, rawRules.Len())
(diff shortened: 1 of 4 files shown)


openstack/compute_secgroup_v2.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func expandComputeSecGroupV2CreateRules(d *schema.ResourceData) []secgroups.CreateRuleOpts {
+	rawRules := d.Get("rule").(*schema.Set)
+	createRuleOptsList := make([]secgroups.CreateRuleOpts, rawRules.Len())
```
ozerovandrei: I think that we need to return empty []secgroups.CreateRuleOpts if `rawRules` is a nil pointer prior trying to run `Len()` over it.

jtopjian: Good point. I've gone ahead and just reverted to the original code so we remove any doubt of behavior.

jtopjian:
@ozerovandrei I've added two unit tests. I'm unable to create a test for `flattenComputeSecGroupV2Rules` at this time since it requires a Gophercloud ServiceClient.

Issue #473, opened 2018-11-04
Compute v2: Add extra_specs to openstack_compute_flavor_v2 data source


jtopjian:
`extra_specs` was added after both the `openstack_compute_flavor_v2` resource and data source were initially created. It was added to the resource, but not the data source.

jtopjian:
Implemented in #480

PR #480, opened 2018-11-11
Compute v2: Adding extra_specs to flavor data source

jtopjian:
For #473


jtopjian:
For now, extra specs are only being used in the result of the query, but it should be possible to query by extra specs. This can be a future enhancement.

PR #486, opened 2018-11-12
Network v2: Rename extra_dhcp_opts to extra_dhcp_option

jtopjian:
For #99


jtopjian:
@ozerovandrei Let me know what you think of this change.

Renaming `extra_dhcp_opts` plural to singular is how fields like this _should_ be named. `allowed_address_pairs` also doesn't follow this convention but we should fix that (by creating a new field and deprecating the old).

`fixed_ip`, however, follows this convention.

I also chose to rename `extra_dhcp_opt` to `extra_dhcp_option` as it's more user-friendly. Same with `opt_name` and `opt_value`.

Thoughts?

jtopjian:
Thanks!

PR #488, opened 2018-11-16
ContainerInfraV1: add Cluster datasource

ozerovandrei:
Add OpenStack ContainerInfraV1 Cluster datasource with tests.

For #402

### DIFF
--- a/openstack/data_source_openstack_containerinfra_cluster_v1.go
+++ b/openstack/data_source_openstack_containerinfra_cluster_v1.go
@@ -1,0 +1,155 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clusters"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceContainerInfraCluster() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceContainerInfraClusterRead,
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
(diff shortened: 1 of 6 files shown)


openstack/data_source_openstack_containerinfra_cluster_v1.go
```diff
@@ -0,0 +1,155 @@
...
+	d.Set("master_flavor", c.MasterFlavorID)
+	d.Set("keypair", c.KeyPair)
+	d.Set("master_count", c.MasterCount)
+	d.Set("node_count", c.NodeCount)
+	d.Set("master_addresses", c.MasterAddresses)
+	d.Set("node_addresses", c.NodeAddresses)
+	d.Set("stack_id", c.StackID)
+
+	if err := d.Set("labels", c.Labels); err != nil {
+		log.Printf("[DEBUG] labels: %s", err)
```
jtopjian: What do you think of changing this to:

```go
log.Printf("[DEBUG] Unable to set labels for cluster %s: %s", c.UUID, err)
```

Same for the others below, too.

openstack/resource_openstack_containerinfra_cluster_v1_test.go
```diff
@@ -126,7 +126,9 @@ resource "openstack_containerinfra_clustertemplate_v1" "clustertemplate_1" {
   coe                   = "kubernetes"
   master_flavor         = "%s"
   flavor                = "%s"
-  floating_ip_enabled   = false
+  floating_ip_enabled   = true
+  volume_driver         = "cinder"
+  docker_storage_driver = "devicemapper"
```
jtopjian: Are these pre-created in a devstack environment?

ozerovandrei: No, devstack doesn't create any default cluster templates for now: [magnum/devstack](https://github.com/openstack/magnum/tree/master/devstack).

Well, I just used those values because Vexxhost and Selectel have them set in the pre-created public cluster templates. So I thought maybe this is a good idea.

Issue #118, opened 2017-09-15
Server group soft anti affinity policy not available if not using the correct microversion API


sebastien-prudhomme:
### Terraform Version
.Terraform v0.10.3

### Affected Resource(s)
- openstack_compute_servergroup_v2

### Terraform Configuration Files
```hcl
resource "openstack_compute_servergroup_v2" "test" {
  name = "test"

  policies = [
    "soft-anti-affinity",
  ]
}
```

### Debug Output
```
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4: 2017/09/15 14:55:30 [DEBUG] Openstack Request Headers:
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4: Accept: application/json
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4: Content-Type: application/json
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4: User-Agent: Terraform 0.10.0-dev (go1.8) gophercloud/2.0.0
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4: X-Auth-Token: ***
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4: 2017/09/15 14:55:30 [DEBUG] OpenStack Request Body: {
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4:   "server_group": {
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4:     "name": "test",
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4:     "policies": [
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4:       "soft-anti-affinity"
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4:     ]
2017-09-15T14:55:30.183+0200 [DEBUG] plugin.terraform-provider-openstack_v0.2.1_x4:   }
[...]

PR #490, opened 2018-11-21
ComputeV2: add soft policies

ozerovandrei:
Set "2.15" microversion for servergroup "soft-anti-affinity" and "soft-affinity" policies.
Add tests and documentation notes.

For #118

### DIFF
--- a/website/docs/r/compute_servergroup_v2.html.markdown
+++ b/website/docs/r/compute_servergroup_v2.html.markdown
@@ -30,10 +30,9 @@ The following arguments are supported:
 * `name` - (Required) A unique name for the server group. Changing this creates
     a new server group.
 
-* `policies` - (Required) The set of policies for the server group. Only two
-    two policies are available right now, and both are mutually exclusive. See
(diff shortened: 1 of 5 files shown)


website/docs/r/compute_servergroup_v2.html.markdown
```diff
@@ -45,6 +44,14 @@ The following arguments are supported:
 * `anti-affinity` - All instances/servers launched in this group will be
     hosted on different compute nodes.
 
+* `soft-affinity` - All instances/servers launched in this group will be hosted
```
jtopjian: Maybe we should mention that the OpenStack Compute environment must support microversion 2.15 or above?

ozerovandrei:
@jtopjian yep, all sounds great, will update it shortly.

jtopjian:
> @jtopjian I don't know why I didn't check it prior vendoring. Sure, it works. I removed vendoring commit.

It's no problem at all. We ran into an odd similar govendor issue a while back so I was more wondering if my vendor commit was faulty.

PR #491, opened 2018-11-22
[DNM]Add jobs to integrate terraform-provider-openstack with generic clouds

h00130372:
Add jobs to integrate terraform-provider-openstack with
public clouds: telefonica, orange, opentelekomcloud and huaweicloud.

Related-Bug: theopenlab/openlab#125


h00130372:
We just want test whether terraform-provider-openstack can integrate with the public clouds, if can't how many tests will be fail, I will close this PR, and will test it in another way, thank you.

PR #494, opened 2018-11-27
fix 493 - explicitly set default bool attributes

### LINKED ISSUE #493: [v0.12] Error: rpc error: code = Unavailable desc = transport is closing
This error is also seen in the debug lot output:
```panic: ConfigValueFromHCL2Block used with unknown value```

### Terraform Version
`Terraform v0.12.0-alpha2`

### Affected Resource(s)
All

I think it's an issue with the provider configuration block

```hcl
provider "OpenStack" {
  auth_url = "https://myopenstackurl:5000/v3/"
  user_domain_name = "foo"
  tenant_id = "bar"
  insecure = true
}

resource "openstack_networking_network_v2" "foo" {
  name = "bar"
}
```

### Debug Output
https://gist.github.com/sjwl/baad0e5dac69ba8e89aae65f8c5a01ff

### Expected Behavior
No errors

### Actual Behavior
```
!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

Terraform crashed! This is always indicative of a bug within Terraform.
A crash log has been placed at "crash.log" relative to your current
working directory. It would be immensely helpful if you could please
report the crash with Terraform[1] so that we can fix this.

When reporting bugs, please include your terraform version. That
information is available on the first line of crash.log. You can also
get it by running 'terraform --version' on the command line.

[1]: https://github.com/hashicorp/terraform/issues

!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
I believe the problem is how the provider config default value for `bool` types. i.e. - `insecure`, `swath`, and `use_octavia`.  The default value should either be `true` or `false` but they default to either `nil` or empty ("").

sjwl:
fix #493

### DIFF
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -131,7 +131,7 @@ func Provider() terraform.ResourceProvider {
 			"insecure": &schema.Schema{
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("OS_INSECURE", false),
+				DefaultFunc: schema.EnvDefaultFunc("OS_INSECURE", nil),
 				Description: descriptions["insecure"],
 			},
 
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -131,7 +131,7 @@ func Provider() terraform.ResourceProvider {
 			"insecure": &schema.Schema{
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("OS_INSECURE", nil),
+				DefaultFunc: schema.EnvDefaultFunc("OS_INSECURE", false),
 				Description: descriptions["insecure"],
 			},
 
@@ -165,14 +165,14 @@ func Provider() terraform.ResourceProvider {
 			"swauth": &schema.Schema{
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("OS_SWAUTH", ""),
+				DefaultFunc: schema.EnvDefaultFunc("OS_SWAUTH", false),
 				Description: descriptions["swauth"],
 			},
 
 			"use_octavia": &schema.Schema{
 				Type:        schema.TypeBool,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("OS_USE_OCTAVIA", ""),
+				DefaultFunc: schema.EnvDefaultFunc("OS_USE_OCTAVIA", false),
 				Description: descriptions["use_octavia"],
 			},
 


sjwl:
Understood.  Even without 0.12, the fix still applies to 0.11, as there is no reason to set the default value of a bool to nil.  It ought to be either false or true.

sjwl:
Also, just checked the aws provider, and all of its bool attribute defaults are either `true` or `false`.

jtopjian:
> Also, just checked the aws provider, and all of its bool attribute defaults are either true or false.

Thanks for checking this. I'll ask about this internally and see if this is something we're going to have to change for v0.12 compatibility. It's a small change, but I want to be cautious about changing default values (since they will affect existing environments).

jtopjian:
@sjwl I've opened an issue for this in Terraform core (see the above link) to see if this is something that should be accounted for in core or in this provider.

Thank you again for reporting this. For now, let's wait to hear a decision :)

jtopjian:
@sjwl Looks like this has been fixed upstream :)

https://github.com/hashicorp/terraform/pull/19521

splashx:
@jtopjian do u have the binary for 0.12 somewhere or tips on how to compile the module to start integration testing with 0.12?

jtopjian:
@sjwl I don't have a set of commands that I feel confident sharing yet. 

Currently I've been doing a series of `govendor fetch ...` to update the `github.com/hashicorp/terraform` packages, looking for compile errors, updating those packages, and then repeat until everything compiles. As v0.12 gets closer to being released, we should have an "official" vendor update process.

sjwl:
so I tried `v0.12.0-alpha4` and it spits out this warning:

```
terraform plan

[...]

Warning: provider set empty string as default value for bool use_octavia

Warning: provider set empty string as default value for bool swauth
```

So while they made 0.12 backwards compatible it's announcing to the world that the provider is not acting correct.

jtopjian:
@sjwl Yes, and technically this is correct. We'll get this fixed appropriately either closer to the v0.12 release or very shortly after.

But at least now it's not giving an error :)

jtopjian:
Oh, and I just noticed my prior comment was directed at @sjwl and not @splashx - sorry about that!

jtopjian:
I'm going to close this in favor of #542

sjwl:
why close this PR in favor of #542?

jtopjian:
Upon revisiting, simply because I forgot that this is almost identical to #542. The only change being that `insecure` is kept as `nil` because it is for a `*bool` and not just `bool`.

jtopjian:
@sjwl if you'd like to make the change to `insecure`, I'm more than happy to merge this one in place of #542.

sjwl:
@jtopjian 
The private struct storing `insecure` is indeed a `*bool` type however as you can see from this excerpt the schema type is a `bool` when read, and is switched to `*bool` when it's copied into the private struct.
```
	v, ok := d.GetOkExists("insecure")
	if ok {
		insecure := v.(bool)
		config.Insecure = &insecure
	}
```

I think the schema attribute `insecure` should *never* be `nil`.  If you think about it, should a user want to change the default `insecure` the user would specify `true` or `false`, not a pointer value, right?  Or am I missing something fundamental here?

Nevertheless I've reverted `insecure` default back to `nil` on my branch, pending your further review.

jtopjian:
> I think the schema attribute insecure should never be nil. If you think about it, should a user want to change the default insecure the user would specify true or false, not a pointer value, right? Or am I missing something fundamental here?

Totally valid point that deserves an in-depth explanation. It's more of a historical thing mixed with Go zero-value edge cases.

When it comes to boolean and zero-values, it's hard to pass an explicit value of "false". In order to do that, you must assign the value to a `*bool` so that `nil` becomes the zero-value and everything else is passed.

So whoever added `insecure` was accounting for:

1. A user passing `true`
2. A user passing `false`
3. A user omitting the option altogether

In this specific case, `insecure` could be changed to `bool` and everything would be fine, but there are the odd edge-cases where nil/true/false is critical - perhaps the contributor was recently working on something like that and had the style in their mind.

As for how this is implemented in the code, it's completely safe. This configuration is never assuming the user is going to pass `nil`. `nil` is only ever going to be the default non-user-supplied value, which is a trigger to say "this item was not explicitly defined nor was it ever set by an environment variable". When you look at it that way, there's value to it.

However, nowadays, one could use `d.GetOkExists` to achieve the same thing, but support for that was only added a few months ago.

Given we're not getting a warning about this item in v0.12, I would prefer to leave it as-is. When it comes to schema items, I don't ever want to change them unless absolutely necessary - the situation where a seemingly innocent change breaks someone's config is the one situation I always try to avoid.

Let me know if that helps clarify things or if you have any other questions.

jtopjian:
> Given we're not getting a warning about this item in v0.12, I would prefer to leave it as-is. When it comes to schema items, I don't ever want to change them unless absolutely necessary - the situation where a seemingly innocent change breaks someone's config is the one situation I always try to avoid.

A bit more on this: Changing `insecure` to a default value of `false` and changing the `*bool` to `bool` is 99.9999% safe to do. There are two reasons why I prefer to keep it as-is:

1. It's a personal rule I have to be strict about changing user-facing pieces that don't need to be changed, no matter their historical implementation.

2. It's quite possible there are some private OpenStack providers in the wild that are using this as the base and adding in-house patches to them. In this context, changing this could cause more work for those situations. Now, if you fork the code, the responsibility is on you to account for things like this, but again, it's working at the moment, so let's just not cause that possibility from happening at all.

jtopjian:
We just did a new release today before the holidays. I'm going to go ahead and merge this so people can use it for v0.12 testing.

Thank you for reporting this issue and working through it. I apologize for the clumsiness of opening a new PR.

Please let me know if you have any questions. I'm also going to open a new issue that will be dedicated to v0.12 discussion.

jtopjian:
https://github.com/terraform-providers/terraform-provider-openstack/issues/548 has been opened and I'm also trying out this fancy new "pinning" feature to make it more visible.

PR #496, opened 2018-11-28
Allow a VM to be in a state of 'migrating'

mifrost:
When applying changes, we have regularly received the error `Invalid power_state for instance <ID>: MIGRATING` , as live migrations occur on instances, preventing any plan / apply operations.

This change adds "migrating" to the list of acceptable states for an instance to be in. See also #428

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -353,7 +353,7 @@ func resourceComputeInstanceV2() *schema.Resource {
 				ForceNew: false,
 				Default:  "active",
 				ValidateFunc: validation.StringInSlice([]string{
-					"active", "shutoff", "migrating",
+					"active", "shutoff",
 				}, true),
 				DiffSuppressFunc: suppressPowerStateDiffs,
 			},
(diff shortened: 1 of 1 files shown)


jtopjian:
Thanks for submitting this patch.

Can you explain in more detail why someone would want to make a change when an instance is in the middle of migrating? My initial impression is that this kind of error is actually a good thing.

mifrost:
@jtopjian No problem

In our case it's because terraform will try to refresh the state of all instances before every `plan` action. This means that even if there are no changes to make to the migrating instance, the whole action will still fail. 

Even if we attempt to avoid the problem with a `-target`, we still hit it due to a shared dependency as we've used a `count` parameter.

jtopjian:
@mifrost OK, that makes sense.

So perhaps instead of having `migrating` as a valid input (line 356) how about we just let it be a valid status (such as how "error" is being used). Thoughts?

mifrost:
@jtopjian very good point - I've updated the PR

jtopjian:
@mifrost This looks good to me. Have you been able to validate this change in your environment? It's a little hard for me to reproduce a case where this would be triggered.

mifrost:
@jtopjian yes - I've used this branch to work around the issue we were seeing

Issue #498, opened 2018-11-29
Error creating OpenStack Neutron network: Resource not found


clayg:
Hi there,

I'm having an issue using this terraform provider to create private networks.

### Terraform Version
```
$ terraform --version
Terraform v0.11.10
+ provider.openstack v1.12.0
```
### Affected Resource(s)

- openstack_networking_network_v2

### Terraform Configuration Files
```
$ cat main.tf 
provider "openstack" {
  user_name   = "${var.rax_user}"
  password    = "${var.rax_pass}"
  auth_url    = "https://identity.api.rackspacecloud.com/v2.0/"
  region      = "DFW"
}

resource "openstack_networking_network_v2" "network_1" {
  name           = "network_1"
  admin_state_up = "true"
}
```

### Debug Output
There's a lot of debug output from:

```
OS_DEBUG=1 TF_LOG=1 terraform apply -auto-approve
```

But the relevant bit is:

```
2018-11-28T17:17:24.528-0600 [DEBUG] plugin.terraform-provider-openstack_v1.12.0_x4: 2018/11/28 17:17:24 [DEBUG] OpenStack Request URL: POST https://dfw.networks.api.rackspacecloud.com/v2.0/v2.0/networks
```

I think the wrong url is a well understood problem at this point [1, 2, 3]

### Expected Behavior

Make the POST to the correct URL w/o the double version numbers

### Actual Behavior

The POST is made to wrong URL

### Steps to Reproduce

1. set rax_user & rax_pass in `terraform.tfvars`
2. create simple `main.tf` with `openstack_networking_network_v2` using example above
3. `terraform apply`

### Important Factoids

I think Rackspace + openstack provider has been broken this way since a couple years ago - but I'm not sure it wouldn't work now if the code only append v2.0 when it's missing?  I'm not having any issues with the `openstack_compute_keypair_v2` or `openstack_compute_instance_v2` resources.

It looked like maybe some progress was made earlier in the year for v3 auth [4] - but Rackspace auth seems more compatible with `v2.0` and does not seem to respond to `/v3/`

### References

1. hasicorp/terraform#1560
2. https://storyboard.openstack.org/#!/story/2004310
3. https://github.com/terraform-providers/terraform-provider-openstack/blob/410a5191973e939afb8ea6bbf3a0f878970719f5/vendor/github.com/gophercloud/gophercloud/openstack/client.go#L324
4. #239

clayg:
[...]

PR #499, opened 2018-11-30
[WIP] Networking v2: Fix cases where the Networking service was published with a versioned endpoint

jtopjian:
For #498 

@clayg I reconfigured my dev environment to have Keystone publish Neutron with a `/v2.0` suffix and this fix works for me. Let me know if you have any problems with it.

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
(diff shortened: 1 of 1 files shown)


clayg:
@jtopjian this let me create a private network on rackspace :+1:

jtopjian:
@clayg Thank you for verifying! Two questions:

1. #501 seems like it would be a better, more flexible solution since it provides support for all endpoints as well as would cover cases when an endpoint may have been erroneously published with `/v2/` instead of `/v2.0/`. Do you feel having to hard-code the Neutron URL in the options would be an annoying or prohibitive solution?

2. In your testing, do you think this one fix enables full network support with Neutron? This is something I can research on my own, but thought I would see if you already knew the answer.

Thanks!

clayg:
Ok, I glanced at #501 - not a "proper" review mind you - I wouldn't be qualified.

I'm still kinda having some trouble with the networking on Rackspace sadly.  `terraform apply` works 100% - but afterwards `terraform plan` still shows the machines as having no networks - which seems to match what I see in openstacksdk too so there might be some weird difference on Rax.  But I haven't given up - just debugging.

I'm traveling next week.  I'll try to figure out what's going on at some point - but even as it stands I feel like it's working 100% better than when it 404's to the wrong URL so THANK YOU!

jtopjian:
No problem at all! If we're able to add fixes that aren't Rackspace-specific (meaning: not hard-coding anything Rackspace), I'm more than happy to continue adding workarounds.

With regard to not seeing networks on the instances, that might be unavoidable. One thing to try is:

```shell
$ TF_LOG=DEBUG OS_DEBUG=1 terraform refresh
```

This will print a _ton_ of logs, but it might be possible to see why the networks aren't being detected.

One downside is that we already do [a lot](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/compute_instance_v2_networking.go) of work to currently detect networks and I'm hesitant to touch this logic. The reason this is so convoluted is mostly historical: Gophercloud didn't support certain API calls at the time to make this easier.

jtopjian:
I'm going to close this PR in favor of #501

Issue #500, opened 2018-11-30
Support Overriding Endpoints


jtopjian:
The discussion in #498 made me wonder how useful it would be to support the ability to override endpoints rather than use the endpoint returned in the keystone/service catalog.

Use of this feature would be for unique and edge cases, but it would be a very useful feature when needed.

The downside to this is that the user must know what they're doing in order to use it correctly. The user must specify the full endpoint URL, including any extra information required (such as hard-coding a tenant/project ID).

PR #501, opened 2018-11-30
[WIP] Provider: Support Overriding Endpoints

jtopjian:
This commit enables users to override service endpoints rather than use
the endpoint URL returned from the service catalog.

For #500 

I'm marking this as `WIP` until there's more discussion around whether this feature should be added or not.


jtopjian:
@clayg here's an alternative solution for #499. There's documentation included in the diff on how to use this.

openstack/config.go
```diff
@@ -208,95 +209,243 @@ func (c *Config) determineRegion(region string) string {
...
+
+	if err != nil {
+		return client, err
+	}
+
+	// Check if an endpoint override was specified for the container-infra service.
+	if endpoint := c.getEndpointOverride("container-infra"); endpoint != "" {
+		client.Endpoint = endpoint
+		client.ResourceBase = ""
+	}
```
clayg: I'm inexperienced with golang, to me all this error checking and getEndpointOverrride looks too similar to not smell like boilerplate? :man_shrugging: 

You know what I would think would be handy, if in all of these places you logged what endpoint ended up getting used for each service (like even if the config didn't override it) - so when you're trying to debug with TF_LOG=1 and OS_DEBUG=1 you can sorta piece together what your override should be..

But I don't really know what that would look like, so you can ignore me.

jtopjian:
> So my first thought was YAGNI. But then I imagined a bug report for a similar issue down the road... I mean aside from this thing with Rackspace - have you heard of this before!?

From having fielded questions for the past 3 years with this provider, nothing surprises me any more :)

> So, yeah.. maybe? What do you think?

I'm leaning on adding support for this. It's a feature that isn't intrusive for those that don't need it and very helpful in the edge cases that would need it.

> I'm inexperienced with golang, to me all this error checking and getEndpointOverrride looks too similar to not smell like boilerplate?

Indeed - I felt the same way. I'll look at refactoring this.

> You know what I would think would be handy, if in all of these places you logged what endpoint ended up getting used for each service

Good call.

> But I don't really know what that would look like, so you can ignore me.

Totally valid points that I won't be ignoring :smile:

jtopjian:
I've refactored the code and also included some logging.

@ozerovandrei thoughts on this PR?

jtopjian:
recheck

jtopjian:
recheck

Issue #430, opened 2018-09-22
Enable Transparent VLAN mode


dustnic:
This request is to support the network **transparent-vlan** option for the resource `openstack_networking_network_v2 `.

The transparent VLAN is a quite important feature especially in NFV use-cases.

https://specs.openstack.org/openstack/neutron-specs/specs/kilo/nfv-vlan-trunks.html
https://docs.openstack.org/python-openstackclient/pike/cli/command-objects/network.html

ozerovandrei:
Support for the `vlan-transparent` extension should be added into Gophercloud first.
I've created an issue: [Gophercloud#1304](
https://github.com/gophercloud/gophercloud/issues/1304).

ozerovandrei:
@dustnic Hello.
Support for the `transparent_vlan` argument/attribute was implemented and it will be available in the `v1.14.0` release.

PR #513, opened 2018-12-08
NetworkingV2: add transparent_vlan network resource attribute 

ozerovandrei:
Add the "transparent_vlan" attribute into networking_network_v2 resource.

For #430

### DIFF
--- a/openstack/resource_openstack_networking_network_v2.go
+++ b/openstack/resource_openstack_networking_network_v2.go
@@ -171,7 +171,7 @@ func resourceNetworkingNetworkV2Create(d *schema.ResourceData, meta interface{})
 	// Add networking segments if specified.
 	if len(segments) > 0 {
 		finalCreateOpts = provider.CreateOptsExt{
-			CreateOptsBuilder: createOpts,
+			CreateOptsBuilder: finalCreateOpts,
 			Segments:          segments,
 		}
 	}
@@ -284,15 +284,6 @@ func resourceNetworkingNetworkV2Update(d *schema.ResourceData, meta interface{})
 	if d.HasChange("description") {
 		description := d.Get("description").(string)
 		updateOpts.Description = &description
-	}
-	if d.HasChange("tags") {
-		tags := networkV2AttributesTags(d)
-		tagOpts := attributestags.ReplaceAllOpts{Tags: tags}
-		tags, err := attributestags.ReplaceAll(networkingClient, "networks", d.Id(), tagOpts).Extract()
-		if err != nil {
-			return fmt.Errorf("Error updating Tags on Network: %s", err)
-		}
-		log.Printf("[DEBUG] Updated Tags = %+v on Network %+v", tags, d.Id())
 	}
 	if d.HasChange("admin_state_up") {
 		asuRaw := d.Get("admin_state_up").(string)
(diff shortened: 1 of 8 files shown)


ozerovandrei:
@jtopjian hmm, I can't understand the reason of the openlab/check failure. Can you help please?

jtopjian:
@ozerovandrei It's happening to other PRs, even in Gophercloud. I think there's an issue right now :(

jtopjian:
@ozerovandrei @kayrus I've opened an issue at OpenLab about the merge errors: https://github.com/theopenlab/openlab/issues/143

mrhillsman:
recheck

openstack/resource_openstack_networking_network_v2.go
```diff
@@ -298,23 +314,41 @@ func resourceNetworkingNetworkV2Update(d *schema.ResourceData, meta interface{})
 			updateOpts.Shared = &shared
 		}
 	}
+
+	// Change tags if needed.
+	if d.HasChange("tags") {
```
jtopjian: Moving this is fine, but the original one needs deleted.

vendor/vendor.json
```diff
@@ -708,6 +708,12 @@
 			"revision": "5a7d5fbc4cc3b39700ad21b65c5eb51ef0d19c84",
 			"revisionTime": "2018-12-12T03:37:06Z"
 		},
+		{
+			"checksumSHA1": "TH4LSJLhZlq0fW36UVWltFlM3KA=",
+			"path": "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/vlantransparent",
+			"revision": "26de66c23d78a75fc37e5dad5b6f16ffbfe9ee31",
```
jtopjian: The revision here doesn't match the others - this will probably need revendored.

ozerovandrei:
@jtopjian sorry for the delay.

Unfortunately I can't test it because I don't have needed environment and devstack is installed with the OVS plugin that needs the following change: [#576687](https://review.openstack.org/#/c/576687/) to support this feature.

jtopjian:
@ozerovandrei My test environment is using linuxbridge, so I was able to enable this extension. It looks like `vlan_transparent` is not able to be updated:

```
openstack_networking_network_v2.network_1: Error updating OpenStack Neutron Network: Bad request with: [PUT http://neutron:9696/v2.0/networks/deaf76d9-9453-4c0a-b3b5-0a45932c6d24], error message: {"NeutronError": {"message": "Cannot update read-only attribute vlan_transparent", "type": "HTTPBadRequest", "detail": ""}}
```

Maybe some drivers allow this to be updated, but I wonder if we should change `vlan_transparent` to `ForceNew` for now as a precaution? If we get confirmation that some drivers allow this to be updated, we can remove `ForceNew`. I think this would be a less disruptive change than doing it the opposite way.

I think we can leave the Gophercloud implementation for now, just in case some drivers allow this to be updated.

ozerovandrei:
@jtopjian sounds great, I will change it.

openstack/resource_openstack_networking_network_v2.go
```diff
@@ -298,23 +305,41 @@ func resourceNetworkingNetworkV2Update(d *schema.ResourceData, meta interface{})
...
 			External:          &isExternal,
 		}
-		log.Printf("[DEBUG] Updating Network %s with options: %+v", d.Id(), externalUpdateOpts)
-		_, err = networks.Update(networkingClient, d.Id(), externalUpdateOpts).Extract()
-	} else {
-		log.Printf("[DEBUG] Updating Network %s with options: %+v", d.Id(), updateOpts)
-		_, err = networks.Update(networkingClient, d.Id(), updateOpts).Extract()
+	}
+	isVLANTransparent := false
+	if d.HasChange("transparent_vlan") {
```
jtopjian: Let's remove this for now. I thought about suggesting to comment it out, but since updating might never be supported, that would only leave to unused code / comments in the file.

ozerovandrei: I deleted update test but forgot about removing this code.

PR #509, opened 2018-12-05
Reflect new gophercloud structures

kayrus:
for #508

Issue #502, opened 2018-11-30
Support manila resources (shared file system)


kayrus:
It would be nice to support Manila API in Terraform

### Terraform Version
0.11.10

### Affected Resource(s)
* openstack_sharedfilesystem_securityservice_v2
* openstack_sharedfilesystem_sharenetwork_v2
* openstack_sharedfilesystem_share_v2
* openstack_sharedfilesystem_share_access_v2

### Terraform Configuration Files
N/A

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
I can create shared file systems via terraform.

### Actual Behavior
I can't create shared file systems via terraform.

### Steps to Reproduce
1. Try to create shared file systems using Terraform. 
2. Be disappointed

### Important Factoids
There are gophercloud dependencies like: https://github.com/gophercloud/gophercloud/issues/1334

### References
* WIP is happening here: https://github.com/kayrus/terraform-provider-openstack/commit/5ce41a64fc5f05284c9b4d288173eaa286d0f334

jtopjian:
> WIP is happening here: kayrus@5ce41a6

@kayrus Please submit one resource per PR at a time. Otherwise the PR will be too large to easily review. Thanks!

jtopjian:
> I'll put first two into one PR, because tests will require both resources:

That should be fine, as long as the PR isn't massive. 

If a resource depends on another resource, then it's common to still have individual tests (and separate files) for each. As an example, `openstack_networking_network_v2` has a few tests which only test the single `openstack_networking_network_v2` resource, however, every other networking resource requires a network, so they also have networks in their test fixtures.

On the topic of tests, OpenLab has Manila enabled for Gophercloud, but I'm not sure about Terraform. If it is not enabled, we will want to disable the tests from being run by using an environment variable flag:

* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider_test.go#L16
* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider_test.go#L108-L114
* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_db_database_v1_test.go#L20

At a very minimum, a `_basic` acceptance test is required for each new resource.

> What is the documentation procedure? Should it be inside the same PR? Can documentation be separated?

Documentation is required for new resources and needs to be in the same PR so we can ensure it isn't forgotten about.

[...]

PR #515, opened 2018-12-10
Manila: Shared file system securityservice and sharenetwork resources

kayrus:
for #502
depends on #509


kayrus:
Object storage failure is not related to this PR. Does something happen in openlab?

```
--- FAIL: TestAccObjectStorageV1Object_objectManifest (1.14s)
    testing.go:518: Step 0 error: Check failed: Check 3/3 error: openstack_objectstorage_object_v1.myfile: Attribute 'etag' expected "\"a84ffba5c71c24627a369146227d701a\"", got "\"207c91804d4cbf698be032d2dd7ff735\""
=== RUN   TestAccOpenStackObjectStorageTempurlV1_basic
```

jtopjian:
> Object storage failure is not related to this PR. Does something happen in openlab?

It looks like tests are passing now. It might have been a transient failure? If so, let that be a lesson about trusting checksums / etags :)

kayrus:
@jtopjian Test are fine. Rebasing to master with a proper commits order and waiting for review.

openstack/resource_openstack_sharedfilesystem_securityservice_v2.go
```diff
@@ -0,0 +1,264 @@
...
+	d.Set("dns_ip", securityservice.DNSIP)
+	d.Set("user", securityservice.User)
+	d.Set("server", securityservice.Server)
+	d.Set("region", GetRegion(d, config))
+
+	if ouErr == nil {
+		d.Set("ou", securityservice.OU)
+	} else {
+		d.Set("ou", "")
+	}
```
jtopjian: Since "ou" is not set to "Computed" in the schema, then it will only have a value if the user specified one in their configuration. Because of this, and similar for all other fields that are only "Optional" and not "Computed", `d.Set` is only used here during importing resources. So, it's safe to just do:

```
d.Set("ou", securityservice.OU)
```

openstack/resource_openstack_sharedfilesystem_securityservice_v2.go
```diff
@@ -0,0 +1,264 @@
...
+		return fmt.Errorf("Error creating OpenStack sharedfilesystem client: %s", err)
+	}
+
+	log.Printf("[DEBUG] Attempting to delete securityservice %s", d.Id())
+	err = securityservices.Delete(sfsClient, d.Id()).ExtractErr()
+	if err != nil {
+		return fmt.Errorf("Error deleting securityservice: %s", err)
+	}
+
+	d.SetId("")
```
jtopjian: nit: this is no longer needed as Terraform automatically does this. If this is being used in any other resource, it's slowly being phased out during an ongoing code cleanup.

openstack/resource_openstack_sharedfilesystem_securityservice_v2_test.go
```diff
@@ -0,0 +1,160 @@
...
+resource "openstack_sharedfilesystem_securityservice_v2" "securityservice_1" {
+  name        = "security_through_obscurity"
+  description = ""
+  type        = "kerberos"
+  server      = "192.168.199.11"
+  dns_ip      = "192.168.199.11"
+  domain      = ""
+  ou          = ""
+  user        = ""
+  password    = ""
```
jtopjian: Instead of setting the fields to `""`, let's just omit them. That's what a user would most likely do, anyway.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,282 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"project_id": &schema.Schema{
```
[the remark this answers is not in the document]
kayrus: Interesting... Manila security service create logic gets this field automatically from the context, but share network create logic can get it from the CreateOpts. Don't want to dive into a new rabbit hole, just removed `Optional: true` and `ForceNew: true`.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,282 @@
...
+				Required: true,
+			},
+
+			"neutron_subnet_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"security_service_ids": &schema.Schema{
+				Type:     schema.TypeSet,
```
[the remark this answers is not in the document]
kayrus: According to random failed tests yesterday and [this commit](https://github.com/terraform-providers/terraform-provider-openstack/pull/515/commits/bc8d1a7da1dadf6a15fb5b1f937ec503254b073c) - random order. Bu if I set it to `TypeList`, I won't be able to use `Difference` method mentioned [here](https://github.com/terraform-providers/terraform-provider-openstack/pull/515/files/9ac77a391917ba7341f8470cde7f24b5f5b7452e#r241263801). Let's keep it as is.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,282 @@
...
+	sharenetwork, err := sharenetworks.Create(sfsClient, createOpts).Extract()
+
+	if err != nil {
+		return fmt.Errorf("Error creating sharenetwork: %s", err)
+	}
+
+	d.SetId(sharenetwork.ID)
+
+	securityServiceIDs := resourceSharedfilesystemSharenetworkSecurityServicesV2(d.Get("security_service_ids").(*schema.Set))
+	if len(securityServiceIDs) > 0 {
```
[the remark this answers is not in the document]
kayrus: True.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,282 @@
...
+
+	if updateOpts != (sharenetworks.UpdateOpts{}) {
+		log.Printf("[DEBUG] Updating sharenetwork %s with options: %#v", d.Id(), updateOpts)
+		_, err = sharenetworks.Update(sfsClient, d.Id(), updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Unable to update sharenetwork %s: %s", d.Id(), err)
+		}
+	}
+
+	if d.HasChange("security_service_ids") {
```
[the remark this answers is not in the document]
kayrus: Ah, was not aware about this method, however it could be used only with the `schema.Set`. It makes sense to ignore [previous comment](https://github.com/terraform-providers/terraform-provider-openstack/pull/515/files/9ac77a391917ba7341f8470cde7f24b5f5b7452e#r241262924)

vendor/vendor.json
```diff
@@ -780,6 +780,24 @@
 			"revision": "26de66c23d78a75fc37e5dad5b6f16ffbfe9ee31",
 			"revisionTime": "2018-12-08T02:58:21Z"
 		},
+		{
+			"checksumSHA1": "cGyXqdYGrNLd3zAbwCmxbeVtPMk=",
+			"path": "github.com/gophercloud/gophercloud/openstack/sharedfilesystems/apiversions",
+			"revision": "5a7d5fbc4cc3b39700ad21b65c5eb51ef0d19c84",
```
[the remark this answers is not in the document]
kayrus: Updated

openstack/sfs_v2_shared.go
```diff
@@ -0,0 +1,54 @@
+package openstack
```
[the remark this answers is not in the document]
kayrus: See my [comment ](https://github.com/terraform-providers/terraform-provider-openstack/pull/515#discussion_r241312286)

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go
```diff
@@ -0,0 +1,518 @@
...
+		return fmt.Errorf("Underlying neutron network should differ")
+	}
+}
+
+const testAccSFSV2SharenetworkConfig_basic = `
+resource "openstack_networking_network_v2" "network_1" {
+  name = "network_1"
+  admin_state_up = "true"
+  lifecycle {
+    ignore_changes = ["tags"]
```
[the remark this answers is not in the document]
kayrus: It is for my local environment, every additional service like lbaas or manila atimatically adds tags on the interface.

kayrus:
Added exception in import tests:

```go
ImportStateVerifyIgnore: []string{
  "password",
},
```

openstack/provider_test.go
```diff
@@ -145,6 +146,15 @@ func testAccPreCheckContainerInfra(t *testing.T) {
 	}
 }
 
+func testAccPreCheckSFS(t *testing.T) {
+	testAccPreCheckRequiredEnvVars(t)
+
+	// TODO: enable when ready in OpenLab
+	if false && OS_SFS_ENVIRONMENT == "" {
```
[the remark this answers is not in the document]
jtopjian: OpenLab doesn't have `OS_SFS_ENVIRONMENT` set, though. So the `false` part is moot.

openstack/resource_openstack_sharedfilesystem_securityservice_v2.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	ou := d.Get("ou").(string)
+	if ou != "" {
+		if ouErr == nil {
+			createOpts.OU = ou
+		} else {
+			return ouErr
+		}
+	}
```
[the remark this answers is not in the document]
jtopjian: That works, too.

openstack/resource_openstack_sharedfilesystem_securityservice_v2.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func resourceSharedfilesystemSecurityserviceV2Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	sfsClient, err := config.sharedfilesystemV2Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack sharedfilesystem client: %s", err)
+	}
+
+	mvSet, ouErr := setManilaMicroversion(sfsClient)
```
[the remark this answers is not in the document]
jtopjian: > We need explicitly set the microversion to get the OU value

Right - and that is explicitly done below with `d.GetOkExists("ou")`. :)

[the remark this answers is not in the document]
jtopjian: >  I don't want to call securityservices.Get twice with the fallback to OU suitable microversion.

I forgot to comment on this part.

I totally understand the nuisance of this, but I have a feeling that, in the long-term, this will be unavoidable. Again, this is probably not the case for these resources, but I have a feeling this is going to be unavoidable for other resources when looking at microversion implementations across all projects (at least those that use microversions).

One example of this is how Nova returns flavor information of an instance in newer microversions: instead of returning information about the flavor, it returns specs of the flavor. I understand and agree with why it's now doing this, but it might cause parsing issues for applications which haven't changed to that structure yet.

So what can be done? Update the application to use the new structure and exclude users of environments which don't support that microversion? Try to support both formats? Or conservatively pick and choose the minimal amount of microversion information needed? I'm going with the last option for now.

And in terms of reducing the amount of requests, that's a very admirable goal to work toward. As long as it's not abused, making more than one call is fine. If you look at the output of the openstack cli tools, they're sometimes doing the same thing.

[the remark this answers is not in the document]
kayrus: Removed function

kayrus:
Changes were applied.

openstack/provider_test.go
```diff
@@ -145,6 +146,16 @@ func testAccPreCheckContainerInfra(t *testing.T) {
 	}
 }
 
+func testAccPreCheckSFS(t *testing.T) {
+	testAccPreCheckRequiredEnvVars(t)
+
+	/* TODO: enable when ready in OpenLab
+	if OS_SFS_ENVIRONMENT == "" {
```
jtopjian: @kayrus It looks like Manila is enabled in OpenLab and it did not cause the total test time to increase by much - that was my main concern. However, when we get to adding actual shares (or whichever resource launches an instance), let's revisit the total test time. 

In the meantime, let's leave this as-is for now.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return CheckDeleted(d, err, "sharenetwork")
+	}
+
+	log.Printf("[DEBUG] Retrieved sharenetwork %s: %#v", d.Id(), sharenetwork)
+
+	securityServiceListOpts := securityservices.ListOpts{ShareNetworkID: d.Id()}
+	securityServicePages, err := securityservices.List(sfsClient, securityServiceListOpts).AllPages()
+	if err != nil {
+		return err
```
jtopjian: nit:

```go
return fmt.Errorf("Unable to list security services for sharenetwork %s: %s", d.Id(), err)
```

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	securityServiceList, err := securityservices.ExtractSecurityServices(securityServicePages)
+	if err != nil {
+		return err
```
jtopjian: nit:

```go
return fmt.Errorf("Unable to extract security services for sharenetwork %s: %s", d.Id(), err)
```

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	log.Printf("[DEBUG] Retrieved security services for sharenetwork %s: %#v", d.Id(), securityServiceList)
+
+	if len(securityServiceList) > 0 {
+		d.Set("security_service_ids", resourceSharedFilesystemShareNetworkSecurityServices2IDsV2(&securityServiceList))
+	} else {
+		d.Set("security_service_ids", []string{})
+	}
```
jtopjian: nit: the `else` statement and body can be removed. The `security_service_ids` attribute will automatically be set to a zero-value like this.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
```diff
@@ -0,0 +1,266 @@
...
+
+	d.Set("name", sharenetwork.Name)
+	d.Set("description", sharenetwork.Description)
+	d.Set("neutron_net_id", sharenetwork.NeutronNetID)
+	d.Set("neutron_subnet_id", sharenetwork.NeutronSubnetID)
+	d.Set("project_id", sharenetwork.ProjectID)
```
jtopjian: nit: move this down with the other Computed attributes.

website/docs/r/sharedfilesystem_sharenetwork_v2.html.markdown
```diff
@@ -0,0 +1,118 @@
...
+* `project_id` - The owner of the Share Network.
+* `name` - See Argument Reference above.
+* `description` - See Argument Reference above.
+* `neutron_net_id` - See Argument Reference above.
+* `neutron_subnet_id` - See Argument Reference above.
+* `security_service_ids` - See Argument Reference above.
+* `network_type` - The share network type. Can either be VLAN, VXLAN, GRE, or flat.
+* `segmentation_id` - The share network segmentation ID.
+* `cidr` - The share network CIDR.
+* `ip_version` - The IP version of the share network. Can either be 4 or 6.
```
jtopjian: We need to add the following to the bottom of this file:

```
## Import

This resource can be imported by specifying the UUID of the share network:

$ terraform import openstack_sharedfilesystem_sharenetwork_v2.sharenetwork_1 <uuid>
```

website/docs/r/sharedfilesystem_securityservice_v2.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+* `project_id` - The owner of the Security Service.
+* `name` - See Argument Reference above.
+* `description` - See Argument Reference above.
+* `type` - See Argument Reference above.
+* `dns_ip` - See Argument Reference above.
+* `ou` - See Argument Reference above.
+* `user` - See Argument Reference above.
+* `password` - See Argument Reference above.
+* `domain` - See Argument Reference above.
+* `server` - See Argument Reference above.
```
jtopjian: We need to add the following to the bottom of this file:

```
## Import

This resource can be imported by specifying the UUID of the security service:

$ terraform import openstack_sharedfilesystem_securityservice_v2.securityservice_1 <uuid>
```

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go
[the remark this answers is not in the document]
jtopjian: I forgot to reply to this: this makes sense to me now - thank you!

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
kayrus: Not sure. What if initially there was a security service and then remote data is changed without involving terraform? How will terraform set 1 -> 0 and identify that the data should be set back to the original status?

website/docs/r/sharedfilesystem_sharenetwork_v2.html.markdown
kayrus: UUID isn't used anywhere, will use ID

website/docs/r/sharedfilesystem_securityservice_v2.html.markdown
kayrus: UUID isn't used anywhere, will use ID

website/docs/r/sharedfilesystem_sharenetwork_v2.html.markdown
jtopjian: Isn't the ID in UUID format? But in any case - you're right, it should be whatever the "ID" field is.

openstack/resource_openstack_sharedfilesystem_sharenetwork_v2.go
jtopjian: That's a really good question. These resources should be designed so users are inviting problems if they decide to modify the resource outside of Terraform. It's just way too complicated to try to account for issues like this.

But at the same time, this is actually useful: imagine a random team member modifies the `security_service_ids` out-of-band of Terraform and doesn't tell anyone. Another team member runs `terraform plan` and sees a change happen - they'll know something was changed.

The only exception I can think of is if a user deletes a resource - that's what the `CheckDeleted` helper function is for.

kayrus: So what do you suggest at the end?

jtopjian: just doing

```go
d.Set("security_service_ids", resourceSharedFilesystemShareNetworkSecurityServices2IDsV2(&securityServiceList))
```

without the if/else should be fine.

kayrus: obviously... :) fixed.

kayrus:
Silly bug :)

```
testing.go:518: Step 3 error: Check failed: Check 5/5 error: API and Terraform security service IDs don't correspond: []string(nil) != []string{}
```

Fixed:

```
diff --git a/openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go b/openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go
index f288259..295419c 100644
--- a/openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go
+++ b/openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go
@@ -207,7 +207,7 @@ func testAccCheckSFSV2ShareNetworkSecSvcExists(n string) resource.TestCheckFunc
 
                api_security_service_ids := resourceSharedFilesystemShareNetworkV2SecSvcToArray(&securityServiceList)
 
-               tf_security_service_ids := []string{}
+               var tf_security_service_ids []string
                for k, v := range rs.Primary.Attributes {
                        if strings.HasPrefix(k, "security_service_ids.#") {
                                continue
```

kayrus:
@jtopjian , no, thank _you_ :)

Issue #507, opened 2018-12-05
LBaaS: allow the default_pool_id update in the openstack_lb_listener_v2


kayrus:
It would be nice to support  the `default_pool_id` update without recreating the `openstack_lb_listener_v2`. [lb_listener_v2 documentation](https://www.terraform.io/docs/providers/openstack/r/lb_listener_v2.html#default_pool_id)

### Terraform Version
0.11.10

### Affected Resource(s)
* openstack_lb_listener_v2

### Terraform Configuration Files
N/A

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
I can update `default_pool_id` without recreating the `openstack_lb_listener_v2`

### Actual Behavior
I can't update `default_pool_id` without recreating the `openstack_lb_listener_v2`

### Steps to Reproduce
1. Try to update `default_pool_id` in `openstack_lb_listener_v2` using Terraform.
2. Be disappointed, because terraform recreates `openstack_lb_listener_v2` and all the dependent resources.

### Important Factoids
There are gophercloud dependencies:
* Issue: https://github.com/gophercloud/gophercloud/issues/1333
* PR: https://github.com/gophercloud/gophercloud/pull/1313

kayrus:
Merged in #516

kayrus:
Requires an update. https://github.com/gophercloud/gophercloud/pull/1356

PR #516, opened 2018-12-10
LBaaS: Allow LB listener default pool update without the listener re-creation

kayrus:
for #507
depends on #509

### DIFF
--- a/openstack/resource_openstack_lb_listener_v2.go
+++ b/openstack/resource_openstack_lb_listener_v2.go
@@ -75,7 +75,6 @@ func resourceListenerV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
-				ForceNew: true,
 			},
 
 			"description": &schema.Schema{
@@ -223,6 +222,10 @@ func resourceListenerV2Update(d *schema.ResourceData, meta interface{}) error {
 	if d.HasChange("connection_limit") {
 		connLimit := d.Get("connection_limit").(int)
 		updateOpts.ConnLimit = &connLimit
+	}
+	if d.HasChange("default_pool_id") {
+		defaultPoolID := d.Get("default_pool_id").(string)
+		updateOpts.DefaultPoolID = &defaultPoolID
 	}
 	if d.HasChange("default_tls_container_ref") {
 		updateOpts.DefaultTlsContainerRef = d.Get("default_tls_container_ref").(string)
--- a/openstack/resource_openstack_lb_listener_v2_test.go
+++ b/openstack/resource_openstack_lb_listener_v2_test.go
@@ -109,10 +109,18 @@ resource "openstack_lb_loadbalancer_v2" "loadbalancer_1" {
   vip_subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
 }
 
+resource "openstack_lb_pool_v2" "pool_1" {
+  name            = "pool_1"
+  protocol        = "HTTP"
+  lb_method       = "ROUND_ROBIN"
+  loadbalancer_id = "${openstack_lb_loadbalancer_v2.loadbalancer_1.id}"
+}
+
 resource "openstack_lb_listener_v2" "listener_1" {
   name = "listener_1"
   protocol = "HTTP"
   protocol_port = 8080
+  default_pool_id = "${openstack_lb_pool_v2.pool_1.id}"
   loadbalancer_id = "${openstack_lb_loadbalancer_v2.loadbalancer_1.id}"
 
 	timeouts {
@@ -147,6 +155,7 @@ resource "openstack_lb_listener_v2" "listener_1" {
   protocol_port = 8080
   connection_limit = 100
   admin_state_up = "true"
+  default_pool_id = ""
   loadbalancer_id = "${openstack_lb_loadbalancer_v2.loadbalancer_1.id}"
 
 	timeouts {
--- a/openstack/resource_openstack_lb_loadbalancer_v2.go
+++ b/openstack/resource_openstack_lb_loadbalancer_v2.go
@@ -215,26 +215,28 @@ func resourceLoadBalancerV2Update(d *schema.ResourceData, meta interface{}) erro
 		updateOpts.AdminStateUp = &asu
 	}
 
-	// Wait for LoadBalancer to become active before continuing
-	timeout := d.Timeout(schema.TimeoutUpdate)
-	err = waitForLBV2LoadBalancer(lbClient, d.Id(), "ACTIVE", nil, timeout)
-	if err != nil {
-		return err
-	}
-
-	log.Printf("[DEBUG] Updating loadbalancer %s with options: %#v", d.Id(), updateOpts)
-	err = resource.Retry(timeout, func() *resource.RetryError {
-		_, err = loadbalancers.Update(lbClient, d.Id(), updateOpts).Extract()
+	if updateOpts != (loadbalancers.UpdateOpts{}) {
+		// Wait for LoadBalancer to become active before continuing
+		timeout := d.Timeout(schema.TimeoutUpdate)
+		err = waitForLBV2LoadBalancer(lbClient, d.Id(), "ACTIVE", nil, timeout)
 		if err != nil {
-			return checkForRetryableError(err)
+			return err
 		}
-		return nil
-	})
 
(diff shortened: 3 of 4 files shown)


kayrus:
PR is ready for review.

jtopjian:
recheck lbaas

jtopjian:
I don't think I've run this check for a while - curious to see the outcome.

kayrus:
recheck lbaas

kayrus:
recheck lbaas

kayrus:
@jtopjian, FYI: looks like amphora loadbalancer doesn't support security groups update.

jtopjian:
> FYI: looks like amphora loadbalancer doesn't support security groups update.

When you say "amphora", do you mean Octavia specifically? How did you come across this?

The reason I ask is because some features have been confirmed to work by upstream devs at Vexxhost and Rackspace. It's possible the default configuration being done in OpenLab needs extra configuration? 

For Octavia load balancing tests, I run all tests with a Vexxhost account I have. This works great, but doesn't allow testing for Neutron LBaaS. So maybe that's where the discrepancy is?

kayrus:
> When you say "amphora", do you mean Octavia specifically? How did you come across this?
update tests work fine with octavia, but fail with amphora. LB for some reason in "ERROR" state. See this log:

```
2018-12-11 15:41:52.878972 | ubuntu-xenial | 2018/12/11 15:41:52 [DEBUG] Waiting for loadbalancer e999fd88-a0d2-4b82-b1bb-0fa64a187111 to become ACTIVE.
2018-12-11 15:41:52.879032 | ubuntu-xenial | 2018/12/11 15:41:52 [DEBUG] Waiting for state to become: [ACTIVE]
2018-12-11 15:41:53.057018 | ubuntu-xenial | 2018/12/11 15:41:53 [DEBUG] Updating loadbalancer e999fd88-a0d2-4b82-b1bb-0fa64a187111 with options: loadbalancers.UpdateOpts{Name:(*string)(nil), Description:(*string)(nil), AdminStateUp:(*bool)(nil)}
2018-12-11 15:41:53.057143 | ubuntu-xenial | 2018/12/11 15:41:53 [DEBUG] Waiting for state to become: [success]
2018-12-11 15:41:53.159114 | ubuntu-xenial | 2018/12/11 15:41:53 [DEBUG] Waiting for loadbalancer e999fd88-a0d2-4b82-b1bb-0fa64a187111 to become ACTIVE.
2018-12-11 15:41:53.159219 | ubuntu-xenial | 2018/12/11 15:41:53 [DEBUG] Waiting for state to become: [ACTIVE]
2018-12-11 15:51:53.159775 | ubuntu-xenial | 2018/12/11 15:51:53 [WARN] WaitForState timeout after 10m0s
2018-12-11 15:51:53.160031 | ubuntu-xenial | 2018/12/11 15:51:53 [WARN] WaitForState starting 30s refresh grace period
2018-12-11 15:51:53.160112 | ubuntu-xenial | 2018/12/11 15:51:53 [ERROR] root: eval: *terraform.EvalApplyPost, err: 1 error(s) occurred:
2018-12-11 15:51:53.160147 | ubuntu-xenial |
2018-12-11 15:51:53.160344 | ubuntu-xenial | * openstack_lb_loadbalancer_v2.loadbalancer_1: Error waiting for loadbalancer e999fd88-a0d2-4b82-b1bb-0fa64a187111 to become ACTIVE: timeout while waiting for state to become 'ACTIVE' (last state: 'ERROR', timeout: 10m0s)
2018-12-11 15:51:53.160444 | ubuntu-xenial | 2018/12/11 15:51:53 [ERROR] root: eval: *terraform.EvalSequence, err: 1 error(s) occurred:
2018-12-11 15:51:53.160468 | ubuntu-xenial |
2018-12-11 15:51:53.160613 | ubuntu-xenial | * openstack_lb_loadbalancer_v2.loadbalancer_1: Error waiting for loadbalancer e999fd88-a0d2-4b82-b1bb-0fa64a187111 to become ACTIVE: timeout while waiting for state to become 'ACTIVE' (last state: 'ERROR', timeout: 10m0s)
```

But wait... Now I get it. Most probably amphora doesn't support empty PUT request... And in this case we have to add a special `update` variable like it is done in [openstack_networking_secgroup_v2](https://github.com/terraform-providers/terraform-provider-openstack/blob/00a81613633b002a9af2a3a24078797c44b1033b/openstack/resource_openstack_networking_secgroup_v2.go#L151)

jtopjian:
> But wait... Now I get it. Most probably amphora doesn't support empty PUT request... And in this case we have to add a special update variable like it is done in openstack_networking_secgroup_v2

That might very well be the case and it's worth trying :)

Also, keep in mind you can set the `OS_DEBUG=1` environment variable to see the actual API traffic. That will help with debugging.

kayrus:
> That might very well be the case and it's worth trying :)

yet another XXL PR :(

jtopjian:
> yet another XXL PR

meh - ignore the labels that are automatically placed on PRs. Think of it more like "one more PR closer to a fully functional service" :)

But before you make any PRs, what test are you running and what environment are you running it on? I'd like to run it on Vexxhost. If it passes there, we can investigate why that is.

kayrus:
> meh - ignore the labels that are automatically placed on PRs
When I mentioned XXL, I meant that I need again to look through dozens of `*.go` files and analyze whether they are affected by empty PUT bug (like `UpdateOpts{Name:(*string)(nil), Description:(*string)(nil), AdminStateUp:(*bool)(nil)}`).

```
$ ls openstack/resource_openstack_*.go | grep -v _test | wc -l
61
```

I just wanted to add 6 extra resources, but in order to achieve this, I have to *clean the Augean stables* (don't get me wrong :smile:)

> But before you make any PRs
I run on mitaka neutron-lbaas with F5 driver. It works fine, when empty PUT request is performed.

kayrus:
@jtopjian before I create a new issue explaining empty JSON, does it make sense to introduce a helper function like:

```
func updateOpt(d *schema.ResourceData, key string, meta *string) bool {
        if d.HasChange(key) {
                val := d.Get(key).(string)
                meta = &val
                return true
        }
        return false
}

func resourceSmthV1Update(d *schema.ResourceData, meta interface{}) error {
        update := false
        updateOpts := smth.UpdateOpts{}

        update = updateOpt(d, "name", updateOpts.Name)

        if update {
          doUpdate(updateOpts)
        }
}
```

which will act like:

```
        if d.HasChange("name") {
                update = true
                name := d.Get("name").(string)
                updateOpts.Name = &name
        }
```

Or does it make sense to detect all empty fields of the `updateOpts` struct before executing the `smth.Update` method? Not really sure how to do this, but I suppose something like json.Unmarshalling then json.Marshalling again.

Or does it make sense to detect empty json struct like:

```
{
    "something": {}
}
```

inside the `smth.Update` method?

What do you think?

kayrus:
Not sure whether it looks mature:

```
func isEmpty(v interface{}) bool {
        j, err := json.Marshal(v)
        if err != nil {
                return true
        }
        // "json.Marshal" returns "{}" for empty structures
        return len(j) == 2
}
```

```
        if empty := isEmpty(updateOpts); !empty {
                _, err = resource.Update(blockStorageClient, d.Id(), updateOpts).Extract()
                if err != nil {
                        return fmt.Errorf("Error updating OpenStack resource: %s", err)
                }
        } else {
                log.Printf("[DEBUG] Skipping %s update due to empty options: %#v", d.Id(), updateOpts)
        }
```

jtopjian:
> Or does it make sense to detect empty json struct like:

This feels like it's performing a safety check for something that shouldn't happen, right?

The `Update` function is only ever called if there's a state change, so _something_ is changing. If `Update` runs and the result is an empty body, that feels like something was missed. I'd rather let the empty body pass and return an error and then put effort into figuring out what went awry.

There are also semi-related cases where if one field is updated, then it's mandatory to send other fields even if they didn't change. One of the firewall resources is like this.

So with those two things in mind, I'm inclined to keep the less abstracted methods in place. I'm a fan of choosing explicitness over abstraction even if it looks repetitive or a bit of a mess. This helps lower the barrier to debugging to make it easier for casual users to jump in.

kayrus:
recheck lbaas

kayrus:
Tiny logic did the trick:

```
if updateOpts != (loadbalancers.UpdateOpts{}) {
```

Tests are fine now. Followup for the rest resources: #518 #519

Issue #511, opened 2018-12-07
Network: add description filed to network resources


kayrus:
### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking_floatingip_v2
- openstack_networking_port_v2
- openstack_networking_router_v2
- openstack_networking_secgroup_v2
- openstack_networking_secgroup_rule_v2
- openstack_networking_subnet_v2
- openstack_networking_subnetpool_v2

data sources:
- openstack_networking_network_v2
- openstack_networking_subnet_v2
- openstack_networking_secgroup_v2
- openstack_networking_subnetpool_v2
- openstack_networking_floatingip_v2
- openstack_networking_router_v2

### Expected Behavior
There should be a way to add and update description field for network resources

### References
https://github.com/gophercloud/gophercloud/issues/1342
https://github.com/gophercloud/gophercloud/issues/1230
https://github.com/terraform-providers/terraform-provider-openstack/issues/419

PR #523, opened 2018-12-14
Networking V2: Add description fields to the floatingip resource

kayrus:
Part of: #511


kayrus:
recheck lbaas

jtopjian:
@kayrus Similar here, we need to split this up into individual PRs for each update. This is how other features are managed, even if they're small. For example, individual PRs were created for each implementation of "tags": https://github.com/terraform-providers/terraform-provider-openstack/blob/master/CHANGELOG.md#1120-november-13-2018

kayrus:
Every one is ready, they just need to be merged in a proper order because of dependencies.

Issue #521, opened 2018-12-13
LBaaSv2: Add import feature


kayrus:
It appeared that the following LBv2 resources don't support import:

* openstack_lb_listener_v2
* openstack_lb_loadbalancer_v2
* openstack_lb_member_v2
* openstack_lb_monitor_v2
* openstack_lb_pool_v2

It should be easy to add import feature.

PR #522, opened 2018-12-13
LBaaS: Add l7rule  resource

kayrus:
￼￼Part of #503
Depends on https://github.com/gophercloud/gophercloud/pull/1354.

PR #524, opened 2018-12-14
LBaaSv2: Add resource import feature

kayrus:
Resolves: #521

### DIFF
--- a/openstack/resource_openstack_lb_pool_v2.go
+++ b/openstack/resource_openstack_lb_pool_v2.go
@@ -18,7 +18,7 @@ func resourcePoolV2() *schema.Resource {
 		Update: resourcePoolV2Update,
 		Delete: resourcePoolV2Delete,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			resourcePoolV2Import,
 		},
 
 		Timeouts: &schema.ResourceTimeout{
@@ -230,14 +230,6 @@ func resourcePoolV2Read(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Retrieved pool %s: %#v", d.Id(), pool)
 
-	// Required by import
-	if len(pool.Listeners) > 0 && pool.Listeners[0].ID != "" {
-		d.Set("listener_id", pool.Listeners[0].ID)
-	}
-	if len(pool.Loadbalancers) > 0 && pool.Loadbalancers[0].ID != "" {
-		d.Set("loadbalancer_id", pool.Loadbalancers[0].ID)
-	}
-
 	d.Set("lb_method", pool.LBMethod)
 	d.Set("protocol", pool.Protocol)
 	d.Set("description", pool.Description)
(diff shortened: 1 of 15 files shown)


kayrus:
recheck lbaas

kayrus:
Some resource dependency helpers were added in #522. Better to wait for #522 to be merged, then rebase and work on this PR.

kayrus:
recheck lbaas

mrhillsman:
recheck-lbaas

mrhillsman:
recheck lbaas

kayrus:
recheck lbaas

kayrus:
ready for review.

kayrus:
recheck lbaas

kayrus:
recheck lbaas

jtopjian:
recheck lbaas

openstack/resource_openstack_lb_pool_v2.go
```diff
@@ -227,6 +230,15 @@ func resourcePoolV2Read(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Retrieved pool %s: %#v", d.Id(), pool)
 
+	lbID := d.Get("loadbalancer_id").(string)
+	listenerID := d.Get("listener_id").(string)
+	// Required by import
```
jtopjian: What are you thoughts on putting this in a dedicated import function? The flow would be something like:

1. User imports via pool ID
2. Import function is triggered
3. Do a Get on the pool
4. Do a `d.Set("listener_id")` and `d.Set("loadbalancer_id")`

This way, this part of the code isn't being triggered upon every read.

website/docs/r/lb_member_v2.html.markdown
```diff
@@ -66,3 +66,12 @@ The following attributes are exported:
...
 * `address` - See Argument Reference above.
 * `protocol_port` - See Argument Reference above.
+
+## Import
+
+Load Balancer Pool Member can be imported using the Pool ID and Member ID
+separated by a slash. e.g.:
+
+```
+$ terraform import openstack_lb_member_v2.member_1 c22974d2-4c95-4bcb-9819-0afc5ed303d5/9563b79c-8460-47da-8a95-2711b746510f
```
jtopjian: I'm OK with this, but I wanted to check and see if you were still interested in trying to discover the various load balancer resources only by the UUID?

This is where we would retrieve all load balancers and begin walking all of them until the UUID is found. I haven't tried this myself, so I don't know if it would work.

kayrus:
recheck lbaas

website/docs/r/lb_member_v2.html.markdown
kayrus: Besides:

> This is where we would retrieve all load balancers and begin walking all of them until the UUID is found

As far as I know, this will work only in Neutron LBaaS extension, since it has `/statuses` location. Octavia has only `/status` and I guess it doesn't contain an info for the whole objects subtree (see [your evidence](https://github.com/gophercloud/gophercloud/issues/1352#issuecomment-447072850))

kayrus:
We're getting closer to the end.

PR #526, opened 2018-12-14
Manila: add share acl resource

kayrus:
Part of #502

### DIFF
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/terraform/helper/validation"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/errors"
 	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
 )
 
@@ -95,7 +96,14 @@ func resourceSharedFilesystemShareAccessV2Create(d *schema.ResourceData, meta in
 	})
 
 	if err != nil {
-		return fmt.Errorf("Error granting access: %s", err)
+		detailedErr := errors.ErrorDetails{}
+		e := errors.ExtractErrorInto(err, &detailedErr)
+		if e != nil {
+			return fmt.Errorf("Error granting access: %s: %s", err, e)
+		}
+		for k, msg := range detailedErr {
+			return fmt.Errorf("Error granting access: %s (%d): %s", k, msg.Code, msg.Message)
+		}
 	}
 
 	d.SetId(access.ID)
@@ -169,7 +177,14 @@ func resourceSharedFilesystemShareAccessV2Delete(d *schema.ResourceData, meta in
 	})
 
 	if err != nil {
-		return fmt.Errorf("Error waiting for OpenStack share ACL on %s to be removed: %s", shareID, err)
+		detailedErr := errors.ErrorDetails{}
+		e := errors.ExtractErrorInto(err, &detailedErr)
+		if e != nil {
+			return fmt.Errorf("Error waiting for OpenStack share ACL on %s to be removed: %s: %s", shareID, err, e)
+		}
+		for k, msg := range detailedErr {
+			return fmt.Errorf("Error waiting for OpenStack share ACL on %s to be removed: %s (%d): %s", shareID, k, msg.Code, msg.Message)
+		}
 	}
 
 	// Wait for access to become deleted before continuing
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -1,0 +1,224 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
+)
+
+func resourceSharedFilesystemShareAccessV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSharedFilesystemShareAccessV2Grant,
+		Read:   resourceSharedFilesystemShareAccessV2Read,
+		Delete: resourceSharedFilesystemShareAccessV2Revoke,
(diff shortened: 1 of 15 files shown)


mrhillsman:
recheck

kayrus:
I'm thinking about merging ACL into the share resource as a list of maps, like in [aws security groups](https://github.com/terraform-providers/terraform-provider-aws/blob/5b6353bf2b014092313d76caafc0cda79d0dff64/aws/resource_aws_security_group.go#L73). Does it make sense?

jtopjian:
What's the reasoning behind merging the two?

In general, I would advise against it. If the ACL is part of the share resource, then a user will never be able to decouple an ACL from a share to do things such as use `count`.

`aws_security_group` looks very similar to `openstack_compute_secgroup_v2` which caused a lot of limitations and problems. Thankfully, the Compute Security Group API became deprecated, so we could consider it a legacy resource and design `openstack_networking_secgroup_v2` and `openstack_networking_secgroup_rule_v2` appropriately.

It looks like an [`aws_security_group_rule`](https://github.com/terraform-providers/terraform-provider-aws/blob/5b6353bf2b014092313d76caafc0cda79d0dff64/aws/resource_aws_security_group_rule.go) resource also exists, probably to get around the same kinds of limitations. I have no idea how this resource and `aws_security_group` work together, though - I'd imagine a user must choose one method or another.

It's a best practice to create a decoupled/independent resource for any API call that can be decoupled.

kayrus:
Thanks for the feedback.

kayrus:
recheck

kayrus:
recheck

kayrus:
recheck

kayrus:
recheck

kayrus:
@jtopjian ready for review

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
+)
+
+func resourceSharedFilesystemShareAccessV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSharedFilesystemShareAccessV2Grant,
```
jtopjian: `Grant` and `Revoke` make sense for this, but let's keep `Create` and `Destroy` as the function names.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -0,0 +1,224 @@
...
+		return fmt.Errorf("Error granting access: %s", err)
+	}
+
+	d.SetId(access.ID)
+
+	pending := []string{"new", "queued_to_apply", "applying"}
+	// Wait for access to become active before continuing
+	err = waitForSFV2Access(sfsClient, shareId, access.ID, "active", pending, timeout)
+	if err != nil {
+		return err
```
jtopjian: Let's do:

```go
return fmt.Errorf("Error waiting for OpenStack share ACL on %s to be applied: %s", shareId, err)
```

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -0,0 +1,224 @@
...
+		return fmt.Errorf("Error creating OpenStack sharedfilesystem client: %s", err)
+	}
+
+	sfsClient.Microversion = minManilaMicroversion
+
+	shareId := d.Get("share_id").(string)
+
+	access, err := shares.ListAccessRights(sfsClient, shareId).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "share")
```
jtopjian: This is only checking for a 404 return code. Will this ever return a 404, even if no acls exist?

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -0,0 +1,224 @@
...
+			d.Set("access_to", v.AccessTo)
+			d.Set("access_level", v.AccessLevel)
+
+			return nil
+		}
+	}
+
+	d.Set("access_type", "")
+	d.Set("access_to", "")
+	d.Set("access_level", "")
```
jtopjian: Instead of the above 3 lines, what if we did something like:

```go
if !found {
  d.SetId("")
}
```

Which would effectively remove the resource from state -- same as `CheckDeleted`.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -0,0 +1,224 @@
...
+	err = resource.Retry(timeout, func() *resource.RetryError {
+		err = shares.RevokeAccess(sfsClient, shareId, revokeOpts).ExtractErr()
+		if err != nil {
+			return checkForRetryableError(err)
+		}
+		return nil
+	})
+
+	if err != nil {
+		return err
```
jtopjian: Let's do:

```go
return fmt.Errorf("Error waiting for OpenStack share ACL on %s to be removed: %s", shareId, err)
```

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	// Wait for access to become deleted before continuing
+	pending := []string{"new", "queued_to_deny", "denying"}
+	err = waitForSFV2Access(sfsClient, shareId, d.Id(), "denied", pending, timeout)
+	if err != nil {
+		return err
```
jtopjian: Let's do:

```go
return fmt.Errorf("Error waiting for OpenStack share ACL on %s to be removed: %s", shareId, err)
```

website/docs/r/sharedfilesystem_share_access_v2.html.markdown
```diff
@@ -0,0 +1,141 @@
+---
+layout: "openstack"
+page_title: "OpenStack: sharedfilesystem_share_access_v2"
+sidebar_current: "docs-openstack-resource-sharedfilesystem-share-access-v2"
+description: |-
+  Configure a Shared File System share access list.
+---
+
+# sharedfilesystem\_share\_access\_v2
```
jtopjian: `openstack\_sharedfilesystem\_share\_access\_v2`

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
kayrus: No, it wont. But if there is no share, it will return 404 and ACLs should be removed since they are subobjects of the share object.

PR #542, opened 2018-12-17
Provider: Fix default values for boolean arguments

jtopjian:
This commit fixes the default value for boolean provider arguments. This
is to resolve more strict type features in Terraform v0.12.

/cc @sjwl

I tested this by compiling a new binary with this change and using it on an existing state. No changes were made as these settings are not stored in the state. Therefore, this should not be user impacting.


jtopjian:
Closing in favor of #494  :)

PR #543, opened 2018-12-17
Internal: block storage volume attach cleanup

jtopjian:
For #456

### DIFF
--- a/openstack/blockstorage_volume_attach_v2.go
+++ b/openstack/blockstorage_volume_attach_v2.go
@@ -1,0 +1,35 @@
+package openstack
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/volumeactions"
+)
+
+func expandBlockStorageV2AttachMode(v string) (volumeactions.AttachMode, error) {
+	var attachMode volumeactions.AttachMode
+	var attachError error
+
+	switch v {
+	case "":
+		attachMode = ""
+	case "ro":
+		attachMode = volumeactions.ReadOnly
+	case "rw":
+		attachMode = volumeactions.ReadWrite
+	default:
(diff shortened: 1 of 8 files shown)


openstack/blockstorage_volume_attach_v2.go
```diff
@@ -0,0 +1,35 @@
...
+	default:
+		attachError = fmt.Errorf("Invalid attach_mode specified")
+	}
+
+	return attachMode, attachError
+}
+
+func blockStorageVolumeAttachV2ParseID(id string) (string, string, error) {
+	parts := strings.Split(id, "/")
+	if len(parts) < 2 {
```
ozerovandrei: Maybe it's better to check for just a specific length?
I think that if length is > 2 is bad too, right?
So, what about `if len(parts) != 2 {return "", fmt.Errorf(...)}`.

openstack/blockstorage_volume_attach_v3.go
```diff
@@ -0,0 +1,35 @@
...
+	default:
+		attachError = fmt.Errorf("Invalid attach_mode specified")
+	}
+
+	return attachMode, attachError
+}
+
+func blockStorageVolumeAttachV3ParseID(id string) (string, string, error) {
+	parts := strings.Split(id, "/")
+	if len(parts) < 2 {
```
ozerovandrei: Same here.

openstack/blockstorage_volume_attach_v2.go
jtopjian: Agreed. I was about to do this but wanted to be strict with not introducing any new validation to these cleanup PRs in order to limit the amount of changes.

This is a good idea to do - maybe as a separate PR. I've made a note to myself about it.

ozerovandrei: Alright, sounds great!

jtopjian:
Thanks, @ozerovandrei!

PR #544, opened 2018-12-18
Internal: Block Storage volumes cleanup

jtopjian:
For #456

### DIFF
--- a/openstack/blockstorage_volume_v1.go
+++ b/openstack/blockstorage_volume_v1.go
@@ -26,7 +26,9 @@ func flattenBlockStorageVolumeV1Attachments(v []map[string]interface{}) []map[st
 func expandBlockStorageVolumeV1Metadata(v map[string]interface{}) map[string]string {
 	metadata := make(map[string]string)
 	for key, val := range v {
-		metadata[key] = val.(string)
+		if strVal, ok := val.(string); ok {
+			metadata[key] = strVal
+		}
 	}
 
 	return metadata
--- a/openstack/blockstorage_volume_v1.go
+++ b/openstack/blockstorage_volume_v1.go
@@ -23,17 +23,6 @@ func flattenBlockStorageVolumeV1Attachments(v []map[string]interface{}) []map[st
 	return attachments
 }
 
-func expandBlockStorageVolumeV1Metadata(v map[string]interface{}) map[string]string {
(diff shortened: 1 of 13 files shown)


jtopjian:
note to self: work on snapshot data sources once this is merged.

openstack/blockstorage_volume_v1.go
```diff
@@ -0,0 +1,63 @@
...
+		attachments[i]["device"] = attachment["device"]
+	}
+
+	return attachments
+}
+
+func expandBlockStorageVolumeV1Metadata(v map[string]interface{}) map[string]string {
+	metadata := make(map[string]string)
+	for key, val := range v {
+		metadata[key] = val.(string)
```
ozerovandrei: Is it safe to ignore possible type assertion errors here?

openstack/blockstorage_volume_v2.go
```diff
@@ -0,0 +1,63 @@
...
+		attachments[i]["device"] = attachment.Device
+	}
+
+	return attachments
+}
+
+func expandBlockStorageVolumeV2Metadata(v map[string]interface{}) map[string]string {
+	metadata := make(map[string]string)
+	for key, val := range v {
+		metadata[key] = val.(string)
```
ozerovandrei: Same question about type assertion.

openstack/blockstorage_volume_v1.go
jtopjian: Good point.

Yes, it is safe because Terraform's maps are always `map[string]string` even though they are presented as `map[string]interface{}`. However, it's probably worthwhile to do an assertion check here in case v0.12 changes this.

I'll update this soon.

jtopjian:
@ozerovandrei I just added a new commit: https://github.com/terraform-providers/terraform-provider-openstack/pull/544/commits/6941a948ed716401902190224237a2cdc2b63f1c

I started working on the DNS cleanup and the same type of expansion is being done. This looks to be frequent enough to have a dedicated helper function to reduce code duplication as well as mistakes. You can see that even currently, the wrong function is being used :slightly_smiling_face: 

https://github.com/terraform-providers/terraform-provider-openstack/pull/544/files#diff-289e29fdc3c85185c9b5e90a8b0bea7bL129

Issue #546, opened 2018-12-18
LBaaSv2 / Octavia: Improve LB objects' checks


kayrus:
Current LBaaSv2 behavior doesn't check for pending LB status, when status check is called from its objects. And if LB is in error state, Terraform will keep trying to update LB objects till reaching the timeout.

Here is an example for terraform-openstack-provider 1.12.0 LB status, when there was an l7policy created manually after the terraform and it is in ERROR state:

```json
{
  "statuses": {
    "loadbalancer": {
      "name": "loadbalancer_1",
      "provisioning_status": "ERROR",
      "listeners": [
        {
          "name": "listener_1",
          "provisioning_status": "ACTIVE",
          "pools": [
            {
              "name": "pool_1",
              "provisioning_status": "ACTIVE",
              "healthmonitor": {},
              "members": [],
              "id": "dc48ef18-ebd9-4c67-81a3-977630fb15e3",
              "operating_status": "ONLINE"
            }
          ],
          "l7policies": [
            {
              "name": "test",
              "provisioning_status": "ERROR",
              "rules": [
                {
                  "type": "PATH",
                  "id": "8f782815-1bb5-4010-a68c-fed5433b0bb6",
                  "provisioning_status": "ACTIVE"
                }
              ],
              "action": "REDIRECT_TO_POOL",
              "id": "8b9b92d1-ebcd-4bff-938e-6d0cae139f4c",
              "operating_status": "DEGRADED"
            }
          ],
          "id": "6399c03f-703e-45d8-8b58-67228bf706ef",
          "operating_status": "DEGRADED"
        }
      ],
      "pools": [
        {
          "name": "pool_1",
          "provisioning_status": "ACTIVE",
          "healthmonitor": {},
          "members": [],
          "id": "dc48ef18-ebd9-4c67-81a3-977630fb15e3",
          "operating_status": "ONLINE"
        }
      ],
      "id": "4a9a810a-3794-4e22-aa16-03356554a43d",
      "operating_status": "DEGRADED"
    }
  }
}
```

I'd like to rename `pool_1` into `pool_2`:

```hcl
resource "openstack_lb_loadbalancer_v2" "loadbalancer_1" {
  name = "loadbalancer_1"
  vip_subnet_id = "11deb7f4-b5ea-478e-bd92-8bba5e678b2e"
}

resource "openstack_lb_pool_v2" "pool_1" {
  name            = "pool_2"
  protocol        = "HTTP"
[...]

PR #550, opened 2018-12-19
LBaaSv2: Detect individual LB object statuses, don't rely on LB status

kayrus:
Part of #546

@jtopjian, please watch your eyes :) This is the best what I could do to detect proper statuses. With this PR, situation described in #546 won't happen. Terraform will update LB objects independently on the LB overall status.

This PR is based on #547, so take a look on the second commit. It also depends on https://github.com/gophercloud/gophercloud/pull/1367

My gut feeling says that there could be pitfalls, let's wait for tests.

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -20,7 +20,7 @@ import (
 var lbPendingStatuses = []string{"PENDING_CREATE", "PENDING_UPDATE"}
 
 // lbPendingDeleteStatuses are the valid statuses a LoadBalancer will be before delete
-var lbPendingDeleteStatuses = []string{"PENDING_UPDATE", "PENDING_DELETE", "ACTIVE"}
+var lbPendingDeleteStatuses = []string{"ERROR", "PENDING_UPDATE", "PENDING_DELETE", "ACTIVE"}
 
 // chooseLBV2Client will determine which load balacing client to use:
 // Either the Octavia/LBaaS client or the Neutron/Networking v2 client.
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -137,7 +137,7 @@ func waitForLBV2Member(lbClient *gophercloud.ServiceClient, parentPool *pools.Po
 
 	var refreshFunc resource.StateRefreshFunc
 	if member.ProvisioningStatus != "" {
-		refreshFunc = resourceLBV2MemberRefreshFunc(lbClient, member.PoolID, member.ID)
+		refreshFunc = resourceLBV2MemberRefreshFunc(lbClient, parentPool.ID, member.ID)
 	} else {
 		lbID, err := lbV2FindLBIDviaPool(lbClient, parentPool)
 		if err != nil {
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -42,13 +42,13 @@ func chooseLBV2AccTestClient(config *Config, region string) (*gophercloud.Servic
 	return config.networkingV2Client(region)
 }
 
-func waitForLBV2Listener(lbClient *gophercloud.ServiceClient, id string, target string, pending []string, timeout time.Duration) error {
+func waitForLBV2Listener(lbClient *gophercloud.ServiceClient, id string, lbID *string, target string, pending []string, timeout time.Duration) error {
 	log.Printf("[DEBUG] Waiting for listener %s to become %s.", id, target)
 
 	stateConf := &resource.StateChangeConf{
 		Target:     []string{target},
 		Pending:    pending,
-		Refresh:    resourceLBV2ListenerRefreshFunc(lbClient, id),
+		Refresh:    resourceLBV2ListenerRefreshFunc(lbClient, id, lbID, target),
 		Timeout:    timeout,
 		Delay:      1 * time.Second,
 		MinTimeout: 1 * time.Second,
@@ -70,15 +70,36 @@ func waitForLBV2Listener(lbClient *gophercloud.ServiceClient, id string, target 
 	return nil
 }
 
-func resourceLBV2ListenerRefreshFunc(lbClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
+func resourceLBV2ListenerRefreshFunc(lbClient *gophercloud.ServiceClient, id string, lbID *string, target string) resource.StateRefreshFunc {
+	log.Printf("[DEBUG] resourceLBV2ListenerRefreshFunc: Passed %s Load Balancer ID, %v", *lbID, lbID)
 	return func() (interface{}, string, error) {
-		listener, err := listeners.Get(lbClient, id).Extract()
-		if err != nil {
-			return nil, "", err
+		// Status func to get Listener's status
+		statusFunc := func(lbClient *gophercloud.ServiceClient, loadbalancerID *string) (interface{}, string, error) {
+			listener, err := listeners.Get(lbClient, id).Extract()
+			if err != nil {
+				return nil, "", err
+			}
+
+			if target == "DELETED" {
+				return listener, "", err
+			}
+
+			if listener.ProvisioningStatus != "" {
+				return listener, listener.ProvisioningStatus, nil
+			}
+
+			// First run without knowing parent Load Balancer ID
+			if len(listener.Loadbalancers) > 0 {
+				// Avoid situations, when nil, but not empty string, has been passed
+				if loadbalancerID != nil {
+					*loadbalancerID = listener.Loadbalancers[0].ID
+				}
+			}
+
+			return nil, "", fmt.Errorf("Unable to detect provisioning status for the %s listener", id)
 		}
 
-		// The listener resource has no Status attribute in Neutron lbaasv2 API
-		return listener, lbV2StatusActiveIfEmpty(listener.ProvisioningStatus), nil
+		return lbV2GetProvisioningStatus(lbClient, statusFunc, id, lbID, target)
 	}
 }
 
@@ -121,13 +142,13 @@ func resourceLBV2LoadBalancerRefreshFunc(lbClient *gophercloud.ServiceClient, id
 	}
 }
 
-func waitForLBV2Member(lbClient *gophercloud.ServiceClient, poolID, memberID string, target string, pending []string, timeout time.Duration) error {
+func waitForLBV2Member(lbClient *gophercloud.ServiceClient, poolID, memberID string, lbID *string, target string, pending []string, timeout time.Duration) error {
 	log.Printf("[DEBUG] Waiting for member %s to become %s.", memberID, target)
 
 	stateConf := &resource.StateChangeConf{
 		Target:     []string{target},
 		Pending:    pending,
-		Refresh:    resourceLBV2MemberRefreshFunc(lbClient, poolID, memberID),
+		Refresh:    resourceLBV2MemberRefreshFunc(lbClient, poolID, memberID, lbID, target),
 		Timeout:    timeout,
 		Delay:      1 * time.Second,
 		MinTimeout: 1 * time.Second,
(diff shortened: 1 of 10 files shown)


kayrus:
Forgot to introduce a proper check for octavia clients...

kayrus:
recheck lbaas

kayrus:
Ready for review
In addition waiting for https://github.com/gophercloud/gophercloud/pull/1367 and https://github.com/gophercloud/gophercloud/pull/1364

kayrus:
recheck lbaas

jtopjian:
I played around with this and rearranged things a bit.

The goal of the following commit was to ensure each resource could confidently determine the load balancer ID. This added some additional GETs to each resource, but overall wasn't bad. In addition, I removed the nested functions being passed around in favor of doing the status checks up-front.

If the resource had a populated `ProvisioningStatus`, then it used that for the status check. If not, then it defaulted to the load balancer's status. Caveat that this would only work for Octavia since all resources are checking for `lbPendingDeleteStatuses` during each of their delete calls. But that's OK, as this was only an intermediary commit to lay a foundation:

https://github.com/jtopjian/terraform-provider-openstack/commit/e1f5fd50044ef00b6d6c872d2dff2dd137552c41


I was hesitant that this would work, but sure enough, it did. My hesitation was trusting the resource's provisioning status and not the master load balancer, but that _seemed_ to work out.

Comparing number of API calls when running the "pools" acceptance test:

* current master: 116
* the above commit with octavia: 110

So slight reduction because we can use the resource's native status field. But even when you include the extra GETs to find the load balancer ID, still reduced.

Next, I added support to have the resources' status checked with the load balancer's master status:

https://github.com/jtopjian/terraform-provider-openstack/commit/7a611d8bddcaa3b87cfd37efab6a911c0273ccd4

The diff is unreadable due to the bulk changes, so you'll have to read the file directly. I also removed a bunch of debug lines simply for code clarity - they should be added back.

If the resource was not found in the status, I opted to do one final GET so a true 404 error is returned. 

I also tested this by not doing a GET and simply returning a `DELETED` status, implying that if the resource is not in the LB's master status, it's deleted. I'm on the fence about which is the better approach. We aren't able to do a GET on the member since that requires the pool ID and it doesn't fit well within this pattern of functions right now.

Comparing number of API calls when running the "pools" acceptance test:

* current master: 116
* the above commit with Neutron LBaaSv2: 120

So this slightly increased the number of calls - probably from the final `GET` to return the true 404. But it's much lower than I expected, so that lends more weight to not being concerned about extra traffic to find the LB ID (at least from what is currently in master).

One other method I thought of is to ignore the status of each resource and always use the load balancer's `provisioning_status`. This method is a hybrid of what we currently have in master and the changes I made in the above commits:

1. Each resource finds its load balancer ID (by way of doing a full GET of the parent resource the way it's done in the first commit).
2. In the `waitForState` functions, if a 404 is returned, it's deleted.
3. If it doesn't 404, look at the load balancer's `provisioning_status`.

This method is safe and conservative. It's a _slight_ enhancement of what is currently done in master, which is known to work in the wild. If we were to change the polling methods to use the resources' `provisioning_status` (whether by their native GET or by the load balancer's status), I would want to have this conservative method ready to go in the event that we get numerous bug reports about things being broken.

kayrus:
recheck lbaas

kayrus:
Thanks @jtopjian. Merged and rebased your commits with minor improvements.

kayrus:
recheck lbaas

kayrus:
recheck lbaas

jtopjian:
There might be an issue either with the member acceptance test or these changes:

```
2018-12-20 14:06:34.717124 | ubuntu-xenial | --- FAIL: TestAccLBV2Member_basic (373.60s)
2018-12-20 14:06:34.717247 | ubuntu-xenial |     testing.go:518: Step 0 error: Error applying: 2 error(s) occurred:
2018-12-20 14:06:34.717275 | ubuntu-xenial |
2018-12-20 14:06:34.717374 | ubuntu-xenial |         * openstack_lb_member_v2.member_1: 1 error(s) occurred:
2018-12-20 14:06:34.717401 | ubuntu-xenial |
2018-12-20 14:06:34.717632 | ubuntu-xenial |         * openstack_lb_member_v2.member_1: Error waiting for member 7075d7d5-923c-4a14-ae99-9cba67c57155 to become ACTIVE: Resource not found
2018-12-20 14:06:34.717750 | ubuntu-xenial |         * openstack_lb_member_v2.member_2: 1 error(s) occurred:
2018-12-20 14:06:34.717779 | ubuntu-xenial |
2018-12-20 14:06:34.717956 | ubuntu-xenial |         * openstack_lb_member_v2.member_2: Error waiting for member 7cfbaf15-0455-460e-bfbf-6d97a90f0b84 to become ACTIVE: Resource not found
```

I always run into issues running the member test. I don't know if it's the figures themselves causing the problem or if these changes are interfering.

I ran this test last night and had to change the members' port to 8081 because 8080 was returning a weird error message that 8080 already existed (possibly from the listener?). But in the past, I've seen similar errors and re-running the test without making any changes has the test pass. :confused:

kayrus:
Can we run the test with the `OS_DEBUG=true` log?

jtopjian:
> Can we run the test with the OS_DEBUG=true log?

Unfortunately, no. We tried this with the initial OpenLab integration. The log sizes ended up being too large and broke the test infrastructure :)

I can try running this in a few environments later today and see if I can spot the problem.

kayrus:
recheck lbaas

kayrus:
recheck lbaas

kayrus:
Found an issue: https://github.com/terraform-providers/terraform-provider-openstack/commit/7dc6428966d9a2ebb5c7c2caae73f6dd5e7a6bf9

due to debug code I noticed that PoolID is empty, replaced it with the parentPool.ID:

```
member: &pools.Member{Name:"", Weight:1, AdminStateUp:true, TenantID:"4862856a5a59487e91276d91caecef81", SubnetID:"0f4e333c-f4f7-4a15-ad3a-1e8fc5580f2f", PoolID:"", Address:"192.168.199.111", ProtocolPort:8080, ID:"04da7a44-69dc-4f35-8561-6963c653812f", ProvisioningStatus:"PENDING_CREATE"}
```

jtopjian:
Brilliant! Nice find.

kayrus:
Ready to merge.

kayrus:
Before you merge this, I'd like to explain two side effects of this PR. Despite of we are now confident, that we detect the actual resource status, however:

1. We're now waiting for the parent object only, not the whole LB. In certain cases this is not enough and we can receive:
```json
{
  "NeutronError": {
    "detail": "",
    "message": "Invalid state PENDING_UPDATE of loadbalancer resource b237d38e-064e-41ee-80b5-3c8569df66c8", 
    "type": "StateInvalid"
  }
}
```
Which could be ignored, since it is 409, and terraform will just retry the request.

2. If a parent object is in ERROR state and it is caused by the underlying object, we cannot update neigter underlying nor parent objects. User has to fix or delete LBaaSv2 object causing the ERROR state manually using the openstack CLI or the dashboard.

We can go through the LB objects list and fix both cases within this PR, or ignore it for now, create an issue and discuss possible solutions there later.

jtopjian:
> Invalid state PENDING_UPDATE of loadbalancer resource b237d38e-064e-41ee-80b5-3c8569df66c8

Ah! Yes, this is the situation I was concerned about, but since I wasn't able to reproduce it in any testing I did, I figured relying on the resource's status was good enough. But I can see how this can happen: If a load balancer has multiple branches (pools, listeners, etc), it's possible for one branch to be ACTIVE while another is PENDING_UPDATE, which will cause the root/parent LB to also be in the same state.

This is also what I meant about a conservative solution of always using the load balancer's status as the deciding status.

> Which could be ignored, since it is 409, and terraform will just retry the request.

Right. I'm OK trying this. But if we begin getting reports that this still isn't working (retries are being exhausted before the LB changes to `ACTIVE`, then it'll be best to just scrap all of the per-resource checks and use the LB's status.

IMO, this is a situation where theory isn't aligning with practice (in theory, we can use the resource's status + leverage retries. In practice (complicated deployed load balancers), that doesn't work out). :disappointed: 

Yet Another Method I just thought of: use `/status` for _everything_ and while we're walking the load balancer's statuses, if a parent is not in `ACTIVE` state, return early with that status. So if we're checking for the status of a "member", if the load balancer or pool is in `PENDING`, return that status before we even reach the member status.

Thoughts?

> If a parent object is in ERROR state and it is caused by the underlying object, we cannot update neigter underlying nor parent objects. User has to fix or delete LBaaSv2 object causing the ERROR state manually using the openstack CLI or the dashboard.

My stance has always been that if a resource is in an `ERROR` state, Terraform should step away and the user needs to figure out what went wrong. What you said sounds like the same idea.

One question I have is if this PR changes the probability of this happening somehow?

kayrus:
1. Neutron LBaaSv2 extension: since we're now using `resourceLBV2LoadBalancerStatusRefreshFunc` to get elements' statuses, we can return actual LB status (not subelements) if it is in `PENDING_*` mode and bypass `ERROR` state (this will allow to modify elements, which are not in error state).
2. Octavia: we need to add this hook into the all functions. Or use two dedicated functions for Neutron and Octavia:
* `resourceLBV2LoadBalancerStatusRefreshFuncNeutron`
* `resourceLBV2LoadBalancerStatusRefreshFuncOctavia`

Concerns?

jtopjian:
My current understanding, and correct me if I'm wrong, is that what we have in this PR is the best we can do to match functionality with Neutron and Octavia without beginning to have two very distinct paths.

Advantage of this PR: we're leveraging the individual resource's status instead of the load balancer's status. This is beneficial because it can allow branches of the load balancer tree to update asynchronously from other branches, right? Or is there another benefit?

Disadvantage is that there's a chance that a branch can time out waiting for its peer to perform an update. 

Testing has proven that this PR works well, but we'll be watching out for user reports of breakage.

If we get reports, and if they are showing that these changes are not compatible with the behavior users were previously expecting, then the fix will be to bypass the per-resource status and go straight to the load balancer. 

Being able to do this is already in place because we've added the ability for each resource to detect its parent load balancer, so implementing this is as easy as calling a GET on the load balancer and returning the `provisioning_status`. 

This will work for both Neutron and Octavia, and would actually simplify the code in `lb_shared_v2.go`, because there will only ever be 1 refresh func.

So one question I have is: does the benefit of this PR outweigh relying on the load balancer's `provisioning_status`?

jtopjian:
> and would actually simplify the code in lb_shared_v2.go, because there will only ever be 1 refresh func.

not entirely true - we need to account for the deletion of each individual resource, too.

Let me think about this. 

To be clear: I'm not trying to dismiss this PR, especially after all of the work we've put into it. I'm now trying to understand what would happen if this PR introduced unaccounted issues and how we would resolve them without going back to the previous methods which weren't perfect to begin with.

kayrus:
@jtopjian, please take a look on [this commit](https://github.com/kayrus/terraform-provider-openstack/commit/fc5249f67294ecaa58d5367ca59ec3bbf37d2e17)

jtopjian:
> I'm planning to take a look on what can I do to decrease the probability of 409's

If you can reduce them, great, but I wouldn't put a lot of effort into this. I'm running some tests now with a load balancer that has 2 listeners, each with a pool. Because of the way Terraform performs actions concurrently, it's nearly impossible to avoid 409s. The only way around this is to introduce a mutex on the resource to serialize all actions.

Probably the best way to reduce 409s is to tweak the refresh timers. But there have been requests in the past to make these as small as possible (ie: trade increase in API traffic for faster provisioning).

So this might just be something to live with.

> This PR, or a separate one?

Definitely a new PR. We're not currently checking for this, so let's defer this for now.

I've reviewed everything and I feel good merging this PR. I've worked on a Plan B here (https://github.com/jtopjian/terraform-provider-openstack/commit/b80d734e12d835844f7dd92512c919f31548a840) that we can use in an emergency.

Are you good with this PR? If so, I'll merge it :)

jtopjian:
> I'm planning to take a look on what can I do to decrease the probability of 409's

Another way to think about this is to not think of a 409 as bad, but as an intermediary status update. Lots of API traffic, sure, but this is a natural side-effect of concurrency in complex resource relationships.

kayrus:
Good approach with the additional resources in tests. Regarding [your commit](https://github.com/jtopjian/terraform-provider-openstack/commit/b80d734e12d835844f7dd92512c919f31548a840) - too conservative comparing to [mine](https://github.com/kayrus/terraform-provider-openstack/commit/fc5249f67294ecaa58d5367ca59ec3bbf37d2e17) :)

> Are you good with this PR? If so, I'll merge it :)

Sure, I'll rebase L7 stuff.

jtopjian:
> Regarding your commit - too conservative comparing to mine :)

These are two separate things :) I still want to look at checking for an error. My commit is purely a solution to "whoops! this did not work out at all".

kayrus:
You too ;)

PR #556, opened 2018-12-21
LBaaSv2: Prioritize loadbalancer's status in favour of subelements

kayrus:
Part of the #546

Let's see the test results

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -48,24 +48,16 @@ func chooseLBV2AccTestClient(config *Config, region string) (*gophercloud.Servic
 func waitForLBV2Listener(lbClient *gophercloud.ServiceClient, listener *listeners.Listener, target string, pending []string, timeout time.Duration) error {
 	log.Printf("[DEBUG] Waiting for listener %s to become %s.", listener.ID, target)
 
-	var refreshFunc resource.StateRefreshFunc
-
 	if len(listener.Loadbalancers) == 0 {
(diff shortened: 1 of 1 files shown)


kayrus:
recheck lbaas

jtopjian:
Just a heads up that I probably won't get a chance to look at this for a few days.

kayrus:
recheck lbaas

kayrus:
recheck lbaas

kayrus:
Rebased on #522

kayrus:
recheck lbaas

kayrus:
@jtopjian ready for review

jtopjian:
ack - I'll look at this very shorty.

jtopjian:
@kayrus Instead of nesting the refresh functions, what about modifying each of the individual resource refresh functions like this:

```go
func resourceLBV2ListenerRefreshFunc(lbClient *gophercloud.ServiceClient, lbID, listenerID string) resource.StateRefreshFunc {
  return func() (interface{}, string, error) {
    lb, err := loadbalancers.Get(lbClient, lbID).Extract()
    if err != nil {
      return nil, "", err
    }

    if !strSliceContains(lbSkipLBStatuses, lb.ProvisioningStatus) {
      return lb, lb.ProvisioningStatus, nil
    }

    listener, err := listeners.Get(lbClient, listenerID).Extract()
    if err != nil {
      return nil, "", err
    }

    return listener, listener.ProvisioningStatus, nil
  }
}
```

The loadbalancer check is repeated, but it makes the code more readable.

kayrus:
recheck lbaas

Issue #548, opened 2018-12-19
v0.12 Discussion


jtopjian:
This issue is meant to be a central area to discuss Terraform v0.12 related items.

### Official Status: Should be Compatible

At this time, all acceptance tests are passing with v0.12.

If you have begun testing this provider with v0.12 and have run into problems, please let us know - reports will help prepare us for full v0.12 support.

### Background

Terraform v0.12 is meant to be a major upgrade which includes _a lot_ of new features. The majority of users will be able to upgrade without having to make any changes. However, in order to reach that point, we need to test the OpenStack provider to make sure it's fully compatible.

Terraform v0.12 has not been released yet. You can download an alpha version of it from [here](https://releases.hashicorp.com/terraform/).

All provider plugins _must_ be recompiled with the Terraform core v0.12 code in order to work with v0.12. If you want to test the OpenStack provider with v0.12, you will have to compile it from source, which includes doing a vendor upgrade to `github.com/hashicorp/terraform`.

We will have a binary release of the OpenStack provider made available once v0.12 is released and all testing has passed. This release will be available by using the usual methods of either `terraform init` or downloading and installing the binary yourself from https://releases.hashicorp.com/terraform-provider-openstack/.

More information about v0.12 can be read about here:

* https://www.terraform.io/upgrade-guides/0-12.html
* https://www.hashicorp.com/blog/terraform-0-1-2-preview

### Known Issues

* Unchecked boxes indicate an issue has been identified.
* Checked boxes indicate the issue was fixed.

- [x] Issues with reading in environment variables. ([link](https://github.com/terraform-providers/terraform-provider-openstack/issues/548#issuecomment-448462322))
- [x] Test fixtures and documentation need to reflect new syntax for declaring arguments/attributes of TypeMap ([link](https://github.com/terraform-providers/terraform-provider-openstack/pull/560))
- [x] Change `TypeString` to `TypeBool` for `admin_state_up` and `shared` arguments of the `openstack_networking_network_v2`. ([link](https://github.com/terraform-providers/terraform-provider-openstack/pull/578))
- [x] Issues with `d.GetOkExists` ([link](https://github.com/hashicorp/terraform/issues/21007))

### How to Test

:warning:  __Do not run this on your production configurations or anything you have already deployed! You will not be able to run Terraform v0.11 afterwards!__ :warning: 

> These instructions _will_ change as v0.12 development progresses!

[...]

PR #560, opened 2018-12-27
Provider: Terraform v0.12 TypeMap Fixes

jtopjian:
This commit updates the example code and documentation to reflect how
arguments/attributes of TypeMap should now be declared.

For #548


jtopjian:
I'm going to inquire about this change after the holidays - it's possible there will be backwards compatibility added by the time v0.12 is released.

jtopjian:
This is still showing up as errors in the acceptance tests for v0.12, so I think this is a change we need to account for.

PR #567, opened 2019-01-06
Networking: add port data source

### LINKED ISSUE #512: Feature request: data source: openstack_networking_port_v2
### Affected Resource(s)

### Expected Behavior
Import neutron ports, which were created automatically by the backend, e.g. owner:
- network:f5lbaasv2
- manila:share

### Actual Behavior
There is no `openstack_networking_port_v2` data resource available

### Important Factoids
Ideally it should be possible to import multiple ports, filtered by ListOpts. Then there should be a way to modify them, like add these ports into the security group. I don't like `terraform import` way, because I'd expect to import resources dynamically in the TF manifest. Maybe someone have another thought on it.

### References
Some thoughts could be found here: https://github.com/hashicorp/terraform/issues/10123
https://www.terraform.io/docs/plugins/provider.html#importer

kayrus:
Part of the #512

### DIFF
--- a/openstack/data_source_openstack_networking_port_v2.go
+++ b/openstack/data_source_openstack_networking_port_v2.go
@@ -1,0 +1,259 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/extradhcpopts"
+	p "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+)
+
+func dataSourceNetworkingPortV2() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceNetworkingPortV2Read,
+
+		Schema: map[string]*schema.Schema{
+			"region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"port_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"admin_state_up": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
+			"network_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"tenant_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"device_owner": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"mac_address": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"device_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"fixed_ip": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"allowed_address_pairs": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Set:      allowedAddressPairsHash,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"ip_address": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"mac_address": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"all_fixed_ips": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"all_security_group_ids": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+
+			"all_tags": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"extra_dhcp_option": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"value": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"ip_version": &schema.Schema{
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceNetworkingPortV2Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	networkingClient, err := config.networkingV2Client(GetRegion(d, config))
+
+	listOpts := p.ListOpts{}
+
+	if v, ok := d.GetOk("port_id"); ok {
+		listOpts.ID = v.(string)
+	}
+
+	if v, ok := d.GetOk("name"); ok {
+		listOpts.Name = v.(string)
+	}
+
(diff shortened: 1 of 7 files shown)


openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,224 @@
...
+				Optional: true,
+			},
+
+			"all_fixed_ips": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"return_ids": &schema.Schema{
```
jtopjian: What is the benefit of this over a separate data source which only returns the IDs of the matched ports?

This is causing the `id` field to be overloaded and act as a comma-separated data store of multiple pieces of data. It also causes the data source to have two distinct abilities:

1. Return a single port and attributes
2. Return multiple port UUIDs stored in an overloaded `id` field with no attributes set

Wouldn't this behavior be confusing to the user?

kayrus: Not sure about the user confusion, but I believe this behaviour is covered in the documentation. I also didnt want to duplicate code.

kayrus: Do you offer to create a separate field for returned port IDs? What could be wrong with the primary id?

jtopjian: >  I also didnt want to duplicate code.

Please don't worry about duplicating code when it is done for clarity, both for developers and for the user.

A data source is used to retrieve a single resource. 

There are valid use-cases for a data source to return multiple resources, but right now that doesn't fit with the above statement. Until this is remedied in Terraform core, all proposed solutions are workarounds. 

There is precedence for a data source which returns _only_ the IDs of multiple resources: https://www.terraform.io/docs/providers/aws/d/ami_ids.html. This is an acknowledged workaround by the Terraform core team and the future of this resource is unknown at this time. The benefit of this workaround is that it does not impact the original `aws_ami` resource. So if/when data sources can return more than one resource, `aws_ami_ids` becomes deprecated and there is no impact to the original `aws_ami`.

Because we want to stay in-line with Terraform best-practices, modeling `aws_ami` and `aws_ami_ids` is the only solution worth looking at.

jtopjian: > Because we want to stay in-line with Terraform best-practices, modeling aws_ami and aws_ami_ids is the only solution worth looking at.

I realized the above can be interpreted as an easy way to veto the `return_ids` implementation here and that's not what I'm trying to do.

As I mentioned, there isn't an official solution of how to handle data sources which return multiple results. Because of that, we have to be very careful with a data source's behavior.

It's entirely possible to treat the concept of a "data source" as an arbitrary "sql select" query - returning lots of information about lots of things, but that might cause issues in the future if, say, Terraform supported the ability to return arbitrary data about arbitrary things in a totally different way.

I also realized I didn't answer:

> Do you offer to create a separate field for returned port IDs? 

Yes, but in a separate data source.

There's also an element of consistent design here. No Terraform resource or data source (in any provider) should have mutually exclusive behavior.

> What could be wrong with the primary id?

Technically nothing, but again, it's more about consistent design. An ID should be a unique identifier. Not a unique identifier that also doubles as a list of data. Even if you would set the ID of `openstack_networking_port_ids_v2` as the comma-separated list _and also_ included an `ids` field, it would be better design than only having the comma-separated list. `id` serves as the unique identifier and `ids` services as the list of data.

But on that note, I recommend making the `id` of `openstack_networking_port_ids_v2` a hash of all of the IPs.

kayrus: Acked, created a separate PR for IDs

kayrus:
@jtopjian I believe I added enough filters, ready for review.

jtopjian:
@kayrus Thanks! 

It looks like HashiCorp is applying some blanket changes to all providers, which I only learned about today. One change involves some struct declaration modifications.

I'm going to hold off on reviewing this until this is resolved. You'll need to make similar changes here for consistency and, until it's resolved, we can't be certain what all of those changes are. :/

ozerovandrei:
@kayrus it's about migrating to Go 1.11.x and running `gofmt`: #570

openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,295 @@
...
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/extradhcpopts"
+	p "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
```
jtopjian: Let's remove this alias.

openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,295 @@
...
+	if v, ok := d.GetOk("device_id"); ok {
+		listOpts.DeviceID = v.(string)
+	}
+
+	tags := networkV2AttributesTags(d)
+	if len(tags) > 0 {
+		listOpts.Tags = strings.Join(tags, ",")
+	}
+
+	pages, err := p.List(networkingClient, listOpts).AllPages()
```
jtopjian: `allPages, err`

openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,295 @@
...
+
+	if len(allPorts) == 0 {
+		return fmt.Errorf("No Port found")
+	}
+
+	var port struct {
+		p.Port
+		extradhcpopts.ExtraDHCPOptsExt
+	}
+	var ports []p.Port
```
jtopjian: I'd prefer calling this `p` instead of aliasing the package. But `finalPorts` or `portResults` or something like that also works.

openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+		}
+		if len(sgPorts) == 0 {
+			return fmt.Errorf("No Port found after the 'security_group_ids' filter")
+		}
+		ports = sgPorts
+	}
+
+	if len(ports) > 1 {
+		return fmt.Errorf("More than one Port found (%d). Try to use in a combination with the 'openstack_networking_port_ids_v2' data source", len(ports))
```
jtopjian: Let's not advertise `openstack_networking_port_ids_v2` since 1) it hasn't been added and 2) it's not a standard Terraform resource.

openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,295 @@
...
+		var sgPorts []p.Port
+		for _, p := range ports {
+			for _, sg := range p.SecurityGroups {
+				if strSliceContains(securityGroups, sg) {
+					sgPorts = append(sgPorts, p)
+				}
+			}
+		}
+		if len(sgPorts) == 0 {
+			return fmt.Errorf("No Port found after the 'security_group_ids' filter")
```
jtopjian: Let's either remove this or change it to a debug statement.

openstack/data_source_openstack_networking_port_v2.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+		ports = sgPorts
+	}
+
+	if len(ports) > 1 {
+		return fmt.Errorf("More than one Port found (%d). Try to use in a combination with the 'openstack_networking_port_ids_v2' data source", len(ports))
+	}
+
+	err = p.Get(networkingClient, ports[0].ID).ExtractInto(&port)
```
jtopjian: Instead of doing the extra Get here, you can do this on line 210. For example: https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/portsecurity/doc.go#L7-L26

openstack/data_source_openstack_networking_port_v2.go
kayrus: Done.

openstack/data_source_openstack_networking_port_v2.go
kayrus: Done

openstack/data_source_openstack_networking_port_v2.go
kayrus: Done

openstack/data_source_openstack_networking_port_v2.go
kayrus: Done

openstack/data_source_openstack_networking_port_v2.go
kayrus: Done

openstack/data_source_openstack_networking_port_v2.go
kayrus: It was done by purpose to support `extradhcpopts.ExtraDHCPOptsExt`. Howeve you're right, I can use `ExtractPortsInto`.

PR #569, opened 2019-01-07
Networking: add port ids data source

kayrus:
Part of the #512
Had to introduce a workaround for [this bug](https://github.com/hashicorp/terraform/issues/11806)


kayrus:
@jtopjian ready for review as well

kayrus:
recheck

kayrus:
@jtopjian ready for review. It will make my life easier if you accept this :)

openstack/data_source_openstack_networking_port_ids_v2.go
```diff
@@ -0,0 +1,252 @@
...
+		portIDs = append(portIDs, p.ID)
+	}
+
+	log.Printf("[DEBUG] Retrieved %d Ports: %+v", len(portsList), portsList)
+	params, err := json.Marshal(listOpts)
+	if err != nil {
+		return fmt.Errorf("Unable to marshal listOpts")
+	}
+
+	d.SetId(fmt.Sprintf("%d", hashcode.String(string(params))))
```
jtopjian: How about just using [`hashcode.Strings`](https://github.com/hashicorp/terraform/blob/d4ac68423c4998279f33404db46809d27a5c2362/helper/hashcode/hashcode.go#L27) on `portIDs`?

openstack/data_source_openstack_networking_port_ids_v2_test.go
```diff
@@ -0,0 +1,137 @@
...
+  tags = [
+    "foo",
+    "bar",
+    "qux",
+  ]
+}
+
+data "openstack_networking_port_ids_v2" "ports" {
+  # depends_on     = ["openstack_networking_port_v2.port_1","openstack_networking_port_v2.port_2"]
+  # workaround for the https://github.com/hashicorp/terraform/issues/11806
```
jtopjian: Let's remove these comments. All of our data source tests reference the resource using the "workaround" already. I never knew this was actually an issue - I figured it was the natural way of referencing a data source to a resource :)

openstack/data_source_openstack_networking_port_ids_v2.go
kayrus: Done

openstack/data_source_openstack_networking_port_ids_v2_test.go
kayrus: Done

kayrus:
@jtopjian I completely forgot that there is an openstack based sorting implemented and added a sort for ids before setting the resource value. Removed :) Looking forward for the merge.

Issue #552, opened 2018-12-21
ressource openstack_networking_port_v2: Refresh of port does not refresh security group 


buec011:
### Terraform Version
Terraform v0.11.11
provider.openstack: version = "~> 1.13"

### Affected Resource(s)
openstack_networking_port_v2

### Terraform Configuration Files
```hcl
resource openstack_networking_port_v2 port {
   network_id="123456"
   no_security_groups="true"
}

output port_id {
   value="${openstack_networking_port_v2.port.id}"
}
```

### Expected Behavior
If someone manually add or remove a security group to a port, a terraform plan should mention changes

### Actual Behavior
terraform plan  and apply mention "No changes. Infrastructure is up-to-date."

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. `openstack port set --security-group default  $PORTID `
3. `terraform plan`

jtopjian:
@buec011 This should be possible to detect via the `all_security_group_ids` attribute. `all_security_group_ids` is an output-only attribute and so it won't trigger a change when you run a subsequent plan/apply.

The reason there is a difference between `security_group_ids` and `all_security_group_ids` is because some OpenStack environments will automatically add security groups to the port upon creation and if the user also specifies security groups (or `no_security_groups`), it's not possible to reconcile all of these possibilities in a single field (`security_group_ids`). Therefore, `security_group_ids` is meant to be used as input and `all_security_group_ids` is meant to be the final reconcilled list of all security groups which can be used as output:

```shell
$ terraform apply
$ terraform show
...
openstack_networking_port_v2.port_1:
  id = ed337ecc-17ff-4de0-8cb7-e7a11720f56c
  admin_state_up = false
  all_fixed_ips.# = 0
  all_security_group_ids.# = 0
  allowed_address_pairs.# = 0
  description = 
  device_id = 
  device_owner = 
  extra_dhcp_option.# = 0
  mac_address = fa:16:3e:30:c4:67
  name = port_1
  network_id = c70575b0-6061-4153-bc69-354024d08d42
  no_security_groups = true
  region = RegionOne
  tags.# = 0
  tenant_id = 20c6b343b2b8453d92018a8654343589

$ openstack port set --security-group default ed337ecc-17ff-4de0-8cb7-e7a11720f56c
$ terraform refresh
$ terraform show
...
openstack_networking_port_v2.port_1:
[...]

PR #574, opened 2019-01-08
Networking: add port secgroup associate resource

kayrus:
Resolves the #512

@jtopjian do you know how to create acceptance tests, which ignore idempotency test for specific resource (i.e. `openstack_networking_port_v2`)?


kayrus:
@jtopjian ready for review

kayrus:
recheck

kayrus:
@jtopjian kindly ping

jtopjian:
@kayrus I apologize - this slipped by me. It looks like you figured out how to do the acceptance test - nice work there. :)

Similar to the discussion here (https://github.com/terraform-providers/terraform-provider-openstack/issues/552#issuecomment-450319390), what if we allowed the specified `security_group_ids` to either be amended to existing groups or completely replace the current groups (`enforce_security_groups` - current behavior in this PR).

The use-case I am thinking of is if a cloud admin creates a port on behalf of a user (there are reported cases where users cannot create their own ports - admins delegate a port to the user). These delegated ports might have a security group applied which conforms to an organization's policy.

The user would then have to remember to specify that security group when using this resource. However, if we allowed the ability to amend to the groups, the default configuration would remain.

Thoughts?

kayrus:
@jtopjian if there are cases, when admins delegate a port to the user, I'd rather prefer `port_manage` solution: https://github.com/terraform-providers/terraform-provider-openstack/issues/512#issuecomment-452369739. It will allow users to modify not only the security groups, but other options too.

regarding the https://github.com/terraform-providers/terraform-provider-openstack/issues/552#issuecomment-450319390, and especially https://github.com/terraform-providers/terraform-provider-openstack/issues/552#issuecomment-450111176, it sounds sane, but I predict some logical issues during the update when both `security_group_ids` and `enforce_security_groups` are changed.

kayrus:
Ok, I guess I've clarified the logic. Considering that the current behavior of the `networking_port_v2` is to enforce (because, when you change the list of SG IDs, they will be replaced on the remote side).

Let me know what do your opinion.

kayrus:
Tested on the local environment. Works fine.

kayrus:
As I mentioned in my previous [comment](https://github.com/terraform-providers/terraform-provider-openstack/pull/574#issuecomment-453792176) there were logical issues. I added more tests and now it should be fine.

The only use case, which this logic doesn't cover correctly:
* port is created manually with "default" security group
* port now is managed by terraform with `enforce=false` and new security group is added. (initial security groups are saved into the `old_security_group_ids`)
* someone adds a new security group manually

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,201 @@
...
+
+			"enforce_security_groups": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			"all_security_group_ids": {
+				Type:     schema.TypeSet,
+				Optional: false,
```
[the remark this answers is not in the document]
kayrus: Ok.

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,201 @@
...
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+
+			"old_security_group_ids": {
+				Type:     schema.TypeSet,
+				Optional: false,
```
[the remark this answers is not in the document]
kayrus: Done

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,201 @@
...
+			return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+		}
+
+		var updateOpts ports.UpdateOpts
+		if v, ok := d.GetOk("enforce_security_groups"); ok && v.(bool) == true {
+			securityGroups := resourcePortSecurityGroupsV2(d.Get("security_group_ids").(*schema.Set))
+			updateOpts = ports.UpdateOpts{SecurityGroups: &securityGroups}
+		} else {
+			var allSecurityGroupsSet *schema.Set
+			if d.HasChange("enforce_security_groups") {
```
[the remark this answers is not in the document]
kayrus: > The placement of this HasChange feels weird.

if uses changes from `false` to `true`, 139-140 lines will be executed.

if uses changes from `true` to `false`, `old_security_group_ids` will be used as a base. And `all_security_group_ids` will be used as a base in a regular update.

> Another solution might be to set enforce_security_groups to ForceNew

I thought about this, but in theory resource re-creation could cause dependent resources re-creation. I'd prefer to avoid `ForceNew`.

openstack/resource_openstack_networking_port_secgroup_associate_v2_test.go
```diff
@@ -0,0 +1,381 @@
...
+  name = "default_2"
+}
+
+data "openstack_networking_port_v2" "hidden_port_1" {
+  name = "hidden_port"
+}
+`
+
+func testAccNetworkingV2PortSecGroupAssociateManifest(testCase int, enforce bool) string {
+	switch testCase {
```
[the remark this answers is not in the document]
kayrus: Well, it was easy for me to work on these test cases without modifying dozens of lines. For sure, once tests are written, I can move them back to vars.

jtopjian:
> I thought about this, but in theory resource re-creation could cause dependent resources re-creation. I'd prefer to avoid ForceNew.

Ah - really good point. Agreed.

kayrus:
@jtopjian pushed the code review changes

jtopjian:
@kayrus I might be wrong, but wouldn't this work for the update function?

```go
func resourceNetworkingPortSecGroupAssociateV2Update(d *schema.ResourceData, meta interface{}) error {
  var updateOpts ports.UpdateOpts
  var refreshSecGroups bool

  config := meta.(*Config)
  networkingClient, err := config.networkingV2Client(GetRegion(d, config))
  if err != nil {
  }

  if d.HasChange("enforce_security_groups") {
    refreshSecGroups = true
  }

  enforce := d.Get("enforce_security_groups").(bool)

  if enforce {
    securityGroups := resourcePortSecurityGroupsV2(d.Get("security_group_ids").(*schema.Set))
    updateOpts = ports.UpdateOpts{SecurityGroups: &securityGroups}
  } else {
    initialSecurityGroupSet = d.Get("old_security_group_ids").(*schema.Set)
    requestedSecurityGroupSet := d.Get("security_group_ids").(*schema.Set)

    newSecurityGroupSet := initialSecurityGroupSet.Union(requestedSecurityGroupSet)
    newSecurityGroups := resourcePortSecurityGroupsV2(newSecurityGroupSet)

    updateOpts = ports.UpdateOpts{SecurityGroups: &newSecurityGroups}
  }

  if refreshSecGroups || d.HasChange("security_group_ids") {
    _, err = ports.Update(networkingClient, d.Id(), updateOpts).Extract()
    if err != nil {
      return fmt.Errorf("Error updating OpenStack Neutron Port: %s", err)
    }
  }

  return resourceNetworkingPortV2Read(d, meta)
}
```

For the use-case of a user wanting to remove all security groups from a port, that is supported (and tested) with this, correct?

https://github.com/terraform-providers/terraform-provider-openstack/blob/d2aa492efca67760bce22704ccf0822c503d850d/openstack/resource_openstack_networking_port_secgroup_associate_v2_test.go#L351-L355

Is it worth documenting this?

kayrus:
> I might be wrong, but wouldn't this work for the update function?

No, you're right. This logic is much easier to read and it will cover the case, when the port is modified manually. Besides I was not aware about the `Union` method. Is it possible to convert slice into a `schema.Set` and apply the same logic on the 86?

openstack/util.go
```diff
@@ -239,3 +239,19 @@ func strSliceContains(haystack []string, needle string) bool {
...
+
+func sliceUnion(a, b []string) []string {
+	var res []string
+	for _, i := range a {
+		if !strSliceContains(res, i) {
+			res = append(res, i)
+		}
+	}
+	for _, k := range b {
+		res = append(res, k)
```
jtopjian: I think this needs removed?

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,194 @@
...
+				Set:      schema.HashString,
+			},
+
+			"enforce_security_groups": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			"restore_state_on_delete": {
```
jtopjian: By not restoring the security groups when this resource is deleted, we are permanently changing a resource Terraform didn't create. What is/are the use-cases for doing this?

As well, "state" might not be obvious to the user. I propose `restore_security_groups_on_delete`.

jtopjian: I apologize, but I'm still not understanding the use-case of permanently deleting security groups on system ports. Terraform isn't creating the port, so it would be best if it could leave the port in the state it was in previously. I'm having trouble, and I apologize for that, understanding why we would want to permanently change the port when this `openstack_networking_port_secgroup_associate_v2 ` resource is deleted.

kayrus: Here is an example.

One project, but different types of instance installations.

Now let;s assume we have a clean project, and the final state should be:

installation1 (first terraform state):
 - app1 -> nfs1 (192.168.1.1 and 192.168.1.2 NFS servers)
 - app2 -> nfs1 (192.168.1.1 and 192.168.1.2 NFS servers)
 - db1
 - db2

installation2 (second terraform state):
 - app1 -> nfs2 (192.168.1.1 and 192.168.1.2 NFS servers)
 - app2 -> nfs2 (192.168.1.1 and 192.168.1.2 NFS servers)

When you deploy installation1, and there are no shares exist, there are no manila ports available, so you cannot add them into the security group. Terraform provisions manila shares, and `openstack_networking_port_secgroup_associate_v2` with the `openstack_networking_port_ids_v2` dependency detects just created manila system ports (let's assume two of them).

When you deploy installation2, there are no shares yet, but there are already two manila system ports available, which were created previously. So you can add them into the security group (use `openstack_networking_port_secgroup_associate_v2` resource). Terraform provisions manila shares, and `openstack_networking_port_secgroup_associate_v2` with the `openstack_networking_port_ids_v2` dependency detects ports, saves previous state and everybody are happy.

*Now the use case*: you don't want to have an `installation1` anymore. You run `terraform destroy` for it. It restores the initial security groups (none) for manila ports. installation2 fails, because security groups were removed from manila system ports and it doesn't have an access to 192.168.1.1 and 192.168.1.2 NFS servers anymore. Operator, who did run `terraform destroy` screwed up.

kayrus: Do you have any new thoughts?

jtopjian: I hope to get back to this shortly. I haven't had much time to dedicate to this.

jtopjian:
I'm sorry for the delay. I've been trying to work out the use-cases of this and finally had a few good hours to sit down and think about it.

Given there's an existing port:

1. If that port has one or more existing security groups:
    
    a. A user chooses `enforce_security_groups`, then this resource will ensure the port only has the security groups defined in `security_group_ids`. Upon update, the current value of `security_group_ids` can be used to set the security groups. Upon deletion, the security groups can be set to `[]`, giving one final axe swing to any security groups on the port. 
    
    b. A user does not choose `enforce_security_groups`, then this resource will always amend security groups to those that are already defined. Upon update, a diff should be done of the changed `security_group_ids` and the security groups should be removed/added as appropriate. Upon deletion, removing the security groups defined in `security_group_ids` will result as if this resource did not exist. This is not to say that the port will only have the original security groups - it could have more.

2. If that port has no existing security groups:
    
    a. And a user chooses `enforce_security_groups`, then this resource will always ensure only the security groups defined in `security_group_ids` will remain. Upon update, the current value of `security_group_ids` can be used to set the security groups. Upon deletion, the security groups can be set to `[]`, giving one final axe swing to any security groups on the port. 

    b. And a user does not choose `enforce_security_groups`, then this resource will only manage the security groups defined in `security_group_ids`. Upon update, a diff should be done of the changed `security_group_ids` and the security groups should be removed/added as appropriate. Upon delete, removing the security groups defined in `security_group_ids` will result in a port whose state represents that if this resource didn't exist.

3. A user defines two or more `openstack_networking_port_secgroup_associate_v2` on the same port.

    a. If each have `enforce_security_groups`, then this is going to cause competing behavior. This should not be a valid combination.

    b. If one has one `enforce_security_groups`, then this is going to cause the `enforce` one to clobber the others. This should also not be a valid combination.

    c. None have `enforce_security_groups`, then the resources will successfully co-exist. This is the only combination that is valid when using more than one resource of this.

None of these use-cases require `old_security_group_ids` or `restore_security_groups_on_delete`. 

`old_security_group_ids` is required for situations when someone wants to use `enforce`, but not fully enforce - only sometimes. It's the same with `restore_security_groups_on_delete` - this is another "enforce but not really enforce" option.

These options exist because we're working with a situation where Terraform is only managing a partial lifecycle of something. If Terraform managed the full lifecycle, then these options wouldn't be needed at all. 

We can take that a step further and define this resource's full lifecycle as that of the `security_group_ids` it defines - meaning, remove `enforce_security_groups`. Now Terraform has full control over the state being defined within it.

What we have now is a pure solution, right? In my opinion, this should be the default definition of this resource. 

I'm tempted to say that this pure solution is the only valid solution, but I can understand the need to have something like `enforce`. Accepting that `enforce` opens the door to a number of complexities, is it fair enough to say that anything that `enforce` cannot cover should be an edge-case (even edgier than the audience already requiring this resource) and that something customized by an external script should be used to solve?

Even while thinking this through, I thought of a few more edge cases that someone might want:

* Don't set the security groups to `[]` on deletion when `enforce` is used. Only unset those defined in `security_group_ids` because someone might want the ability to have security groups, which were applied by some other means before deletion, to remain.

* When `enforce` is used, only ever delete pre-existing security groups during the Create phase. Never during update.

tl;dr: IMO, we should consider the applicable use-cases of this resource those that have been outlined above and remove `old_security_group_ids` and `restore_security_groups_on_delete`. That's not written in stone - I realize I might have forgotten something key to all of this - but the use cases I outlined provide users with more ability than they had before when confronted with non-standard ports and provide a clear list of ways this resource can be used.

kayrus:
Rebased with a new logic. I hope it is clear now, however I'm free to add more use cases into the documentation.

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,176 @@
...
+
+	port, err := ports.Get(networkingClient, portID).Extract()
+	if err != nil {
+		return fmt.Errorf("Unable to get %s Port: %s", portID, err)
+	}
+
+	log.Printf("[DEBUG] Retrieved Port %s: %+v", portID, port)
+
+	var updateOpts ports.UpdateOpts
+	if v, ok := d.GetOk("enforce"); ok && v.(bool) == true {
```
jtopjian: If the default value is changed to false in the schema, instead of `GetOk`, we will want to do:

```go
var enforce bool
if v, ok := d.GetOkExists("enforce"); ok {
  enforce := v.(bool)
}

if enforce {
...
```

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,176 @@
...
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	port, err := ports.Get(networkingClient, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "port")
+	}
+
+	if v, ok := d.GetOk("enforce"); ok && v.(bool) == true {
+		d.Set("security_group_ids", port.SecurityGroups)
```
jtopjian: I think this would cause a diff.

If someone adds a security group to a port (outside of Terraform), then `port.SecurityGroups` will be different than `security_group_ids`.

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,176 @@
...
+
+	_, err = ports.Update(networkingClient, portID, updateOpts).Extract()
+	if err != nil {
+		return fmt.Errorf("Error associating %s port with '%s' security groups: %s", portID, strings.Join(securityGroups, ","), err)
+	}
+
+	d.SetId(portID)
+
+	log.Printf("[DEBUG] Storing old security group IDs into the 'old_security_group_ids' attribute: %#v", port.SecurityGroups)
+	d.Set("old_security_group_ids", port.SecurityGroups)
```
jtopjian: `old_security_group_ids` is no longer defined in the schema, so we should remove this.

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	var updateOpts ports.UpdateOpts
+
+	if d.Get("enforce").(bool) {
+		securityGroups := expandToStringSlice(d.Get("security_group_ids").(*schema.Set).List())
+		updateOpts = ports.UpdateOpts{SecurityGroups: &securityGroups}
+	} else {
```
jtopjian: I think this `else` case will look similar to this: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_instance_v2.go#L752-L783

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			return fmt.Errorf("Error updating OpenStack Neutron Port: %s", err)
+		}
+	}
+
+	return resourceNetworkingPortV2Read(d, meta)
+}
+
+func resourceNetworkingPortSecGroupAssociateV2Delete(d *schema.ResourceData, meta interface{}) error {
+	if d.Get("enforce").(bool) == false {
```
jtopjian: If `enforce` is false, then we will want to remove the groups set in `security_group_ids`, right?

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
jtopjian: Ah - you're right.

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
jtopjian: Yep, right again - thanks!

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
jtopjian: hmm... I thought the use-cases I outlined (https://github.com/terraform-providers/terraform-provider-openstack/pull/574#issuecomment-457952943) accounted for this? 

If enforce is set to false, then this resource should reverse / backtrack on everything it did (1b and 2b). `security_group_ids` is the user-defined set of security groups and does not include security groups which existed on the port beforehand or which other resources or actions placed on the port at other times. Removing these should not interfere with any other security groups placed on it (3c).

Attempting to illustrate it:

```
Phase 1:

existing port secgroups: 1, 2
openstack_networking_port_secgroup_associate_v2.foo.security_group_ids: 3, 4

result on port: 1, 2, 3, 4

Phase 2:

existing port secgroups: 1, 2, 3, 4
openstack_networking_port_secgroup_associate_v2.foo.security_group_ids: 3, 4
ad-hoc command: 5
openstack_networking_port_secgroup_associate_v2.bar.security_group_ids: 6

result on port: 1, 2, 3, 4, 5, 6

Phase 3

existing port secgroups: 1, 2, 3, 4
*destroy* openstack_networking_port_secgroup_associate_v2.foo.security_group_ids: 3, 4
openstack_networking_port_secgroup_associate_v2.bar.security_group_ids: 6

result on port: 1, 2, 5, 6
```

Let me know if this sounds correct - I want to make sure I'm not missing something important here.

kayrus: Yep, this is the current logic I introduced after [your comment](https://github.com/terraform-providers/terraform-provider-openstack/pull/574#discussion_r251347866).
Before I misunderstood [this comment](https://github.com/terraform-providers/terraform-provider-openstack/pull/574#issuecomment-457952943) and aligned it with my [use case](https://github.com/terraform-providers/terraform-provider-openstack/pull/574#discussion_r247961493), but then I re-read your comment and realized that my use case could be solved using a separate (not shared) security group.

kayrus: Just an update, we've missed the use case, when `ad-hoc command` removes the desired SGs from the target port. `enforce=false` won't detect the case, when the desired security group was removed.

Should we cover this situation?

jtopjian: Good catch.

This situation makes sense when you think about it from the `Create` phase - the security group would be added. But since `d.Set("security_group_ids")` is only being done when enforce is true, there will be no way to generate a diff. 

Therefore, we could compare the `all_security_group_ids` versus what's in `security_group_ids`. In other words, do a `HasChange`-like check during `Read` and make sure the `security_group_ids` set is completely in the `all_security_group_ids` set. If a group is missing, then do `d.Set("security_group_ids")` with the remaining groups. This should trigger a diff and reconcile.

I feel that we should cover this situation because this would be the standard behavior of a "normal" resource without any kind of need for "enforce". The goal of Terraform is to bring a resource to a state that matches what the user defined, and re-adding the security group falls in line with this.

If we didn't support this resource, then a user could remove _all_ security groups and this resource would end up serving no purpose. It'd be a weird situation.

Thoughts?

jtopjian: > In addition it looks like a new logic should come into the openstack_networking_port_v2 resource as well.

Yeah, I'm keeping that in mind. Let's tackle one thing at a time, though The port resource is delicate :)

kayrus: Now the code handles ad-hoc changes.

openstack/resource_openstack_networking_port_secgroup_associate_v2.go
```diff
@@ -0,0 +1,208 @@
...
+
+	d.Set("all_security_group_ids", port.SecurityGroups)
+
+	if enforce {
+		d.Set("security_group_ids", port.SecurityGroups)
+	} else {
+		allSet := d.Get("all_security_group_ids").(*schema.Set)
+		desiredSet := d.Get("security_group_ids").(*schema.Set)
+		actualSet := allSet.Intersection(desiredSet)
+		if !actualSet.Equal(desiredSet) {
```
kayrus: @jtopjian, I have doubts about this condition. Basically we can always set the `security_group_ids` attribute here.

kayrus: ok, I'll keep it.

kayrus:
@jtopjian any further comments or use cases, which were not covered?

kayrus:
Is it required to add a region/tenantid into the attributes schema? If not, I'm ok to merge what we have.

jtopjian:
> Is it required to add a region/tenantid into the attributes schema?

Region is already in the schema.

With regard to tenantid, it's definitely not required. If someone has a use-case of needing to use this resource on behalf of another tenant, they can open a feature request or PR - it can easily be added later without issue.

kayrus:
I just realized, that we work with the existing port_id directly, no need to have a tenantid

kayrus:
No more local-exec hooks. Cheers!

jtopjian:
> I just realized, that we work with the existing port_id directly, no need to have a tenantid

There are some users that run Terraform as an admin user and use the networking tenant/project fields to scope to a project and perform administrative tasks. A tenant/project ID field in this resource would be for the same type of scenario.

kayrus:
Admins have an access to all ports. port data source, which resolves port_id - has a [tenant filter](https://www.terraform.io/docs/providers/openstack/d/networking_port_v2.html#project_id)

jtopjian:
Right, that will return the ports owned by a certain tenant.

But if you use a `tenant_id` or `project_id` field as _input_ to a resource, Neutron allows you to scope to that project (without having to generate a new token) and perform the action as if you were that project.

PR #571, opened 2019-01-08
Internal: Floating IP cleanup

ozerovandrei:
Move Networking V2 networks and floating IPs helpers functions outside
of resource_openstack_networking_floatingip_v2.go.
Update debug and error messages.

For #456

PR #578, opened 2019-01-09
Networking V2: Networks cleanup

ozerovandrei:
Add StateRefreshFunc and expandNetworkSegments functions into the
"networking_network_v2.go".
Fix error and debug messages.
Set tags only after ensuring that network is created.
Change internal type of the "admin_state_up" and "shared" arguments to
boolean.
Cleanup resource website documentation.


jtopjian:
recheck

jtopjian:
@ozerovandrei Nice work on this :)

With regard to the type changes being made in the schema:

* The reason that the bool fields were originally strings was because `GetOkExists` wasn't available yet. So we'll want to use `GetOkExists` when retrieving those fields. I think this is the cause of the acceptance test failure. 

Also, please test and make sure Terraform will allow existing configurations that use `"true"` (string with quotes) when the type has been changed to `bool`. The only way to do this is to mirror what a user would do:

1. Use the latest release and create a network with `admin_state_up` defined as `"true"`.
2. Compile a binary with this patch.
3. Run `terraform init` and `terraform plan` on the existing configuration and see if Terraform reports changes or problems.

If a change is reported, we will have to go back to `TypeString` and either do a state migration separately or wait until v2.

It's worth trying out this change before Terraform v0.12 comes out.

* The changes from `TypeList` to `TypeSet` are [allowed](https://www.terraform.io/docs/extend/best-practices/versioning.html#example-minor-number-increments), but do we need to make this change? Personally, I prefer `TypeList` since it uses 0-based indexing.

jtopjian:
I did some experimenting with the `TypeBool` stuff:

* As far as I can tell, we can make the change and users _should not_ see any issues.
* Ignore what I said about using `GetOkExists`. `Get` is the correct one to use to match previous functionality.
* I'm still able to declare `"true"` even though the field is `TypeBool`, which means users won't have to make changes, and that's the most important thing.

However, on lines 249-251, there are some boolean-to-string conversions being done that need to be removed:

```go
d.Set("admin_state_up", n.AdminStateUp)
d.Set("shared", n.Shared)              
d.Set("external", n.External)          
```

(external is actually a bug)

I'd prefer this change to `TypeBool` be done in a dedicated PR to isolate the change and have a changelog entry. Let's do this soon (before v0.12).

With regard to the `TypeSet` change, I'm still curious if this is required or not. If it is, I'm totally fine doing it, but again, maybe a separate PR.

ozerovandrei:
@jtopjian 

I have several ideas: 

1. About Set/List. I prefer Set because it's unordered collection: [TypeSet|https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset] and the ordering itself has no impact on the behavior of the resource.
I think that [TypeList|https://www.terraform.io/docs/extend/schemas/schema-types.html#typelist] is better for cases when we need to make a resource depended on the ordering of the elements in some attribute.

2. About users who did not explicitly declare `admin_state_up` in their configs. Can we just use `Default` for that argument? Like that:
```
"admin_state_up": {
    Type:     schema.TypeBool,
    Default: true,
    Optional: true,
    ForceNew: false,
    Computed: true,
},
```

jtopjian:
> About Set/List.

Correct - this is about ordering, but just to make sure we're talking about the same thing it's the order that the user defined things, not a sorted order like alphabetical or numerical.

If the user sends `[c, a, b]` and the API returns that same order, then the type should be `TypeList`. 

But if the user sends `[c, a, b]` and the API returns `[a, b, c]`, then we should use `TypeSet`. 

Maybe these fields have always done the latter and no one has reported it. In that case, then a change to `TypeSet` is correct.

> About users who did not explicitly declare admin_state_up in their configs.

I don't think we can do default+computed - I think Terraform will give an error. And, IMO, we need to preserve the "not-set" behavior. By setting a default value, `admin_state_up` will always be sent in the API request. If this was a brand new field, I could see using "default", but since we need to ensure nothing changes on the user's side, I think the steps I put in the summary are the most we can do.

I recall asking about this in Slack a few months ago and it was recommended to make this change before Terraform v0.12, since there will be more explicit types used. So I do think we need to make this change, it's just making sure it won't break backwards compatibility.

ozerovandrei:
Makes sense.

I will remove `string` -> `bool` changes from this PR and create a different issue to address those changes.

jtopjian:
Looks like Lines 144, 145, 153, and 154 need to be removed. Tests pass for me with those changes.

ozerovandrei:
recheck

jtopjian:
@ozerovandrei My bad - I should have let you merge #571. Too many PRs :)

I felt bad and decided to handle the rebase on your behalf :)

Also, the test was failing due to one remaining `net` => `network_1` reference. 

Pending test results, this looks good to me. If you're comfortable with the List/Set changes, I trust your judgement here.

Merge when ready :)

PR #584, opened 2019-01-10
Internal: cleanup Networking V2 Ports

ozerovandrei:
Add StateRefreshFunc(), expandNetworkingPortFixedIpsV2 into the
"networking_port_v2.go".

Change "fixed_ip" argument type to "TypeSet".

Use "expandToStringSlice()" function to retrieve "security_groups"
value. This change also affects "openstack_lb_loadbalancer_v2" resource
code.

Add common "boolPtrArgument()" into "util.go" to work with some special
arguments like "admin_state_up".

Fix debug and error messages.

For #456

### DIFF
--- a/openstack/data_source_openstack_networking_port_ids_v2.go
+++ b/openstack/data_source_openstack_networking_port_ids_v2.go
@@ -180,16 +180,16 @@ func dataSourceNetworkingPortIDsV2Read(d *schema.ResourceData, meta interface{})
 
 	allPages, err := ports.List(networkingClient, listOpts).AllPages()
 	if err != nil {
-		return fmt.Errorf("Unable to list Ports: %s", err)
+		return fmt.Errorf("Unable to list openstack_networking_ports_v2: %s", err)
 	}
 
 	allPorts, err := ports.ExtractPorts(allPages)
 	if err != nil {
-		return fmt.Errorf("Unable to retrieve Ports: %s", err)
+		return fmt.Errorf("Unable to retrieve openstack_networking_ports_v2: %s", err)
 	}
 
 	if len(allPorts) == 0 {
-		return fmt.Errorf("No Port found")
+		return fmt.Errorf("No openstack_networking_port_v2 found")
 	}
 
 	var portsList []ports.Port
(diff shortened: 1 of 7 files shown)


jtopjian:
Tests are failing due to the change from `TypeList` to `TypeSet`.

These List/Set changes are giving me scary flashbacks. For this port resource, see https://github.com/hashicorp/terraform/pull/13056 (and all links) for lots and lots of details about this specific topic.

You can see that I once tried to change this to `TypeSet` but immediately got bug reports about things breaking :fearful: 

For Fixed IPs specifically, users are relying on the 0-based indexing to reference the DHCP-assigned IP of the port.

It's because of this that I am always uneasy about changing List/Set once they are set in the schema.

I'm absolutely open to discussion about changing the types of certain fields, but let's please not do it to this resource :sweat_smile:  Additionally, I'd prefer to do these changes outside of the cleanup in case they trigger bug reports like the above link.

Changing from List to Set is a forwards-compatible change _except_ when the user has begun relying on either the 0-based indexing or hash-based indexing for dependant resources. We need to be really careful when doing this.

jtopjian:
@ozerovandrei The test is failing since `fixed_ip` also needs changed back to `TypeList`.

There are cases where users are using the zero-based indexing for _both_ `fixed_ip` and `all_fixed_ips`.

The test fixture `testAccNetworkingV2Port_allowedAddressPairs_1` is a good example of how users were reported to be using `fixed_ip` as a list.

Here's where it gets confusing, specifically for this port scenario:

The user cannot use `all_fixed_ips` here because of the case where they defined the vrrp port with two `fixed_ip`s. The order of `fixed_ip` and `all_fixed_ips` could be different, so the user needs to be confident about which fixed ip they are using in `allowed_address_pairs`.

It's a complicated scenario because of how ports allow you to define IPs, have Neutron allocate IPs, and the API enforcing a numerical sort of that order :/

Maybe it's possible to totally refactor this system of `fixed_ip` and `all_fixed_ips` using things like `SuppressDiff` and `CustomizeDiff` (which were not available two years ago when I ran into these problems). But we'd need to either create a new port resource or add new fields to the existing port resource to cleanly do this.

ozerovandrei:
@jtopjian totally makes sense, thanks!

PR #586, opened 2019-01-11
Internal: Update CheckDeleted message

jtopjian:
This commit revises the CheckDeleted message of cleaned up resources.

For #456 

@ozerovandrei This adds `CheckDeleted` to the `Delete` function when it's appropriate to do so. It also adds the ID of the resource to the message - this would have bugged me until I fixed it :)

### DIFF
--- a/openstack/resource_openstack_blockstorage_volume_v1.go
+++ b/openstack/resource_openstack_blockstorage_volume_v1.go
@@ -175,7 +175,8 @@ func resourceBlockStorageVolumeV1Read(d *schema.ResourceData, meta interface{}) 
 
 	v, err := volumes.Get(blockStorageClient, d.Id()).Extract()
 	if err != nil {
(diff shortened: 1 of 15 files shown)


jtopjian:
@kayrus Are you saying you shouldn't have gotten that message or that this PR would make it better because it includes the ID?

jtopjian:
Actually, what would be easier is to just modify this line:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/util.go#L41

To include `d.Id()`.

I will open a new PR tomorrow.

@kayrus I'm still curious if you ran into a problem with this, though.

kayrus:
@jtopjian this PR would fix the issue, which I faced recently. I don't understand why did you close it...

jtopjian:
>  I don't understand why did you close it...

Because there's a much easier way of handling this fix 🙂 I'll have a new PR shortly.

PR #587, opened 2019-01-11
Compute v2: add flavor_id to flavor data source

### LINKED ISSUE #474: Compute v2: Add flavor_id to openstack_compute_flavor_v2 data source
Adding a `flavor_id` field to the `openstack_compute_flavor_v2` data source will allow users to retrieve a flavor when only the ID is known. The field must be called `flavor_id` so it doesn't clash with the internal `id` field.

This field should be set to `ConflictsWith` with the name field, if that will work.

kayrus:
Resolves #474

### DIFF
--- a/openstack/data_source_openstack_compute_flavor_v2.go
+++ b/openstack/data_source_openstack_compute_flavor_v2.go
@@ -22,10 +22,18 @@ func dataSourceComputeFlavorV2() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"flavor_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"name"},
+			},
+
 			"name": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"flavor_id"},
 			},
 
 			"min_ram": {
@@ -92,6 +100,15 @@ func dataSourceComputeFlavorV2Read(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
 	}
 
+	var flavor *flavors.Flavor
+	if v := d.Get("flavor_id").(string); v != "" {
+		flavor, err = flavors.Get(computeClient, v).Extract()
+		if err != nil {
+			msg := fmt.Sprintf("Error retrieving openstack_compute_flavor_v2 %s", v)
+			return CheckDeleted(d, err, msg)
+		}
+	}
+
 	listOpts := flavors.ListOpts{
 		MinDisk:    d.Get("min_disk").(int),
 		MinRAM:     d.Get("min_ram").(int),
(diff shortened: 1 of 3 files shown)


kayrus:
@jtopjian ready for review

openstack/data_source_openstack_compute_flavor_v2.go
```diff
@@ -92,6 +100,15 @@ func dataSourceComputeFlavorV2Read(d *schema.ResourceData, meta interface{}) err
 		return fmt.Errorf("Error creating OpenStack compute client: %s", err)
 	}
 
+	var flavor *flavors.Flavor
+	if v := d.Get("flavor_id").(string); v != "" {
+		flavor, err = flavors.Get(computeClient, v).Extract()
+		if err != nil {
+			msg := fmt.Sprintf("Error retrieving openstack_compute_flavor_v2 %s", v)
+			return CheckDeleted(d, err, msg)
```
jtopjian: `CheckDeleted` isn't needed here. `CheckDeleted` removes a resource from the Terraform state when the resource can't be found (404).

Since this is a data source, it's more of a "stateless" query that is refreshed each time Terraform is run.

Instead, we should return the same message as on lines 179-180.

openstack/data_source_openstack_compute_flavor_v2.go
```diff
@@ -100,78 +117,79 @@ func dataSourceComputeFlavorV2Read(d *schema.ResourceData, meta interface{}) err
 
 	log.Printf("[DEBUG] openstack_compute_flavor_v2 ListOpts: %#v", listOpts)
 
-	var flavor flavors.Flavor
-	allPages, err := flavors.ListDetail(computeClient, listOpts).AllPages()
-	if err != nil {
-		return fmt.Errorf("Unable to query OpenStack flavors: %s", err)
-	}
+	if flavor == nil {
```
jtopjian: Let's move this to line 111 so `listOpts` becomes part of this block.

openstack/data_source_openstack_compute_flavor_v2.go
jtopjian: Oh, sorry. 

We should still get rid of `CheckDeleted` and do:

```go
return fmt.Errorf("Error retrieving openstack_compute_flavor_v2 %s: %s", v, err)
```

jtopjian: Nope, I got it wrong again.

We need to do two things:

1. Account for a 404 where we return an error that says "Your query returned no results..."
2. Account for any other error and return "Error retrieving openstack_compute_flavor_v2..."

`CheckDeleted` comes close to doing this, but it does `return nil` instead of accounting for the first item.

I suggest doing something like:

```go
if err != nil {
    return fmt.Errorf("Your query returned no results ..."
  }

  return fmt.Errorf("Error retrieving openstack_compute_flavor_v2 %s: %s", v, err)
}
```

kayrus: I disagree. What if user run `terraform destroy` and the flavor doesn't exist. It will raise the same error I faced [here](https://github.com/terraform-providers/terraform-provider-openstack/pull/586#issuecomment-453383292).

kayrus: Ah, I mixed up data source and resource. This should be ok.

openstack/data_source_openstack_compute_flavor_v2.go
kayrus: I did differently. Now it looks more logical.

jtopjian:
recheck

jtopjian:
@kayrus Thinking about this more, and from looking at #591:

I agree that doing a `Get` when specifying `flavor_id` is correct since `ListOpts` does not allow and `ID`. 

But should more fields be added to `flavor_id`'s `ConflictsWith`?

For example, what if a user queries by `flavor_id` _and_ `min_ram`?

I propose the following:

* Use `ConflictsWith` on `flavor_id` and these three fields: https://github.com/terraform-providers/terraform-provider-openstack/blob/dded0c91a6498278021451b7160dac9b655c6de0/openstack/data_source_openstack_compute_flavor_v2.go#L117-L119

* If a flavor is returned by `flavor_id`, add it to the `allFlavors` slice so it goes through the filtering done here: https://github.com/terraform-providers/terraform-provider-openstack/blob/dded0c91a6498278021451b7160dac9b655c6de0/openstack/data_source_openstack_compute_flavor_v2.go#L135

This makes the Get and List mutually exclusive. It also removes us having to do an ugly long string of `ConflictsWith` for _every_ field.

Thoughts?

kayrus:
@jtopjian sounds sane, but it will change the existing behavior, when `name`, `min_disk` and `min_ram` are specified, but the result has len=1 and the name doesn't correspond the actual flavor name. Probably this is a bug, therefore my next commit with your suggestion will fix it.

kayrus:
@jtopjian pushed the code review changes

openstack/data_source_openstack_compute_flavor_v2_test.go
```diff
@@ -207,3 +244,15 @@ func testAccComputeV2FlavorDataSource_extraSpecs(flavorName string) string {
...
 }
+
+func testAccComputeV2FlavorDataSource_flavorID(flavorName string, flavor *flavors.Flavor) string {
+	flavorResource := testAccComputeV2Flavor_extraSpecs_1(flavorName)
+
+	return fmt.Sprintf(`
+          %s
+
+          data "openstack_compute_flavor_v2" "flavor_1" {
+            flavor_id = "%s"
```
jtopjian: This test is currently failing (admin tests are not run in OpenLab).

This needs to be changed to:

```
flavor_id = "${openstack_compute_flavor_v2.flavor_1.id}"
```

So that it runs after the resource has been read.

kayrus: Done, thanks.

PR #591, opened 2019-01-11
Networking: minor fixes for the port data source

kayrus:
Part of the #512 

A couple of fixes/improvements

### DIFF
--- a/openstack/data_source_openstack_networking_port_v2.go
+++ b/openstack/data_source_openstack_networking_port_v2.go
@@ -24,66 +24,81 @@ func dataSourceNetworkingPortV2() *schema.Resource {
 			"region": {
 				Type:     schema.TypeString,
 				Optional: true,
+				ForceNew: true,
 			},
 
 			"port_id": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:          schema.TypeString,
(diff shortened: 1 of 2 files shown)


kayrus:
recheck

kayrus:
@jtopjian ready for review

jtopjian:
@kayrus I see this as slightly different than #587 

With the compute flavor API, it is not possible to perform a List using a flavor ID. Therefore, it is mutually exclusive to the other resources.

However, with networking resources, it is possible to do a List query with both the ID and Name. We already allow this for the network data source: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_networking_network_v2.go#L28-L35

I understand it might seems silly to query both, but maybe someone wants to make extra sure that the result needs to match both? The important thing is the API allows it.

I propose not doing `ConflictsWith` here.

kayrus:
@jtopjian are you ok just to leave the typo fix in the documentation?

jtopjian:
> are you ok just to leave the typo fix in the documentation?

That's something small enough that I can fix directly through Github - no PR needed.

kayrus:
ok, then let's close.

PR #592, opened 2019-01-11
Internal: Add ID to CheckDeleted

jtopjian:
This commit adds the resource ID to all CheckDeleted error messages for
easier detection of what resource the error came from.

It also adds CheckDeleted to the Delete function of missing resources
which have already gone through the cleanup process.

/cc @kayrus


jtopjian:
Thanks!

PR #593, opened 2019-01-12
Networking v2: Fix boolean fields of network resource

jtopjian:
openstack_networking_network_v2 was originally created with the fields
"admin_state_up" and "shared" as TypeString. This was to work around
Terraform mistaking an omitted boolean argument as "false".

Now that `GetOkExists` is available, we can correctly account for this
case.

This commit changes these two fields to TypeBool while still retaining
the prior behavior that TypeString was working around.

For #548

### DIFF
--- a/openstack/resource_openstack_networking_network_v2.go
+++ b/openstack/resource_openstack_networking_network_v2.go
@@ -3,7 +3,6 @@ package openstack
 import (
 	"fmt"
 	"log"
-	"strconv"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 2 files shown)


jtopjian:
@ozerovandrei I can rebase this after #578 is merged.

I included an extensive test suite for this, but felt it was a good idea to check all conditions. In addition, I ran this same test suite on master and the tests all passed - so that's a really good indication that this is not a breaking change.

This is because Terraform does an internal string-to-bool conversion on the user's behalf, so `"true"` works both as a string and boolean.

kayrus:
[Here](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_networking_network_v2.go#L65..L66) as well?

Issue #588, opened 2019-01-11
Networking: ports are created with admin_state_up set to false


rfc1459:
### Terraform Version
```
$ terraform -v
Terraform v0.11.11
+ provider.openstack v1.13.0
```

### Affected Resource(s)
- openstack_networking_port_v2

### Terraform Configuration Files
```hcl
terraform {
  backend "local" {}
}

provider "openstack" {
  version = "~> 1.13"
}

data "openstack_networking_secgroup_v2" "default_secgroup" {
  name = "default"
}

resource "openstack_networking_network_v2" "net1" {
  name = "net1"
}

resource "openstack_networking_subnet_v2" "subnet1" {
  name       = "subnet1"
  network_id = "${openstack_networking_network_v2.net1.id}"
  cidr       = "192.168.0.0/24"
  ip_version = 4
}

resource "openstack_networking_port_v2" "port1" {
  name       = "port1"
  network_id = "${openstack_networking_network_v2.net1.id}"

  fixed_ip {
    subnet_id = "${openstack_networking_subnet_v2.subnet1.id}"
  }

  security_group_ids = [
    "${data.openstack_networking_secgroup_v2.default_secgroup.id}"
  ]
}

output "port1_admin_state_up" {
  value = "${openstack_networking_port_v2.port1.admin_state_up}"
}
```

### Debug Output
https://gist.github.com/rfc1459/974b44498fcf39c772f66ed683fc3c59

### Expected Behavior
Port should have been created with `admin_state_up` set to `true`.

### Actual Behavior
Port is created with `admin_state_up` set to `false`.

### Steps to Reproduce
1. `terraform init`
2. `terraform apply -auto-approve`

### Important Factoids
In [`resourcePortAdminStateUpV2`](https://github.com/terraform-providers/terraform-provider-openstack/blob/7154fd2187e6c6b1e418fd17ca3fc63d3a2b1463/openstack/resource_openstack_networking_port_v2.go#L531) the default value for `admin_state_up` is `false` unless explicitly specified in the resource. This is a significant deviation from the default behaviour of the [Neutron Port API](https://developer.openstack.org/api-ref/network/v2/index.html?expanded=create-port-detail#create-port) which specifies the default value as `true` for newly created ports.

Also, the provider documentation lists the attribute as optional and does not state that its default value is different from the one of the Neutron API.

### References
None

jtopjian:
@rfc1459 Thank you for reporting this.

[...]

PR #594, opened 2019-01-12
Networking v2: Fix Port admin_state_up behavior

jtopjian:
If "admin_state_up" is not specified in the API request, then the port
should default to an admin_state_up of "UP".

Currently, this resource will always send a value of "false", even if
the user did not specify it in their resource configuration.

This commit fixes this behavior.

For #588

### DIFF
--- a/openstack/resource_openstack_networking_port_v2.go
+++ b/openstack/resource_openstack_networking_port_v2.go
@@ -204,7 +204,6 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
 		ports.CreateOpts{
 			Name:                d.Get("name").(string),
 			Description:         d.Get("description").(string),
-			AdminStateUp:        resourcePortAdminStateUpV2(d),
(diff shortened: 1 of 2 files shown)


jtopjian:
This PR will change the behavior of this resource, but surprisingly this is OK:

https://www.terraform.io/docs/extend/best-practices/versioning.html#example-patch-number-increments

because this resource is not conforming to the API (when a user does not specify `admin_state_up`, the port should be UP).

I think this should be OK because everyone should already be explicitly setting `admin_state_up` in their resource configurations already (since you can't do anything with a port that is not UP).

I am always skittish when making changes to the port resource, but I think this is for the benefit of the user. I will field any issues that come from this.

jtopjian:
it looks like there are a number of other networking resources that need this change, too. If we merge this, I suggest we use this as a canary. If no issues were reported, we can do a bulk bug fix of the other fields.

jtopjian:
All good - thank you!

PR #595, opened 2019-01-12
Internal: Networking V2 Router interface cleanup

ozerovandrei:
Cleanup error and debug messages.
Use resourceNetworkingRouterInterfaceV2StateRefreshFunc() for state
checking.

### DIFF
--- a/openstack/networking_router_interface_v2.go
+++ b/openstack/networking_router_interface_v2.go
@@ -1,9 +1,13 @@
 package openstack
 
 import (
+	"log"
+
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
 )
 
 func resourceNetworkingRouterInterfaceV2StateRefreshFunc(networkingClient *gophercloud.ServiceClient, portID string) resource.StateRefreshFunc {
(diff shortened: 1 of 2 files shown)


jtopjian:
recheck

jtopjian:
@ozerovandrei It looks like the test failures here are legitimate.

Good catch on line 91. But it seems that is where the problem is coming from: the tests are showing that the port is in a `DOWN` state when we are only checking for `ACTIVE` (note: this is different from `admin_state_up` - it means that the port exists, but nothing is active on it yet). So we need to change:

```go
Target:     []string{"ACTIVE"},
// to
Target:     []string{"ACTIVE", "DOWN"},
```

I think we need to restore the `waitForRouterInterfaceDelete` function because it is handling a 409 case: https://github.com/terraform-providers/terraform-provider-openstack/commit/ed66773e

And whenever a 409 is involved, that means we need to retry the whole process again, hence embedding the `RemoveInterface` in the refreshfunc.

Embedding the `ports.Get` into the refreshfunc is also necessary, but it's not easy to understand why. Thankfully, someone added some really good documentation to Gophercloud: https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/layer3/routers/requests.go#L158-L178

```
// The PortID that is returned after using Extract() on the result of this
// operation can either be the same PortID passed in or, on the other hand, the
// identifier of a new port created by this operation. 
```

What this means is that if someone attaches an interface by a subnet, a new port is created. Because `port_id` in this resource is set to `Computed`, the value is set to this port (and why we can't use `ConflictsWith`). So when you go to delete the interface, _sometimes_ a port will be deleted.

If a user uses a `port_id` instead of a `subnet_id`, the port will never delete, so we wait until `routers.RemoveInterface` returns a 404.

ozerovandrei:
@jtopjian wow, that's interesting. I'll update this PR.

jtopjian:
@ozerovandrei Thanks!

We still need to make the change on line 83 to be:

```go
Target:     []string{"ACTIVE", "DOWN"},
```

The reason is because you (correctly) added an err check on line 91. The error that is happening in OpenLab (and causing the tests to fail) has _always_ happened - it's just that we were never checking for it :slightly_smiling_face:

PR #599, opened 2019-01-14
LBaaSv2: add CheckDeleted into the Delete functions

kayrus:
For #456

### DIFF
--- a/openstack/resource_openstack_lb_l7policy_v2.go
+++ b/openstack/resource_openstack_lb_l7policy_v2.go
@@ -343,13 +343,13 @@ func resourceL7PolicyV2Delete(d *schema.ResourceData, meta interface{}) error {
 	// Get a clean copy of the listener.
 	listener, err := listeners.Get(lbClient, listenerID).Extract()
 	if err != nil {
-		return fmt.Errorf("Unable to retrieve listener %s: %s", listenerID, err)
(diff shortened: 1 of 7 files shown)


kayrus:
@jtopjian ready for review

openstack/resource_openstack_lb_l7policy_v2.go
```diff
@@ -343,13 +343,13 @@ func resourceL7PolicyV2Delete(d *schema.ResourceData, meta interface{}) error {
 	// Get a clean copy of the listener.
 	listener, err := listeners.Get(lbClient, listenerID).Extract()
 	if err != nil {
-		return fmt.Errorf("Unable to retrieve listener %s: %s", listenerID, err)
+		return CheckDeleted(d, err, fmt.Sprintf("Unable to retrieve parent listener (%s) for the L7 Policy", listenerID))
```
jtopjian: I don't think this is a good idea.

I understand that a listener and l7policy are tightly coupled and that _most_ of the time, if the listener doesn't exist then it's very probable that the l7policy also does not exist.

But I think we should only shortcut the removal of the resource from the Terraform state if the resource itself doesn't exist - not if one of its parents are missing.

kayrus: Ack.

PR #600, opened 2019-01-14
Manila: add CheckDeleted into the Delete functions

kayrus:
For #456


kayrus:
@jtopjian ready for review

PR #601, opened 2019-01-15
Acceptance tests: panic on schema error

kayrus:
Honeypot for invalid schema types.


kayrus:
@jtopjian ready for review. Do I need to split this PR apart?

jtopjian:
`TF_SCHEMA_PANIC_ON_ERROR` - I didn't know this was available. We will need to work through the errors and, once all fixed, I can open a request in OpenLab.

This will need split into separate PRs, though, because each change will need to have a changelog entry (thus PR).

Going through the list of changes:

1. We cannot update the makefile - this is something managed by HashiCorp. Most likely the reason this isn't part of the standard makefile template is because tests would fail :slightly_smiling_face: Having it added as an env var in OpenLab (once everything is fixed) is the best we can do to standardize this.

2. `data_source_openstack_images_image_v2.go`: The changes look legitimate, but definitely a separate PR. 

3. `data_source_openstack_networking_subnet_v2.go`: Same

4. `data_source_openstack_networking_subnetpool_v2.go`: Same

5. `resource_openstack_images_image_v2.go`: We cannot rename the field, even if it's technically not working. It's possible someone has used this field somewhere (again, even though it never gets a value). We need to follow the process outlined here: https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-a-computed-attribute

6. `resource_openstack_networking_network_v2.go`: There's already a PR for this.

7. `resource_openstack_networking_subnet_v2.go`: Same note about separate PR.

8. `resource_openstack_networking_subnetpool_v2.go`: Same

The work here is definitely appreciated - thank you for taking the time to do this. Please let me know if you have any questions.

kayrus:
Thank you for your feedback.

> We cannot update the makefile

If you take a look on other providers, you can see this ENV var, e.g. [azure](https://github.com/terraform-providers/terraform-provider-azurerm/blob/55068f02b313e955e4c97ac510c00d08c487edab/GNUmakefile#L6).

> resource_openstack_networking_network_v2.go: There's already a PR for this.

I took your commit from the #593 PR

jtopjian:
> If you take a look on other providers, you can see this ENV var, e.g. azure.

But it's not in AWS or MySQL (two I randomly spot-checked). I can inquire about this later, but this isn't something I feel is a priority to have added. We have a lot of other work going on right now.

kayrus:
@jtopjian ok, closing this.

PR #603, opened 2019-01-15
Manila: fix doc captions

kayrus:
`openstack_` caption prefix was forgotten. Fortunately the links are ok.


jtopjian:
>  Fortunately the links are ok.

As an aside, there is a Travis test that runs to validate that all links work.

PR #608, opened 2019-01-15
Internal: use terraform validation for slices

kayrus:
Code cleanup #548


jtopjian:
@kayrus These changes are legitimate and appreciated, but it would be best to work through them doing whole resources instead of piecemeal changes for each resource. This allows the person doing the clean up to not have to keep in mind that some changes have been done, possibly making them forget to check for other changes.

Of course, human error occurs and changes can be missed, but doing the clean up by individual resources helps keep changes localized and not spread.

I have the identity suite done - I just need to find the time to open a PR after fielding reviews :wink:

PR #609, opened 2019-01-16
Internal: identity auth scope cleanup

jtopjian:
For #456

### DIFF
--- a/openstack/identity_auth_scope_v3.go
+++ b/openstack/identity_auth_scope_v3.go
@@ -5,13 +5,13 @@ import (
 )
 
 func flattenIdentityAuthScopeV3Roles(roles []tokens.Role) []map[string]string {
-	var allRoles []map[string]string
+	allRoles := make([]map[string]string, len(roles))
 
-	for _, r := range roles {
-		allRoles = append(allRoles, map[string]string{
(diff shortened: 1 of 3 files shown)


openstack/identity_auth_scope_v3.go
```diff
@@ -0,0 +1,19 @@
+package openstack
+
+import (
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+)
+
+func flattenIdentityAuthScopeV3Roles(roles []tokens.Role) []map[string]string {
+	var allRoles []map[string]string
+
+	for _, r := range roles {
```
ozerovandrei: We can pre-create allRoles with len of roles:
```
allRoles := make([]map[string]string, len(roles))
```

And then populate them by index:
```
for i, r := range roles {
  allRoles[i] = map[string]string{
    ...
  }
}
```

This will help with memory usage (by not using the append), but it will be helpful only in case of big amount of roles in API response.

jtopjian: Agreed.

jtopjian:
recheck

jtopjian:
recheck

PR #613, opened 2019-01-16
Internal: identity roles cleanup

jtopjian:
For #456

### DIFF
--- a/openstack/identity_role_assignment_v3.go
+++ b/openstack/identity_role_assignment_v3.go
@@ -1,0 +1,54 @@
+package openstack
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Role assignments have no ID in OpenStack.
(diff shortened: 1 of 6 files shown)


openstack/identity_role_assignment_v3.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Role assignments have no ID in OpenStack.
+// Build an ID out of the IDs that make up the role assignment
+func identityRoleAssignmentV3ID(domainID, projectID, groupID, userID, roleID string) string {
+	return fmt.Sprintf("%s/%s/%s/%s/%s", domainID, projectID, groupID, userID, roleID)
+}
+
+func identityRoleAssignmentV3ParseID(roleAssignmentID string) (string, string, string, string, string) {
+	split := strings.Split(roleAssignmentID, "/")
+	return split[0], split[1], split[2], split[3], split[4]
```
ozerovandrei: Maybe we should check that len(split) == 5 prior retrieving slice elements by index? It may panic in case of bad role assignment ID.

jtopjian: Yeah, I've been trying to be cautious and not introducing new logic into these cleanups, but I think this is safe to do.

jtopjian:
recheck

jtopjian:
recheck

PR #615, opened 2019-01-16
Internal: fix schema set for the image data source

kayrus:
Refs to #601
For #456


kayrus:
@jtopjian ready for review. test failed on the non-related `TestAccComputeV2Instance_crazyNICs`

PR #617, opened 2019-01-16
Internal: fix schema set for the image resource

kayrus:
Refs to #601
For #456


kayrus:
@jtopjian ready for review

openstack/resource_openstack_images_image_v2.go
```diff
@@ -176,10 +176,16 @@ func resourceImagesImageV2() *schema.Resource {
 				Computed: true,
 			},
 
+			// Deprecated "use updated_at instead"
```
jtopjian: Instead of a comment, do:

```go
			"update_at": {
				Type:     schema.TypeString,
				Computed: true,
                                Deprecated: "Use updated_at instead",
			},
```

website/docs/r/images_image_v2.html.markdown
```diff
@@ -106,7 +106,8 @@ The following attributes are exported:
 * `status` - The status of the image. It can be "queued", "active"
    or "saving".
 * `tags` - See Argument Reference above.
-* `update_at` - The date the image was last updated.
+* `updated_at` - The date the image was last updated.
+* `update_at` - (Deprecated - use `updated_at` instead)
```
jtopjian: This should be `**Deprecated**`

jtopjian:
This looks like it might need rebased from master - the networking boolean changes got mixed up in here. I'm not sure if I missed that yesterday or if it just happened.

kayrus:
Rebased

PR #618, opened 2019-01-16
Internal: fix schema set for the subnet resource

kayrus:
Refs to #601
For #456

### DIFF
--- a/openstack/resource_openstack_networking_subnet_v2.go
+++ b/openstack/resource_openstack_networking_subnet_v2.go
@@ -115,6 +115,7 @@ func resourceNetworkingSubnetV2() *schema.Resource {
 				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: false,
+				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"destination_cidr": {
@@ -266,7 +267,7 @@ func resourceNetworkingSubnetV2Read(d *schema.ResourceData, meta interface{}) er
 	d.Set("description", s.Description)
 	d.Set("tenant_id", s.TenantID)
 	d.Set("dns_nameservers", s.DNSNameservers)
-	d.Set("host_routes", s.HostRoutes)
+	d.Set("host_routes", resourceSubnetHostRoutesToMapV2(s.HostRoutes))
 	d.Set("enable_dhcp", s.EnableDHCP)
 	d.Set("network_id", s.NetworkID)
 	d.Set("ipv6_address_mode", s.IPv6AddressMode)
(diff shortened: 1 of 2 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
@kayrus Is it possible to include an acceptance test that validates this?

kayrus:
@jtopjian it is possible :)

kayrus:
recheck lbaas

kayrus:
recheck fwaas

kayrus:
recheck vpnaas

kayrus:
recheck designate

kayrus:
recheck trove

kayrus:
recheck lbaas

kayrus:
recheck fwaas

kayrus:
recheck designate

kayrus:
recheck trove

kayrus:
@jtopjian any update here?

kayrus:
@jtopjian looks like I opened yet another Pandora box.

Your branch looks fine. Only one addition: does it make sense to introduce a `DiffSuppressFunc` function in your implementation as well? If I'm not mistaken, it can help to handle ordering issue:

```go
DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
        // Only work with existing instances
        if d.Id() == "" {
                return false
        }
        // Only work with missing configurations
        if new != "" {
                ... ... ...
        }
        // Only work when already set in Terraform state
        if old == "" {
                ... ... ...
        }
        return true
},
```

Can we detect the ordering inside the `DiffSuppressFunc` for TypeSets?

jtopjian:
> Only one addition: does it make sense to introduce a DiffSuppressFunc function in your implementation as well? If I'm not mistaken, it can help to handle ordering issue:

It's possible, but I'm hesitant to do it. I looked into this for the DNS recordset solution and there was something about it that made me not want to do it - I need to revisit it. I hope to be able to get back to that shortly.

jtopjian:
`DiffSuppressFunc` and the other functions that handle state manipulation actually write the modified state. So if a user specified an order as 1, 2, 3 and the returned order was 2, 3, 1, then `DiffSuppressFunc` _could_ say there is no diff, but it would write 2, 3, 1 to state. I wish there was an ability to not write to state.

After reviewing everything after some days, I still think the fix I've done in my branch is good to go. It's essentially the same thing as the DNS stuff, but without any bracket manipulation.

Unless there are objections, I've opened #645 and will use that as the solution.

kayrus:
I'll take another look on the #645 within this week

kayrus:
Closing in favor of #668

PR #619, opened 2019-01-16
Internal: fix schema set for the subnetpool resource

kayrus:
Refs to #601
For #456


kayrus:
@jtopjian ready for review

PR #626, opened 2019-01-17
Networking V2: Add trunk data source

### LINKED ISSUE #622: Networking V2: add trunk data source
Trunk resource was added in #446. It would be great to have a data source as well:

- openstack_networking_trunk_v2

kayrus:
Resolves #622

### DIFF
--- a/openstack/util.go
+++ b/openstack/util.go
@@ -1,7 +1,6 @@
 package openstack
 
 import (
-	"bytes"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -13,7 +12,6 @@ import (
 	"github.com/Unknwon/com"
 	"github.com/gophercloud/gophercloud"
 	"github.com/hashicorp/terraform/flatmap"
-	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
@@ -276,41 +274,4 @@ func sliceUnion(a, b []string) []string {
 		}
 	}
 	return res
-}
-
-// serializeValueForHash appends a serialization of the given resource config
-// to the given buffer, guaranteeing deterministic results given the same value
-// and schema.
-//
-// It is copied from the schema.SerializeResourceForHash, but with a minor changes
-// to perform a proper serialization of the computed attributes
-func serializeResourceForHash(buf *bytes.Buffer, val interface{}, resource *schema.Resource) {
-	if val == nil {
-		return
-	}
-	sm := resource.Schema
-	m := val.(map[string]interface{})
-	var keys []string
-	for k := range sm {
-		keys = append(keys, k)
-	}
-	sort.Strings(keys)
-	for _, k := range keys {
-		innerSchema := sm[k]
-		buf.WriteString(k)
-		buf.WriteRune(':')
-		innerVal := m[k]
-		schema.SerializeValueForHash(buf, innerVal, innerSchema)
-	}
-}
-
-func resourceAttributesHash(parentSchema func() *schema.Resource, key string) func(interface{}) int {
-	return func(v interface{}) int {
-		m := v.(map[string]interface{})
-		resource := parentSchema().Schema[key].Elem.(*schema.Resource)
-
-		var buf bytes.Buffer
-		serializeResourceForHash(&buf, m, resource)
-		return hashcode.String(buf.String())
-	}
 }
--- a/openstack/util.go
+++ b/openstack/util.go
@@ -1,6 +1,7 @@
 package openstack
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -12,6 +13,7 @@ import (
 	"github.com/Unknwon/com"
 	"github.com/gophercloud/gophercloud"
 	"github.com/hashicorp/terraform/flatmap"
+	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
@@ -274,4 +276,41 @@ func sliceUnion(a, b []string) []string {
 		}
 	}
 	return res
+}
+
+// serializeValueForHash appends a serialization of the given resource config
+// to the given buffer, guaranteeing deterministic results given the same value
+// and schema.
+//
+// It is copied from the schema.SerializeResourceForHash, but with a minor changes
+// to perform a proper serialization of the computed attributes
+func serializeResourceForHash(buf *bytes.Buffer, val interface{}, resource *schema.Resource) {
+	if val == nil {
+		return
+	}
+	sm := resource.Schema
+	m := val.(map[string]interface{})
+	var keys []string
+	for k := range sm {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		innerSchema := sm[k]
+		buf.WriteString(k)
+		buf.WriteRune(':')
+		innerVal := m[k]
+		schema.SerializeValueForHash(buf, innerVal, innerSchema)
+	}
+}
+
+func resourceAttributesHash(parentSchema func() *schema.Resource, key string) func(interface{}) int {
+	return func(v interface{}) int {
+		m := v.(map[string]interface{})
+		resource := parentSchema().Schema[key].Elem.(*schema.Resource)
+
+		var buf bytes.Buffer
+		serializeResourceForHash(&buf, m, resource)
+		return hashcode.String(buf.String())
+	}
 }
--- a/openstack/data_source_openstack_networking_trunk_v2.go
+++ b/openstack/data_source_openstack_networking_trunk_v2.go
@@ -167,7 +167,7 @@ func dataSourceNetworkingTrunkV2Read(d *schema.ResourceData, meta interface{}) e
 	d.Set("port_id", trunk.PortID)
 	d.Set("admin_state_up", trunk.AdminStateUp)
 	d.Set("status", trunk.Status)
-	d.Set("tenant_id", trunk.TenantID)
+	d.Set("project_id", trunk.ProjectID)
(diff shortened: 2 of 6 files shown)


kayrus:
@jtopjian I identified the schema problem in the trunk resource. Now sure whether this commit is correct and whether it should be a dedicated PR https://github.com/terraform-providers/terraform-provider-openstack/pull/626/commits/25c88ace7d170b8583fd29fcb1955d5e90d9ca4b

jtopjian:
>  I identified the schema problem in the trunk resource. Now sure whether this commit is correct and whether it should be a dedicated PR 25c88ac

I spent the past two hours digging into this because something seemed odd.

I was very confident that `TypeSet` has a default hashing function applied to it and it's recommended to default to using it unless otherwise necessary. This is true.

What I learned is that data sources _do not_ have this function applied to it. I'm not sure if this is by design or a bug. I'll ask about this. But either way, we will have to supply a custom Set function for all `TypeSet` fields in data sources (cc @ozerovandrei). 

With that in mind, unfortunately we can't just apply a new hashing function to the resource. This will cause existing hashes to change and _might_ break existing deployments. Personally, I don't see anyone doing this because Set hash IDs are impossible to work with, but we need to consider it. If we _really_ wanted to proceed with applying the function to the resource, we need to do a schema migration. Take a look at `*_migrate* in the AWS provider for examples on how to do this.

Alternatively, a better way might be to dissect the way Terraform is calculating hashes by default and use that for the `sub_port` hash func. I took a quick run at this tonight and while things didn't blow up, I didn't get it correct on the first try.

The last option is to leave the hash IDs between the resource and data source different.

I'd like to see how possible it is to recreate the default hash ID as it would probably be useful elsewhere. Maybe the AWS or another major resource has done this already.

kayrus:
@jtopjian AWS provider has the same issues. I reported a bugreport there too: https://github.com/terraform-providers/terraform-provider-aws/issues/7198

jtopjian:
@kayrus Nice. That should provide some input on why this is happening. Even if it's a bug, though, it'll mean re-vendoring a future release of Terraform, which will probably be v0.12+. So it might be a while until it's fully resolved. However, it'll be good to know anyway.

jtopjian:
Got it:

```go
func resourceNetworkingTrunkV2SubPortHash(v interface{}) int {
  var buf bytes.Buffer
  m := v.(map[string]interface{})
  var keys []string
  for k := range m {
    keys = append(keys, k)
  }
  sort.Sort(sort.StringSlice(keys))

  for _, k := range keys {
    buf.WriteString(k)
    buf.WriteRune(':')
    val := m[k]
    switch val.(type) {
    case string:
      buf.WriteString(val.(string))
    case int:
      buf.WriteString(strconv.Itoa(val.(int)))
    }
    buf.WriteRune(';')
  }

  return hashcode.String(buf.String())
}
```

Adding a few other types should allow this to become a general util function for situations where there is a "flat" `TypeSet` in data sources. This would not support nested sets, but I don't think we have any of those.

kayrus:
@jtopjian I took a closer look into this issue and figured out that the `Set` function is not required, when you specify `Optional: true` for all subattributes in data source.

kayrus:
Here is the terraform code, responsible for this: https://github.com/hashicorp/terraform/blob/ac4fff416318bf0915a0ab80e062a99ef3724334/helper/schema/serialize.go#L103

jtopjian:
> I took a closer look into this issue and figured out that the Set function is not required, when you specify Optional: true for all subattributes in data source.

🤔 

Yes, though technically those attributes aren't optional (users cannot set them at all due to the Set itself being computed). I'm wondering if there's a case when this won't work in all situations?

What's wrong with keeping the schema defined correctly and using the hash function I pasted above?

kayrus:
> What's wrong with keeping the schema defined correctly and using the hash function I pasted above?

Well, I think it is better not to introduce new entities, when you can avoid this. In existing situation we can definitely avoid new entities.

What exactly alarms you of using `Optional` attribute when the parent set is `Computed`?

jtopjian:
> What exactly alarms you of using Optional attribute when the parent set is Computed?

It's not a true representation of the schema and it may introduce bugs or other problems later on. It's a workaround to trigger an area of the code Terraform core that we have no easy control over.

We know we need to generate a hash - your original implementation did this. So why not make the hash calculation match what the resource is doing without requiring a schema migration or a breaking change? In addition, this kind of calculation can be used for other data sources with similar schemas.

kayrus:
@jtopjian your words make sense, however I disagree that we need to introduce the hashing function from scratch. What if we take [SerializeResourceForHash](https://github.com/hashicorp/terraform/blob/ac4fff416318bf0915a0ab80e062a99ef3724334/helper/schema/serialize.go#L87) as a base, remove `if !(innerSchema.Required || innerSchema.Optional) {` condition and reuse the exported [SerializeValueForHash](https://github.com/hashicorp/terraform/blob/ac4fff416318bf0915a0ab80e062a99ef3724334/helper/schema/serialize.go#L10) ? It already supports "a few other types" and nested sets.

jtopjian:
I tried to find some way of calling the existing functions, but I don't see an easy way. For example, the function signature for `SerializeResourceForHash` is:

```go
SerializeResourceForHash(buf *bytes.Buffer, val interface{}, resource *Resource)
```

`*Resource` isn't being passed to the `Set` function, which means we would need to build it from scratch within the `Set` function.

If you can find a way to do that, it's worth looking at as a solution.

jtopjian:
I figured out how to trigger the default hashing logic:

```go
func resourceNetworkingTrunkV2SubPortHash(v interface{}) int {
       m := v.(map[string]interface{})
       subPortResource := dataSourceNetworkingTrunkV2().Schema["sub_port"].Elem.(*schema.Resource)

       f := schema.HashResource(subPortResource)
       return f(m)
}
```

This is being used in a few AWS tests: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_waf_rule_test.go#L192-L209

kayrus:
As I expected, it uses the same `SerializeResourceForHash` function:

```golang
func HashResource(resource *Resource) SchemaSetFunc {
        return func(v interface{}) int {
                var buf bytes.Buffer
                SerializeResourceForHash(&buf, v, resource)
                return hashcode.String(buf.String())
        }
}
```

I'll take a closer look on what exactly is happening after the `schema.Resource` assertion.

jtopjian:
> As I expected, it uses the same SerializeResourceForHash function:

Right. The difficult part was trying to figure out how to dig into the schema to get the `TypeSet` in the right type cast.

kayrus:
I created a Frankenstein function, but it works and it could be used anywhere.

openstack/util.go
```diff
@@ -239,3 +241,40 @@ func strSliceContains(haystack []string, needle string) bool {
...
+	for _, k := range keys {
+		innerSchema := sm[k]
+		buf.WriteString(k)
+		buf.WriteRune(':')
+		innerVal := m[k]
+		schema.SerializeValueForHash(buf, innerVal, innerSchema)
+	}
+}
+
+func resourceAttributesHash(parentSchema func() *schema.Resource, key string) func(interface{}) int {
```
jtopjian: This resource works for exactly like you defined it: a utility function to re-use. But we can now safely re-use `schema.HashResource`:

```go
func resourceAttributesHash(parentSchema func() *schema.Resource, key string) func(interface{}) int {
	return func(v interface{}) int {
		m := v.(map[string]interface{})
		resource := parentSchema().Schema[key].Elem.(*schema.Resource)
                f := schema.HashResource(resource)
                return f(m)
        }
}
```

We might be able to tighten it up even more with:

```go

...
  Set: resourceAttributesHash(dataSourceNetworkingTrunkV2(), "sub_port"),
...

func resourceAttributesHash(r *schema.Resource, key string) schema.SchemaSetFunc {
	return func(v interface{}) int {
		m := v.(map[string]interface{})
		resource := r.Schema[key].Elem.(*schema.Resource)
                f := schema.HashResource(resource)
                return f(m)
        }
}
```

(not tested)

The reason I originally copied the hashcode algorithm is because I couldn't find a way to trigger it in Terraform core naturally. Using schema workarounds wasn't a safe thing to do (it was making a data source schema look like it wasn't a data source which could be dangerous), so the only solution was to reproduce the code. However, now that we can call the code without a schema hack, we no longer need to duplicate the hashing code.  

Thoughts?

jtopjian: 🤦‍♂️ 

That's what I get for working while I should be listening to conference talks. I was working on a branch which had `Optional: true` set. You're absolutely correct.

I think this all looks good then. 

Any final comments / thoughts? Are you happy with this solution?

jtopjian: If you'd like, we can always wait until the AWS issue is resolved and then match the implementation.

kayrus: This PR was a brains workout, and there is no hurry.
BTW, Openstack provider community looks much more supportive comparing to AWS one :)

kayrus:
@jtopjian I found yet two places, where the hash bug probability is high:

* [openstack_blockstorage_volume_v3](https://github.com/terraform-providers/terraform-provider-openstack/blob/946422d2faa7c6436cc6291b19d0e67f7f7f1e3d/openstack/resource_openstack_blockstorage_volume_v3.go#L118..L137). if there are multiple attachments of the same block storage on one instance (not sure, whether it is possible), but different attach IDs, hashes will overlap, since hash is calculated based on the instance id only.

* extra_dhcp_option in [data.openstack_networking_port_v2](https://github.com/terraform-providers/terraform-provider-openstack/blob/946422d2faa7c6436cc6291b19d0e67f7f7f1e3d/openstack/data_source_openstack_networking_port_v2.go#L142..L160). all fields are computed. if there are more than one `extra_dhcp_option`, hashes will overlap.

kayrus:
@jtopjian do you also suggest to use `TypeList` here as well?

jtopjian:
Yes, let's do a `TypeList` here, too.

kayrus:
recheck

kayrus:
@jtopjian ready

jtopjian:
For posterity: I see that `all_tags` is a `TypeSet`. In theory, we could have used `TypeList` for this, per previous discussions, but given `all_tags` does not exhibit the noted problem _and_ since it's already widely deployed in other data sources, we'll keep it at `TypeSet`.

PR #628, opened 2019-01-17
Networking V2: Omit empty ExternalFixedIP IPAddress for the router

### LINKED ISSUE #449: How does ip_address in external_fixed_ip argument of openstack_networking_router_v2 get calculated?
Hi there,

I'd like to ask how the  ip_address gets calculated (https://www.terraform.io/docs/providers/openstack/r/networking_router_v2.html#ip_address) for openstack_networking_router_v2 resource? The reason for asking is the following use case:

My openstack project has an external network which is divided into two subnets. 

The default configuration for router looks as follows:

```hcl
resource "openstack_networking_router_v2" "terrakube_router" {
  name                = "${var.k8s_cluster_name}_router"
  admin_state_up      = true
  external_network_id = "${data.openstack_networking_network_v2.terrakube_external_network.id}"
}
```

When I create a router like this and specify only the  `external_network_id`, the first one of the subnets is being attached successfully. If I try to attach the second one instead, I specify the `external_fixed_ip` argument as follows:

```hcl
resource "openstack_networking_router_v2" "terrakube_router" {
  name                = "${var.k8s_cluster_name}_router"
  admin_state_up      = true
  external_network_id = "${data.openstack_networking_network_v2.terrakube_external_network.id}"

  external_fixed_ip {
    subnet_id = "${data.openstack_networking_subnet_v2.terrakube_external_subnet.id}"
  }
}
```
This code causes an error:

```
* module.terrakube.openstack_networking_router_v2.terrakube_router: 1 error(s) occurred:                                                  
                                                                                                                                         
* openstack_networking_router_v2.terrakube_router: Error creating OpenStack Neutron router: Bad request with: [POST https://network-3.eu-de-2.cloud.sap/v2.0/r
outers], error message: {"NeutronError": {"message": "Invalid input for external_gateway_info. Reason: '' is not a valid IP address.", "type": "HTTPBadRequest
", "detail": ""}}
``` 

I could hardcode the value of the IP but that seems to be an ugly solution, so I would like to find out how to get it via terraform data sources or using some other approach?

### Terraform Version
Terraform v0.11.8
+ provider.ccloud (unversioned)
+ provider.openstack v1.10.0

Thanks!

kayrus:
Resolves #449
Requires https://github.com/gophercloud/gophercloud/issues/1414 to be fixed

### DIFF
--- a/openstack/resource_openstack_networking_router_v2.go
+++ b/openstack/resource_openstack_networking_router_v2.go
@@ -92,6 +92,7 @@ func resourceNetworkingRouterV2() *schema.Resource {
 						"ip_address": {
 							Type:     schema.TypeString,
 							Optional: true,
+							Computed: true,
 						},
 					},
 				},
--- a/openstack/resource_openstack_networking_router_v2_test.go
+++ b/openstack/resource_openstack_networking_router_v2_test.go
@@ -107,6 +107,30 @@ func TestAccNetworkingV2Router_vendor_opts_no_snat(t *testing.T) {
 	})
 }
 
+func TestAccNetworkingV2Router_extFixedIPs(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckAdminOnly(t)
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckNetworkingV2RouterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNetworkingV2Router_extFixedIPs,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"openstack_networking_router_v2.router_2", "name", "router_2"),
+					resource.TestCheckResourceAttr(
+						"openstack_networking_router_v2.router_2", "external_fixed_ip.#", "2"),
+					resource.TestCheckResourceAttr(
+						"openstack_networking_router_v2.router_2", "enable_snat", "true"),
+				),
+			},
+		},
+	})
+}
+
 func testAccCheckNetworkingV2RouterDestroy(s *terraform.State) error {
 	config := testAccProvider.Meta().(*Config)
 	networkingClient, err := config.networkingV2Client(OS_REGION_NAME)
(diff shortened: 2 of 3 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
@kayrus I feel like I'm going crazy.

Step 2 of the acceptance test isn't showing that the `external_fixed_ip` has an update, only `enable_snat`. Are you seeing the same thing?

Even adding logging statements in the `d.HasChange("external_fixed_ip")` block isn't showing the logs being made.

kayrus:
Good catch, I was wondering why these tests has been passed, because I wrote them blindfold, since my environment is not capable to update routes. For some reason these tests are skipped in openlab env, because there is no admin username. I removed these checks, as far as I know it should not be a requirement.

> Even adding logging statements in the d.HasChange("external_fixed_ip") block isn't showing the logs being made.

Looks like it is impossible to remove default `external_fixed_ip`... Besides this field marked as `Computed` therefore omitting it won't change the plan. Setting an empty list also doesn't give any effects. I've pushed an update to remove the 3rd ext IP. Could you please confirm whether it works for you?

kayrus:
What happened? I didn't close this ticket? And why github marks it is merged?
screenshot_20190226_141911

kayrus:
@jtopjian can you run acceptance tests against your environment?

kayrus:
@jtopjian I managed to run tests in my env. And I have some bad news:

1) TypeList considers ordering, and when we have more than one `external_fixed_ip`, terraform tries to update the router
2) We need to switch the `external_fixed_ip` to TypeMap with the hashing based on the `subnet_id`. If there is no custom hashing func, the computed `ip_address` will trigger the diff, and terraform will suggest to update the router
3) Custom hashing func based on the `subnet_id` will ignore any `ip_address` modifications made by user, update func won't work on  `ip_address` modification.

jtopjian:
@kayrus Thank you for the notes and for looking into this. I will review and see if I come to similar conclusions when I get some time. :)

jtopjian:
@kayrus I know I'm behind on this, but I have not forgotten about it.

jtopjian:
I'm sorry for the delay. I had some time to review this.

The problem you're describing already exists: if you declare multiple `external_fixed_ip` blocks and set the `subnet_id` and `ip_address` of both and if they are returned in a different order, there will be a diff.

I think there are two problem statements here:

1. Being able to specify an `external_fixed_ip` without an `ip_address`. This is what #449 is reporting.
2. Declaring multiple `external_fixed_ip` blocks. This problem already exists.

Since no one has reported 2 yet, I have to wonder if anyone is declaring multiple `external_fixed_ip` blocks to begin with -- is there a legitimate use for this? I'm not trying to say the problem isn't valid, but since this problem already exists, is there anything preventing us from fixing #449 and looking at this second issue separately?

If we were to look at this second problem, we would not change `external_fixed_ip` to a `TypeMap`. It's a clever idea, but it's too much of a breaking change. We'll need to either look into a [`CustomizeDiff`](https://www.terraform.io/docs/extend/resources.html#customizing-differences) or create a separate resource to handle a router's gateway information.

kayrus:
@ozerovandrei I think this is ready to be merged. Other related issues are addressed in #922, since there is no easy way to fix them.

ozerovandrei:
recheck

PR #631, opened 2019-01-17
Networking V2: Fix port data sources

### LINKED ISSUE #630: Networking V2: Fix port data sources
Affected data sources:
- openstack_networking_port_ids_v2

Bugs:
- `network_id` filter uses `status` list option
- `allowed_address_pairs` for openstack_networking_port_v2 doesn't have a `resourceNetworkingPortV2AllowedAddressPairsHash`
- empty ids list is not returned by the `openstack_networking_port_ids_v2` ([data.aws_ebs_snapshot_ids](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/data_source_aws_ebs_snapshot_ids.go) allows to return)

kayrus:
Resolves #630


jtopjian:
Just for clarity: re-using the hash func here is correct since we already have one for ports.

PR #632, opened 2019-01-20
Internal: Networking V2 Trunk resource cleanup

ozerovandrei:
Use extend and flatten functions for work with the trunk "sub_port"
argument/attribute.

Provide a single resource.StateRefreshFunc for the trunk resource.

Update debug messages for the trunk resource.

Fix "openstack_networking_trunk_v2" website documentation.

### DIFF
--- a/openstack/networking_trunk_v2.go
+++ b/openstack/networking_trunk_v2.go
@@ -1,0 +1,69 @@
+package openstack
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func networkingTrunkV2StateRefreshFunc(client *gophercloud.ServiceClient, trunkID string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		trunk, err := trunks.Get(client, trunkID).Extract()
+		if err != nil {
+			if _, ok := err.(gophercloud.ErrDefault404); ok {
+				return trunk, "DELETED", nil
+			}
+
+			return nil, "", err
+		}
+
+		return trunk, trunk.Status, nil
+	}
+}
+
+func flattenNetworkingTrunkSubportsV2(subports []trunks.Subport) []map[string]interface{} {
+	trunkSubports := make([]map[string]interface{}, len(subports))
+
+	for i, subport := range subports {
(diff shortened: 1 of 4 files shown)


openstack/networking_trunk_v2.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+
+			return nil, "", err
+		}
+
+		return trunk, trunk.Status, nil
+	}
+}
+
+func flattenNetworkingTrunkSubportsV2(subports []trunks.Subport) []map[string]interface{} {
```
jtopjian: nit with the naming: `flattenNetworkingTrunkV2Subports`

```
service-resource-version-attribute
```

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -181,62 +188,70 @@ func resourceNetworkingTrunkV2Update(d *schema.ResourceData, meta interface{}) e
...
-			subports := resourceNetworkingTrunkV2Subports(toRemove)
-			log.Printf("[DEBUG] Removing subports from trunk %s: %#v", d.Id(), subports)
-			for _, subport := range subports {
-				v := trunks.RemoveSubport{PortID: subport.PortID}
-				subportsToRemove.Subports = append(subportsToRemove.Subports, v)
+		o, n := d.GetChange("sub_port")
+		oldSubport := o.(*schema.Set)
+		newSubport := n.(*schema.Set)
+
+		// Delete all old subports, regardless of if they still exist.
```
jtopjian: I'm OK with this method, though we might get reports that users don't like the temporary disassociation of trunks that are never removed. We can keep this method and revert if that's an issue.

openstack/resource_openstack_networking_trunk_v2.go
```diff
@@ -181,62 +188,70 @@ func resourceNetworkingTrunkV2Update(d *schema.ResourceData, meta interface{}) e
...
 		}
 
-		if len(toAdd.List()) > 0 {
-			var subportsToAdd trunks.AddSubportsOpts
-			subportsToAdd.Subports = resourceNetworkingTrunkV2Subports(toAdd)
-
-			log.Printf("[DEBUG] Adding subports to trunk %s: %#v", d.Id(), subportsToAdd.Subports)
+		// Add any new sub_port and re-add previously set subports.
+		if newSubport.Len() != 0 {
+			addSubports := expandNetworkingTrunkSubportsV2(oldSubport)
```
jtopjian: `s/oldSubport/newSubport/`

Tests should pass after this.

PR #633, opened 2019-01-20
Internal: Networking V2 Subnetpool cleanup

ozerovandrei:
Provide a single resource.StateRefreshFunc for the subnetpool resource.

Update debug messages for the subnetpool resource.

Use expandToStringSlice() for the subnetpool "prefixes" argument.

### DIFF
--- a/openstack/resource_openstack_networking_subnetpool_v2.go
+++ b/openstack/resource_openstack_networking_subnetpool_v2.go
@@ -8,7 +8,6 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 
-	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/subnetpools"
 )
@@ -22,10 +21,12 @@ func resourceNetworkingSubnetPoolV2() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(10 * time.Minute),
 			Delete: schema.DefaultTimeout(10 * time.Minute),
 		},
+
 		Schema: map[string]*schema.Schema{
 			"region": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


openstack/resource_openstack_networking_subnetpool_v2.go
```diff
@@ -324,72 +348,23 @@ func resourceNetworkingSubnetPoolV2Delete(d *schema.ResourceData, meta interface
...
-			return s, "ACTIVE", err
-		}
-
-		err = subnetpools.Delete(networkingClient, subnetPoolID).ExtractErr()
-		if err != nil {
-			if _, ok := err.(gophercloud.ErrDefault404); ok {
-				log.Printf("[DEBUG] Successfully deleted OpenStack Subnetpool %s", subnetPoolID)
-				return s, "DELETED", nil
-			}
-			if errCode, ok := err.(gophercloud.ErrUnexpectedResponseCode); ok {
```
jtopjian: Do you think we should add this back to the refreshfunc? Is there a time when a pool will ever trigger a 409?

ozerovandrei: This looks strange. I think it was copied from another resource implementation.
Why do we need a 409 check here or in ipsec_policy_v2/fw_policy_v1/subnet_v2/secgroup_v2/router_interface_v2/lb_monitor_v1/objectstorage_container_v1 ?

jtopjian: 409s are triggered when a parent resource cannot be deleted until its children resources are deleted.

As an example, a subnet cannot be deleted until its ports are deleted. Additionally, a network cannot be deleted until ports are deleted, too.

Since ports do not require a subnet, when a user defines a network, subnet, and port, Terraform will try to delete the port and subnet at the same time. This triggers the 409.

So a 409 shouldn't be thought of as an "error" but as a way to detect and resolve weak relationships with OpenStack resources.

jtopjian: Also, 

> I think it was copied from another resource implementation.

That's always possible.

For subnet and subnet pool, the subnet resource has a `subnetpool_id` resource which creates the relationship (Terraform will first delete the subnet and then delete the subnet pool, thus not causing a 409).

So I'm _pretty_ sure we don't need the 409 here. My rule of thumb has always been to check for 409s (or allow them when someone submits them) just out of safety.

ozerovandrei: I guess it's safe to merge this as-is then 😄

ozerovandrei:
recheck

PR #635, opened 2019-01-21
Vet: fix unkeyed fields for importers

ozerovandrei:
Fix unkeyed fields for schema.ResourceImporter of the OpenStack LB and
Shared filesystem resources.


jtopjian:
OpenLab tests failed, but it looks like a general timeout issue of some sort - most likely transient. I don't see how this PR would cause it. I'm going to merge this now since it'll cause other PRs to fail.

Issue #627, opened 2019-01-17
DNS V2: recordset records are returned in a random order


kayrus:
### Affected Resource(s)
- openstack_dns_recordset_v2

Designate test fails:

> --- FAIL: TestAccDNSV2RecordSet_ipv6 (46.57s)
>    testing.go:518: Step 0 error: Check failed: Check 3/4 error: openstack_dns_recordset_v2.recordset_1: Attribute 'records.0' expected "fd2b:db7f:6ae:dd8d::1", got "fd2b:db7f:6ae:dd8d::2"

jtopjian:
That's a good catch. 

While OpenLab does not test Designate by default (you can trigger it by doing [`recheck designate`](https://www.terraform.io/docs/providers/openstack/index.html#recheck-designate), I test DNS in another cloud regularly and have not seen this problem.

I even checked my shell history to confirm I ran all tests before I submitted #568.

However, I just ran the `TestAccDNSV2RecordSet_ipv6` again and it failed. But then I ran it a second time and it passed:

```
2019/01/18 04:18:00 [DEBUG] OpenStack Response Body: {
  "action": "NONE",
  "created_at": "2019-01-18T04:17:16.000000",
  "description": "a record set",
  "id": "a59a7733-04ad-449c-b965-25bc3adcd54f",
  "name": "ACPTTEST-zone-s3fp1.com.",
  "records": [
    "fd2b:db7f:6ae:dd8d::1",
    "fd2b:db7f:6ae:dd8d::2"
  ],
  "status": "ACTIVE",
  "ttl": 3000,
  "type": "AAAA",
  "updated_at": null,
  "version": 1,
  "zone_id": "ad54b156-46ab-4252-9401-8de68167c101"
}
```

And then I ran it a third time and saw Designate reporting the correct order, but mid-way through the refresh func, Designate started sending the records back in a different order, meaning Designate decided to change the ordering.

I am hesitant about changing the `TypeList` to `TypeSet`. I imagine that a lot of users use the 0-based indexing to reference the records. Perhaps the `DiffSuppressFunc` will need further modified to suppress out of order diffs.

jtopjian:
> DiffSuppressFunc works on elements of a list/set and not on the list/set as a whole, so further modification there isn't going to help.

This is incorrect. `DiffSuppressFunc` provides access to the full schema, but it's still not a valid solution.

I think I have a solution to resolve this, but will need to work on it. 

[...]

PR #636, opened 2019-01-23
DNS v2: Fix recordset diffs

jtopjian:
For #627 

This commit fixes an issue where the records attribute in the
`openstack_dns_recordset_v2` resource would sometimes be returned in a
different order than what the user specified.

The hardest part about implementing this fix is managing the internal
feature of removing brackets from IPv6 addresses.

In order to successfully continue removing the brackets as well as
retain backwards compatibility, the following changes were made:

1. The `records` field is no longer being set in the Read function. This
is because the `records` field is not computed, so it does not need to
be set during Read.

2. An import function has been created so `records` is set when a user
imports a recordset.

3. `DiffSuppressFunc` has been replaced with a `StateFunc`. Because
`records` is no longer being set in Read, a diff will never occur, but
the modified IPv6 records still need to have records stripped.

4. `records` is being re-set in state during the Create function with
the IPv6 brackets stripped out.

I've tested this pretty extensively, including running this patch against an existing state file to confirm that no diffs will occur.  Importing also works, but I did not include an IPv6 import test because the records might be set out of order (this is an acceptable situation for this to happen).

### DIFF
--- a/openstack/dns_recordset_v2.go
+++ b/openstack/dns_recordset_v2.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform/helper/resource"
-	"github.com/hashicorp/terraform/helper/schema"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/dns/v2/recordsets"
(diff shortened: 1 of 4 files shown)


jtopjian:
recheck designate

Issue #639, opened 2019-01-24
Volume V2: Unable to destroyed volume attached via openstack_compute_volume_attach_v2


samcarpentier:
Hi,

I am encountering an issue with _openstack_compute_volume_attach_v2_ and/or _openstack_blockstorage_volume_v2_.

Once a _openstack_blockstorage_volume_v2_ has been attached to a compute instance via the _openstack_compute_volume_attach_v2_ resource, running `terraform destroy` returns the following error:

```
Error: Error applying plan:

1 error(s) occurred:

* openstack_blockstorage_volume_v2.myvol (destroy): 1 error(s) occurred:

* openstack_blockstorage_volume_v2.myvol: Error detaching openstack_blockstorage_volume_v2 f95e268e-d57d-401d-9132-0af08ebebbb0 from 653a0f69-a561-4faf-9800-09416bd3444d: Resource not found
```

### Terraform Version

```
[samcarpentier@MBP ~]$ terraform -v
Terraform v0.11.11
+ provider.openstack v1.14.0
```

### Affected Resource(s)

- openstack_blockstorage_volume_v2
- openstack_compute_volume_attach_v2

### Terraform Configuration Files
```hcl
provider "openstack" {
  auth_url         = "${var.auth_url}"
  tenant_name      = "${var.tenant_name}"
  region           = "${var.region}"
  user_domain_name = "${var.user_domain_name}"
}

resource "openstack_blockstorage_volume_v2" "myvol" {
  name = "myvol"
  size = 1
}

resource "openstack_compute_instance_v2" "myinstance" {
  name        = "myinstance"
  image_id    = "3210d2d3-601a-4d94-b299-bfaeb27f47bc"
  flavor_name = "m1.small"

  network {
    uuid = "1f93c662-04d4-44ec-9e59-c2315497c0a2"
  }
}

resource "openstack_compute_volume_attach_v2" "attached" {
  instance_id = "${openstack_compute_instance_v2.myinstance.id}"
  volume_id   = "${openstack_blockstorage_volume_v2.myvol.id}"
}
```

### Debug Output

https://gist.github.com/samcarpentier/c6d79175ece1971cb01716a8f889523a

### Expected Behavior

The _openstack_blockstorage_volume_v2_ should be destroyed correctly when used with the _openstack_compute_volume_attach_v2_ resource.

### Actual Behavior

[...]

PR #641, opened 2019-01-24
Block Storage v2: Fix volume detachment 404s

jtopjian:
For #639 

Amending #640, we shouldn't use CheckDeleted for the volume detach
portion since CheckDeleted will remove the resource from state. Instead,
we should just consider a 404 an acceptable error and continue.

Too many Deletes and CheckDelete on my mind.

### DIFF
--- a/openstack/resource_openstack_blockstorage_volume_v1.go
+++ b/openstack/resource_openstack_blockstorage_volume_v1.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v1/volumes"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/volumeattach"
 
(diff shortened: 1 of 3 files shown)


jtopjian:
OpenLab unit test error is ignorable - it looks to be a transient "connection reset by peer". Travis passed, so unit tests are fine.

jtopjian:
recheck

kayrus:
@jtopjian could you please also cover this use case?

> `* openstack_blockstorage_volume_v3.volumes.4: Error detaching openstack_blockstorage_volume_v3 c9f17e72-380b-42e3-bcba-20e89b2b5af2 from 04f9a74c-f1e5-4bd9-9b93-1bcbd198352c: Expected HTTP response code [202 204] when accessing [DELETE https://compute-3.eu-nl-1.cloud.sap:443/v2.1/servers/04f9a74c-f1e5-4bd9-9b93-1bcbd198352c/os-volume_attachments/c9f17e72-380b-42e3-bcba-20e89b2b5af2], but got 409 instead
{"conflictingRequest": {"message": "Cannot 'detach_volume' instance 04f9a74c-f1e5-4bd9-9b93-1bcbd198352c while it is in task_state deleting", "code": 409}}`

jtopjian:
@kayrus That's a really good find!

I've pushed a new commit which I _think_ will resolve it? Do you have a way of accurately reproducing this situation?

kayrus:
@jtopjian I'm performing so many use cases last days, so I already forgot what was the actual issue.

jtopjian:
No problem 🙂 

The error message was clear enough, so I think it's now accounted for.

ozerovandrei:
@jtopjian I'll go and merge it since we have a new release in plan.

jtopjian:
@ozerovandrei sounds good - thank you! :smile:

Issue #643, opened 2019-01-29
Networking V2: Unable to delete the network race condition


kayrus:
### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking_network_v2

### Debug Output

`openstack_networking_network_v2.network_1: Error deleting openstack_networking_network_v2 3f5591ea-8b16-4b57-b1b2-c68c386fbed9: Expected HTTP response code [202 204] when accessing [DELETE https://network:8080/v2.0/networks/3f5591ea-8b16-4b57-b1b2-c68c386fbed9], but got 409 instead`

In #578, the `waitForNetworkDelete` function was replaced with the `resourceNetworkingNetworkV2StateRefreshFunc`, which checks only for 404.
Relates to #639 as well. We have to revise delete functions again.
/cc @ozerovandrei

jtopjian:
> Relates to #639 as well. 

_Kind of_. #639 is a bug. The issue you've reported is a regression bug. 

I'll get this fixed and merged now. Thank you for reporting it.

jtopjian:
Resolved in #644

PR #644, opened 2019-01-29
Networking v2: Fix 409 bug with network refresh

jtopjian:
For #643


jtopjian:
This is a small bug fix, so I'm going to go ahead and merge it.

PR #645, opened 2019-01-29
Internal: fix schema set for the subnet resource

jtopjian:


kayrus:
Only one question, can a user create a new subnet without specifying the "host_routes", then specify custom routes? Will neutron rewrite old `host_routes` or just append new ones?

kayrus:
I haven't got an access to the openstack environment, and I was not sure whether `host_routes` is a kind of auto-generated field like `allocation_pools`.
I just confirmed that it is not, therefore this PR is LGTM.

jtopjian:
OK, cool. That scared me for a minute :slightly_smiling_face: 

This does highlight the situation where a user would modify the host routes outside of Terraform. There are several places in this provider where that situation isn't accounted for and it's usually not on purpose.

We've had several discussions about what to do when a user's defined arguments conflict with upstream and that usually ends up doing some complicated merging. But perhaps the situation here is as easy as just overwriting whatever is already set on the resource.

So in this case, if a user creates a subnet with host routes and that is changed outside of Terraform, then `Update` should just overwrite/clobber the host routes with what the user has defined. No merging, no enforcing, just overwrite.

This would involve changing this field to `TypeSet`, but it would also prevent a possible future issue where a user is reporting that changes to host routes aren't being reflected in Terraform.

jtopjian:
I've decided to revert the original implementation and change `host_routes` to a `TypeSet`. This is because I feel a host route is something that has a good chance of being modified outside of Terraform and someone will, rightfully, want to see Terraform make sure it is set correctly.

jtopjian:
This is one big rabbit hole. We're now running into issues with `openstack_networking_subnet_route_v2` causing diffs on `openstack_networking_subnet_v2`. I'll look into it.

jtopjian:
Closing in favor of #668

PR #647, opened 2019-01-29
Compute V2: Don't fail on delete, when compute instance doesn't exist

kayrus:
For #456


kayrus:
@jtopjian ready for review

PR #650, opened 2019-01-30
Keymanager V1: secrets resource

simonre:
Added an openstack_keymanager_secrets_v1 resource
Added an openstack_keymanager_secrets_metadata_v1 resource

@jtopjian I tried to adhere to the coding guidelines, please tell me if I missed anything.

Also, thinking about it now I feel like metadata should not be its own resource although it has its own create, update etc. functions in gophercloud. What do you think?


openstack/resource_openstack_keymanager_secret_metadata_v1.go
```diff
@@ -0,0 +1,126 @@
+package openstack
```
kayrus: Does it really make sense to split the metadata and its parent resource?

openstack/keymanager_secret_metadata_v1.go
```diff
@@ -0,0 +1,25 @@
...
+			return "", "NOT_CREATED", nil
+		}
+		return metadata, "ACTIVE", nil
+	}
+}
+
+// SecretMetadataCreateOpts represents the attributes used when creating a new Barbican secret.
+type SecretMetadataCreateOpts struct {
+	secrets.MetadataOpts
+	ValueSpecs map[string]string `json:"value_specs,omitempty"`
```
jtopjian: This is only needed if there's a need for `value_specs`. This usually gets determined when someone opens a request saying "my cloud has added extra options to do ..." 

Additionally, in order to use `ValueSpecs`, the initial `CreateOpts` struct will need to be changed to look like https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_keypair_v2.go#L69-L75

Let's remove this until requested.

openstack/keymanager_secret_v1.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/keymanager/v1/secrets"
+	"github.com/hashicorp/terraform/helper/resource"
+	"strings"
+)
+
+func keymanagerSecretV1WaitForSecretDeletion(kmClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		err := secrets.Delete(kmClient, id).Err
+		if err == nil {
```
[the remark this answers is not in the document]
jtopjian: It's been a number of months since I reviewed Barbican, so I can't remember if 409s are a thing. I _do_ remember that the API docs were trustworthy, though (https://docs.openstack.org/barbican/latest/api/reference/secrets.html). And since they don't mention a 409, I think we can leave them out until discovered.

But yes, a 404 check is important here - a 404 error shouldn't be considered an error, but success that the delete happened.

Additionally, I can't remember if `Delete` is a synchronous/blocking call. Do we need to poll to check if the secret was deleted? Meaning, should we perform an initial delete and then do `keymanagerSecretV1WaitForSecretDeletion` which will poll with a Get? Upon receiving a 404, that would indicate that the resource has been removed.

openstack/keymanager_secret_v1.go
```diff
@@ -0,0 +1,65 @@
...
+			return "", "DELETED", nil
+		}
+
+		return nil, "ACTIVE", err
+	}
+}
+
+func keymanagerSecretV1SecretType(v string) secrets.SecretType {
+	var stype secrets.SecretType
+	switch v {
```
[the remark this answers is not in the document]
jtopjian: If we use a `StringInSlice` validation in the schema, we can safely omit a default clause.

openstack/keymanager_secret_v1.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	return stype
+}
+
+func keymanagerSecretV1WaitForSecretCreation(kmClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		fmt.Println("ID is %v", id)
+		secret, err := secrets.Get(kmClient, id).Extract()
+		if err != nil {
```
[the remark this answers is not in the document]
jtopjian: Since this is for secret creation, we can continue to return the error and do a `CheckDeleted` on the returned error.

openstack/keymanager_secret_v1.go
```diff
@@ -0,0 +1,65 @@
...
+	// so we are only interested in the last part
+	ref_split := strings.Split(ref, "/")
+	uuid := ref_split[len(ref_split)-1]
+	return uuid
+}
+
+// SecretCreateOpts represents the attributes used when creating a new Barbican secret.
+type SecretCreateOpts struct {
+	secrets.CreateOpts
+	ValueSpecs map[string]string `json:"value_specs,omitempty"`
```
jtopjian: Same comment here re `value_specs`

openstack/resource_openstack_keymanager_secret_metadata_v1.go
jtopjian: As @simonre mentioned, he debated this. I can understand the debate - since metadata has CRUD actions, it _could_ be a dedicated resource. I'm loathe to say this because metadata can be a pain to work with within a resource, but let's not have a dedicated metadata resource - only manage it in the secrets resource. If a need comes up for a dedicated resource, we can create one later.

`openstack_compute_instance_v2` is probably the best place to copy a metadata implementation. Including a possible `all_metadata` field.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,232 @@
...
+		Payload:                d.Get("payload").(string),
+		PayloadContentType:     d.Get("payload_content_type").(string),
+		PayloadContentEncoding: d.Get("payload_content_encoding").(string),
+		SecretType:             secretType,
+	}
+
+	log.Printf("[DEBUG] Create Options for resource_keymanager_secret_v1: %#v", createOpts)
+
+	var secret *secrets.Secret
+	secret, err = secrets.Create(kmClient, createOpts).Extract()
```
jtopjian: missing err check

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,232 @@
...
+		Refresh:    keymanagerSecretV1WaitForSecretCreation(kmClient, uuid),
+		Timeout:    d.Timeout(schema.TimeoutCreate),
+		Delay:      0,
+		MinTimeout: 2 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack barbican secret: %s", err)
```
jtopjian: Per the previous comment about `CheckDeleted`, do:

```go
return CheckDeleted(d, err, "Error creating openstack_keymanager_secret_v1")
```

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,232 @@
...
+	d.Set("algorithm", secret.Algorithm)
+	d.Set("creator_id", secret.CreatorID)
+	d.Set("mode", secret.Mode)
+	d.Set("secret_ref", secret.SecretRef)
+	d.Set("secret_type", secret.SecretType)
+	d.Set("status", secret.Status)
+	d.Set("created", secret.Created)
+	d.Set("updated", secret.Updated)
+	d.Set("expiration", secret.Expiration)
+	d.Set("content_types", secret.ContentTypes)
```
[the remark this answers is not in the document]
jtopjian: Agreed - these should be added to the schema as computed-only fields.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,232 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"status": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+			"payload": {
```
kayrus: I believe that the payload should be `Sensitive: true` in order to avoid diff printed with a private key printed.

kayrus: It is very dangerous in case of using remote backend, and even local states, because by default they have open read permissions. There should be a note in documentation. I just verified that terraform doesn't mask sensitive data in the state. Terraform 0.12 [has a possibility](https://github.com/hashicorp/terraform/pull/19464) to [mask sensitive fields](https://github.com/hashicorp/terraform/blob/1a8ddc26a5d6c7360c4b8806a08f1646de6cc246/backend/remote/backend_context.go#L84), but not 0.11.x

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,232 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+			},
+			"bit_length": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+			"algorithm": {
```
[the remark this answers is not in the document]
simonre: I think 'algorithm' and 'mode' are not limited in values they can have (unless I'm missing something)

simonre:
@jtopjian @kayrus sorry for the long delay, I was a bit busier than I thought I'd be over the last few weeks. Does this look better?

kayrus:
recheck

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+		hasChange = true
+		updateOpts.ContentEncoding = d.Get("content_encoding").(string)
+	}
+	if d.HasChange("payload") {
+		hasChange = true
+		updateOpts.Payload = d.Get("payload").(string)
+	}
+
+	if hasChange {
+		err := secrets.Update(kmClient, d.Id(), updateOpts).Err
```
kayrus: @jtopjian , not sure whether it's recommended to print `updateOpts` debug log. If it is recommended, the `payload` sensitive attribute should be considered and it should not be printed.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+				if oldKey == newKey {
+					found = true
+				}
+			}
+
+			if !found {
+				metadataToDelete = append(metadataToDelete, oldKey)
+			}
+		}
+
```
kayrus: there should be a debug log with the list of elements to be deleted.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+						metadataToUpdate = append(metadataToUpdate, newKey)
+					}
+				}
+			}
+
+			if !found {
+				metadataToAdd = append(metadataToAdd, newKey)
+			}
+		}
+
```
kayrus: there should be a debug log with the list of elements to be updated.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+		for _, key := range metadataToUpdate {
+			var metadatumOpts secrets.MetadatumOpts
+			metadatumOpts.Key = key
+			metadatumOpts.Value = newMetadata.(map[string]interface{})[key].(string)
+			_, err := secrets.UpdateMetadatum(kmClient, d.Id(), metadatumOpts).Extract()
+			if err != nil {
+				return fmt.Errorf("Error updating OpenStack secret (%s) metadata: %s", d.Id(), err)
+			}
+		}
+
```
kayrus: there should be a debug log with the list of elements to be added.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+		for _, key := range metadataToAdd {
+			var metadatumOpts secrets.MetadatumOpts
+			metadatumOpts.Key = key
+			metadatumOpts.Value = newMetadata.(map[string]interface{})[key].(string)
+			err := secrets.CreateMetadatum(kmClient, d.Id(), metadatumOpts).Err
+			if err != nil {
+				return fmt.Errorf("Error updating OpenStack secret (%s) metadata: %s", d.Id(), err)
```
kayrus: Error \_adding\_ OpenStack secret

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+	_, err = stateConf.WaitForState()
+
+	if err != nil {
+		return CheckDeleted(d, err, "Error creating openstack_keymanager_secret_v1")
+	}
+
+	var metadataCreateOpts secrets.MetadataOpts
+	metadataCreateOpts = keymanagerSecretMetadataV1(d)
+
+	log.Printf("[DEBUG] Create Options for resource_keymanager_secret_metadata_v1: %#v", createOpts)
```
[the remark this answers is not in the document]
simonre: I'm not quite sure what you mean by this comment?

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+	d.Set("mode", secret.Mode)
+	d.Set("secret_ref", secret.SecretRef)
+	d.Set("secret_type", secret.SecretType)
+	d.Set("status", secret.Status)
+	d.Set("created", secret.Created.Format(time.RFC3339))
+	d.Set("updated", secret.Updated.Format(time.RFC3339))
+	d.Set("expiration", secret.Expiration.Format(time.RFC3339))
+	d.Set("content_types", secret.ContentTypes)
+
+	metadataMap, err := secrets.GetMetadata(kmClient, d.Id()).Extract()
```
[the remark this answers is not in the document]
simonre: Good catch, will add that

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+
+		for _, key := range metadataToDelete {
+			err := secrets.DeleteMetadatum(kmClient, d.Id(), key).ExtractErr()
+			if err != nil {
+				return fmt.Errorf("Error deleting metadata (%s) from secret (%s): %s", key, d.Id(), err)
+			}
+		}
+
+		oldMetadata, newMetadata = d.GetChange("metadata")
```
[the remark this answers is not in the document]
simonre: No, just a Copy and paste error I think

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+			metadatumOpts.Key = key
+			metadatumOpts.Value = newMetadata.(map[string]interface{})[key].(string)
+			_, err := secrets.UpdateMetadatum(kmClient, d.Id(), metadatumOpts).Extract()
+			if err != nil {
+				return fmt.Errorf("Error updating OpenStack secret (%s) metadata: %s", d.Id(), err)
+			}
+		}
+
+		for _, key := range metadataToAdd {
+			var metadatumOpts secrets.MetadatumOpts
```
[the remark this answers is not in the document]
simonre: There is no possibility to update metadata with the secrets.MetadataOpts type. As far as I know, the only possibility would be to use CreateMetaData as that would delete the old metadata and replace it with the new metadata. I thought it would be better to keep the metadata update code similar across the provider (as seen in the compute_instance_v2). Looping over the metadata also allows to print debug logs with lists of elements to be added, updated and deleted. 
If the interest is in keeping the code short, I'd gladly replace this with just a single call to "createMetadata".

openstack/resource_openstack_keymanager_secret_v1.go
simonre: I think while we're technically adding metadata here, that wouldn't matter to the user as from the point of view of the user, he is updating the metadata. Or is there something I'm missing?

simonre:
@kayrus Thanks for taking the time to review, I appreciate it :)
I took the "keymanager" name from gophercloud, I'm not opposed to changing it though. I'll wait for @jtopjian to chime in

openstack/resource_openstack_keymanager_secret_v1.go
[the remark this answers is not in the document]
kayrus: `Create Options` debug already took place on the 144 line. Here should be `Metadata Create Options`.

openstack/resource_openstack_keymanager_secret_v1.go
kayrus: Then something like: `error adding/updating metadata while updating OpenStack secret`?

kayrus:
@simonre my suggestion is based on `BlockStorage` in `blockstorage`, or `ObjectStorage` in `objectstorage`

jtopjian:
Hi all,

This is on my list to look at. 

But with regard to the name, honestly as long as it's consistent, I'm fine with it. The only rule is to be all lowercase for things like the resource names (`keymanager`).

The best case scenario would be to use `keyManager` and `KeyManager` where appropriate, but it's not a huge deal if occurrences are missed. If you look closely at all the code, there are small inconsistencies everywhere :)

TylerHaze:
Hey, my company was actually just looking at how we can get Barbican Secrets in Terraform! Is there anything I could help you with on this, I've been looking to play with Go and I love working with Terraform, so I'd love to contribute.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+	metadataMap, err := secrets.GetMetadata(kmClient, d.Id()).Extract()
+	d.Set("all_metadata", metadataMap)
+
+	// Set the region
+	d.Set("region", GetRegion(d, config))
+
+	return nil
+}
+
+func resourceKeymanagerSecretV1Update(d *schema.ResourceData, meta interface{}) error {
```
jtopjian: Related to @kayrus's comment here (https://github.com/terraform-providers/terraform-provider-openstack/pull/650/files#r259406141), since the only attributes that can be updated are `payload`, `payload_content_type`, `payload_content_encoding`, and `metadata`, then all other schema options need to be set to `ForceNew: true` (mode, secret_type, etc)

openstack/resource_openstack_keymanager_secret_v1.go
jtopjian: I can go either way with this. Since `Payload` is the only relevant field and since it's sensitive, I can see not printing a debug line.

But it might be worthwhile to print with `ContentEncoding` and `ContentType`. In this case, I would add a debug line before:

```go
if d.HasChange("payload") {
```

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	if hasChange {
+		err := secrets.Update(kmClient, d.Id(), updateOpts).Err
+		if err != nil {
+			return err
+		}
+	}
+
+	if d.HasChange("metadata") {
```
jtopjian: I did not test this, but I cleaned up this area of code to look like:

```go
  if d.HasChange("metadata") {
    var metadataToDelete []string
    var metadataToAdd []string
    var metadataToUpdate []string

    o, n := d.GetChange("metadata")
    oldMetadata := o.(map[string]interface{})
    newMetadata := n.(map[string]interface{})

    // Determine if any metadata keys were removed from the configuration.
    // Then request those keys to be deleted.
    for oldKey := range oldMetadata {
      if _, ok := newMetadata[oldKey]; !ok {
        metadataToDelete = append(metadataToDelete, oldKey)
      }
    }

    for _, key := range metadataToDelete {
      err := secrets.DeleteMetadatum(kmClient, d.Id(), key).ExtractErr()
      if err != nil {
        return fmt.Errorf("Error deleting openstack_keymanager_secret_v1 %s metadata %s: %s", d.Id(), key, err)
      }
    }

    // Determine if any metadata keys were updated or added in the configuration.
    // Then request those keys to be updated or added.
    for newKey, newValue := range newMetadata {
      if oldValue, ok := oldMetadata[newKey]; ok {
        if newValue != oldValue {
          metadataToUpdate = append(metadataToUpdate, newKey)
        }
      } else {
        metadataToAdd = append(metadataToAdd, newKey)
      }
    }

    for _, key := range metadataToUpdate {
      var metadatumOpts secrets.MetadatumOpts
      metadatumOpts.Key = key
      metadatumOpts.Value = newMetadata[key].(string)
      _, err := secrets.UpdateMetadatum(kmClient, d.Id(), metadatumOpts).Extract()
      if err != nil {
        return fmt.Errorf("Error updating openstack_keymanager_secret_v1 %s metadata %s: %s", d.Id(), key, err)
      }
    }

    for _, key := range metadataToAdd {
      var metadatumOpts secrets.MetadatumOpts
      metadatumOpts.Key = key
      metadatumOpts.Value = newMetadata[key].(string)
      err := secrets.CreateMetadatum(kmClient, d.Id(), metadatumOpts).Err
      if err != nil {
        return fmt.Errorf("Error creating openstack_keymanager_secret_v1 %s metadata %s: %s", d.Id(), key, err)
      }
    }
  }
```

openstack/resource_openstack_keymanager_secret_v1.go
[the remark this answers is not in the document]
jtopjian: I agree that the approach being used is valid.

The reason to act on individual metadata in `Update` is to not clobber any metadata that the API service might have set on its own.

openstack/resource_openstack_keymanager_secret_v1.go
jtopjian: Agree with "adding"

openstack/keymanager_secret_v1.go
```diff
@@ -0,0 +1,82 @@
+package openstack
+
+import (
```
jtopjian: nit: standard packages go first, then a newline, and then the rest:

```go
import (
	"fmt"
	"strings"

	"github.com/gophercloud/gophercloud"
	"github.com/gophercloud/gophercloud/openstack/keymanager/v1/secrets"
	"github.com/hashicorp/terraform/helper/resource"
	"github.com/hashicorp/terraform/helper/schema"
)
```

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+		MinTimeout: 2 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack barbican secret metadata: %s", err)
+	}
+
+	d.SetId(uuid)
```
jtopjian: This `Create` function has two phases:

1. Create the secret
2. Set the metadata on the secret

If phase 2 fails, then the secret will be created and orphaned. Upon the next create, a new secret will be created.

Instead, we should be able to use `d.Partial` here.

After secret creation, but before metadata creation, do:

```go
d.SetId(uuid)
d.SetPartial(true)
```

And then after metadata creation:

```go
d.Partial(false)
```

I think that should work - I'm not too familiar with `Partial`.

openstack/resource_openstack_keymanager_secret_v1_test.go
```diff
@@ -0,0 +1,266 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/keymanager/v1/secrets"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccSecretV1_basic(t *testing.T) {
```
jtopjian: nit: test function names should be `TestAccKeymanagerSecretV1_`.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return CheckDeleted(d, err, "Error creating openstack_keymanager_secret_v1")
+	}
+
+	var metadataCreateOpts secrets.MetadataOpts
+	metadataCreateOpts = keymanagerSecretMetadataV1(d)
+
+	log.Printf("[DEBUG] Create Options for resource_keymanager_secret_metadata_v1: %#v", createOpts)
+
+	_, err = secrets.CreateMetadata(kmClient, uuid, metadataCreateOpts).Extract()
```
jtopjian: We should only run this is the length of `metadataCreateOpts` is > 0.

openstack/import_openstack_keymanager_secret_v1_test.go
```diff
@@ -0,0 +1,31 @@
+package openstack
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccSecretV1_importBasic(t *testing.T) {
```
jtopjian: nit: `TestAccKeyManagerSecretV1_importBasic`

openstack/keymanager_secret_v1.go
```diff
@@ -0,0 +1,89 @@
...
+
+func keymanagerSecretV1GetUUIDfromSecretRef(ref string) string {
+	// secret ref has form https://{barbican_host}/v1/secrets/{secret_uuid}
+	// so we are only interested in the last part
+	ref_split := strings.Split(ref, "/")
+	uuid := ref_split[len(ref_split)-1]
+	return uuid
+}
+
+func flattenKeyManagerSecretMetadataV1(d *schema.ResourceData) map[string]string {
```
jtopjian: nit: `flattenKeyManagerSecretV1Metadata`

openstack/resource_openstack_keymanager_secret_v1_test.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		if metadatum.Value != value {
+			return fmt.Errorf("Metadata does not match. Expected %v but got %v", metadatum, value)
+		}
+
+		return nil
+	}
+}
+
+var testAccKeymanagerSecretV1_basic = fmt.Sprintf(`
```
jtopjian: nit: since you're not doing any variable interpolation here, you can just use `const`:

```go
const testAccKeyManagerSecretV1_basic = `
...
`
```

website/docs/r/keymanager_secret_v1.html.markdown
```diff
@@ -0,0 +1,80 @@
...
+page_title: "OpenStack: openstack_keymanager_secret_v1"
+sidebar_current: "docs-openstack-resource-keymanager-secret-v1"
+description: |-
+  Manages a V1 Barbican secret resource within OpenStack.
+---
+
+# openstack\_keymanager\_secret\_v1
+
+Manages a V1 Barbican secret resource within OpenStack.
+
```
jtopjian: We need to add a security notice similar to here: https://raw.githubusercontent.com/terraform-providers/terraform-provider-openstack/master/website/docs/r/compute_keypair_v2.html.markdown

I think the following will work:

```
~> **Important Security Notice** The payload of this resource will be stored
*unencrypted* in your Terraform state file. **Use of this resource for production
deployments is *not* recommended**.
```

I understand that telling people not to use this in production is diminishing the value of this resource and whether they still want to use it in production is up to them. But we need to clearly convey the security concern here.

simonre:
@jtopjian I hope the vendor commit is okay now. Thank you for your patience with me 😄

kayrus:
I'll take a look later today.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+				ForceNew:  false,
+			},
+
+			"payload_content_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
```
kayrus: Set ForceNew to true as well, since I get this error:

> Secret already has data, cannot modify it.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+	var updateOpts secrets.UpdateOpts
+	if d.HasChange("payload_content_type") {
+		hasChange = true
+	}
+	// This is not optional so we have to set it regardless
+	updateOpts.ContentType = d.Get("payload_content_type").(string)
+
+	if d.HasChange("payload_content_encoding") {
+		hasChange = true
+		updateOpts.ContentEncoding = d.Get("content_encoding").(string)
```
kayrus: there is a typo. it must be: `payload_content_encoding`, not `content_encoding`

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+
+			"payload_content_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
+
+			"payload_content_encoding": {
+				Type:     schema.TypeString,
+				Optional: true,
```
kayrus: `payload_content_encoding` must be either `base64` or `binary` and only when the `payload_content_type` is set to `application/octet-stream`

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+			},
+
+			"payload": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+				ForceNew:  false,
+			},
+
+			"payload_content_type": {
```
kayrus: Please limit the input options:

> payload_content_type is not one of ['text/plain', 'text/plain;charset=utf-8', 'text/plain; charset=utf-8', 'application/octet-stream', 'application/pkcs8']

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
+
+			"payload_content_encoding": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
```
kayrus: Set `ForceNew: true`. Got an error:

> Secret already has data, cannot modify it.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+	secret, err = secrets.Create(kmClient, createOpts).Extract()
+
+	if err != nil {
+		return fmt.Errorf("Error creating openstack_keymanager_secret_v1: %s", err)
+	}
+
+	uuid := keyManagerSecretV1GetUUIDfromSecretRef(secret.SecretRef)
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"NOT_CREATED"},
```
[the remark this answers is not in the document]
jtopjian: Are you sure `ERROR` is a valid pending state? Is the resource ever going to recover from the error?

Otherwise, I agree with returning `secret.Status`.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+			"payload": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+				ForceNew:  false,
+			},
+
+			"payload_content_type": {
+				Type:     schema.TypeString,
+				Optional: true,
```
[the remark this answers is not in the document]
jtopjian: > Can we set the default value to be text/plain? 

Instead of setting a default value, I would prefer just limiting the input to the values you specified in another comment.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"updated_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"expiration": {
```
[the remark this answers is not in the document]
jtopjian: Expiration is not a valid field in Gophercloud at this time.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+	d.Set("created_at", secret.Created.Format(time.RFC3339))
+	d.Set("updated_at", secret.Updated.Format(time.RFC3339))
+	d.Set("expiration", secret.Expiration.Format(time.RFC3339))
+	d.Set("content_types", secret.ContentTypes)
+
+	metadataMap, err := secrets.GetMetadata(kmClient, d.Id()).Extract()
+	if err != nil {
+		log.Printf("[DEBUG] Unable to set metadata: %s", err)
+	}
+	d.Set("all_metadata", metadataMap)
```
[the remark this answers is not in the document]
jtopjian: This should be fine. `metadata` should be used as input-only from Terraform. `all_metadata` is from the API.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	secret, err := secrets.Get(kmClient, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "Error retrieving openstack_keymanager_secret_v1")
+	}
+
+	log.Printf("[DEBUG] Retrieved openstack_keymanager_secret_v1 %s: %#v", d.Id(), secret)
+
+	d.Set("name", secret.Name)
```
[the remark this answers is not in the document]
jtopjian: The payload should definitely be imported. Otherwise importing isn't really useful.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,389 @@
...
+			"status": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"payload": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+				ForceNew:  false,
```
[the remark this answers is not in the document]
jtopjian: Is there a valid use-case for having a secret with no initial payload? If so, then we shouldn't change this to true.

But I'm leaning towards agreeing here. For Terraform, I can't think of a use-case where a payload would be set later. I might be wrong.

openstack/resource_openstack_keymanager_secret_v1.go
jtopjian: Related to `ForceNew` of `payload`.

jtopjian:
@kayrus all good comments - thanks!

> I don't feel comfortable, when the acls are not built-in inside the resource. I'd suggest to set default ACLs to be strict. Then an additional resource can set the permissive acls, if needed. What do you think?

I think we should adhere to the defaults of the API and document this behavior.

jtopjian:
Github doesn't do too well with lengthy discussions. I hope this doesn't muddy the water, but I wanted to try to summarize @kayrus's comments (which are very relevant):

1. Importing

I totally missed that the payload was not being imported. We need to make sure that it can and is imported (along with the related arguments). If it is not possible to import those, then IMO this resource shouldn't support importing. 

2. Returning `secret.Status` and adding `PENDING` to the pending states: yes - good catch. I think polling function should now look like:

```go
func keyManagerSecretV1WaitForSecretCreation(kmClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
	return func() (interface{}, string, error) {
		secret, err := secrets.Get(kmClient, id).Extract()
		if err != nil {
				return "", "NOT_CREATED", nil
			}

			return "", "NOT_CREATED", err
		}

                if secret.Status == "ERROR" {
                        return "", secret.Status, fmt.Errorf("Error creating secret")
                }

		return secret, secret.Status, nil
	}
}
```

However, I think we only need to do this for secret creation and not metadata creation.

3. Should an empty payload be possible? 

This is open to discussion. I think there is a reason for this in the API, possibly when combined with Barbican's "orders". But I don't know if "orders" are a very good thing to implement in Terraform.

Therefore, it's possible that the payload (and related fields) should all be `ForceNew`. I think then the only thing possible to update is the metadata.

However, if it should be possible to create an initial secret with an empty payload, then we will need to add an internal check, per @kayrus's comments (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269315208)

4. Extra check: (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269334875): Yes

5. case-sensitive validation (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269319520): Yes

6. Computed `secret_type` (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269320287): Yes, sounds right.

7. Setting `expiration` (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269321063): Unfortunately this is not possible at this time.

8. Limiting `payload_content_type` (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269335420): Yes, as well as add this to the documentation.

9. Typo: https://github.com/terraform-providers/terraform-provider-openstack/pull/650#discussion_r269335420

In addition, we should add a few more acceptance tests to validate some of the above.

Finally, per my comment here (https://github.com/terraform-providers/terraform-provider-openstack/pull/650#issuecomment-476880745): 

I still think this is valid. I understand the concern of the default security possibly being too lax, but since we can create an ACL resource, I think it would be best to leave the default security as-is and place the responsibility on the user to ensure they secure the secret correctly.

There are two reasons for this:

1. A future change to Barbican (or a custom middleware) could alter this and conflict with our hard-coded default settings.
2. We are taking on responsibility for what we think the default ACL should be. This can be very subjective and debatable. 

I think that's everything? Again, I'm sorry if this actually more problems for readability.

@simonre Please do not feel rushed to make these changes. Additionally, if you'd like, I could take over (Github is telling me I can amend commits to your branch).

kayrus:
> Therefore, it's possible that the payload (and related fields) should all be ForceNew.

agree.

1) @jtopjian I couldn't find the `TypeMap` fix for the `content_types` in your summarize comment.

2) Regarding the `expiration` parameter. It is easy to introduce it. I already tested it in my environment. I created a corresponding issue in gophercloud: https://github.com/gophercloud/gophercloud/issues/1526. I expect you'll reject it and I'm ok with that, but in certain cases I prefer to prepare the PR in advance.

jtopjian:
> I expect you'll reject it and I'm ok with that, but in certain cases I prefer to prepare the PR in advance.

I wouldn't outright reject it. It might be easier to defer adding that until this PR is merged. I'll leave that up to @simonre, though.

andrewsc0:
Hello! I am interested in this module that has been built. I was wondering when it will get merged and if it will come out with a new version of terraform provider openstack or as just an add-on? Any word would be great!

simonre:
Thanks for the review @kayrus . I'll push the changes asap. 
I think its easier to add the expiration change later. 
I also cannot think of a usecase for setting the payload after creating the secret in terraform. Should I delete the functionality then? @jtopjian 
@TylerHaze Sorry, I missed your reply for some reason. Feel free to review my code for errors 😄

jtopjian:
> I also cannot think of a usecase for setting the payload after creating the secret in terraform. Should I delete the functionality then? @jtopjian

Yes, it sounds like that's the consensus. It's also easier to add this ability later than remove it later.

simonre:
@jtopjian Feel free to amend commits to this and take over from here

simonre:
@jtopjian 
@kayrus  
Just want to make sure you've seen this 😄

jtopjian:
@simonre Yep, acknowledged - I'll get to this when I have some time. Thank you so much for your help!

kayrus:
@simonre could you please fix the conflicts? Or I can create take a look on this PR this weekend and create a new PR based on yours if I find some issues during the tests.

simonre:
@kayrus I fixed the conflicts. Feel free to commit a new PR based on mine if you feel like it's not moving fast enough(I can only work on this a few hours a week unfortunately) or my PR still has huge issues😄

ozerovandrei:
@simonre I think that current PR is great and I'm happy to merge it after the successful tests results.

kayrus:
@ozerovandrei please hold on till the end of this week. I'd like to perform manual tests.

ozerovandrei:
@kayrus sure.

openstack/resource_openstack_keymanager_secret_v1.go
```diff
@@ -0,0 +1,365 @@
...
+			"status": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"payload": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+				ForceNew:  true,
```
kayrus: I found a bug here. If you create the payload with whitespaces, then openstack API strips them: https://github.com/openstack/barbican/blob/0477799ea7421d884ab31d0f495ebccdc47fbce3/barbican/api/__init__.py#L63

It's easy to reproduce if you use this snippet:

```hcl
resource "openstack_keymanager_secret_v1" "secret_1" {
  name = "terraform"
  payload = <<EOF
123
EOF
  secret_type = "private"
  payload_content_type = "text/plain"
}
```

Not sure whether it is an openstack bug, but I'd add `DiffSuppressFunc` here.

openstack/resource_openstack_keymanager_secret_v1.go
kayrus: If `payload_content_encoding` is `base64` and the payload is in base64 format, then `terraform apply` is not idempotent. The `CustomizeDiff` is a potential solution.

ozerovandrei:
Merged in #807

@simonre @kayrus thank you for working on this!

PR #652, opened 2019-01-31
Add ability to list availability zones for blockstorage and sharedfilesystem

### LINKED ISSUE #651: Add possibility to list available availability zones
In order when user wants to create multiple clusters bound to the availability zone, it would be great to fetch the list of availability zones as a data source.

### Affected Data Source(s)

- openstack_availability_zones

kayrus:
Resolves #651


kayrus:
@jtopjian ready for review

jtopjian:
There should be 3 data sources for these - one for each service.

kayrus:
How do you suggest to call new data sources?
* openstack_compute_az
* openstack_compute_azs
* openstack_compute_availability_zones

kayrus:
For HA deployments, scaled accross multiple regions. Some regions have different amount of AZ. For sure, they can be specified manually in a list, but if it can be automated, I'd automate it, especially when you have more than 10 regions. Do you have other thoughts?

kayrus:
@jtopjian ready for review

kayrus:
@jtopjian could you please take another look on this PR?

PR #654, opened 2019-02-02
Manila: recreate share and its access resources, when the region is changed

### LINKED ISSUE #653: Manila: share and its access is not recreated, when the region has been changed
### Affected Resource(s)

- openstack_sharedfilesystem_share_v2
- openstack_sharedfilesystem_share_access_v2

There is a typo inside the resource schema and documentation

kayrus:
Resolves #653


kayrus:
@jtopjian ready for review

kayrus:
This was my bad. I didn't know the code well (I don't know it now though), and, probably, copy-pasted it from the data source.

Issue #659, opened 2019-02-07
openstack_networking_port_v2 does not allow cidr in allowed_address_pairs


iwagner-inmar:
Adding a openstack_networking_port_v2 resource with a cidr range in allowed_address_pairs will cause the terraform plan to throw an error.

This appears to have broken with version 1.15.0. It works correctly with 1.14.0

### Terraform Version
Terraform v0.11.11
+ provider.openstack v1.15.0

### Affected Resource(s)
- openstack_networking_port_v2

### Terraform Configuration Files
```hcl
provider "openstack" {
        version = "<= 1.15.0"
}

resource "openstack_networking_port_v2" "network_port" {
  name           = "myport"
  network_id     = "a88e32ea-8720-4112-8df8-b536c7daa637"
  admin_state_up = true


  allowed_address_pairs = [
    {
      ip_address = "10.1.1.0/24"
    }
  ]

}
```

### Debug Output
There is no debug output prior to the validation errors in `terraform plan`

### Expected Behavior
Plan should have validated OK.

### Actual Behavior
```$ terraform plan

Error: openstack_networking_port_v2.network_port: expected allowed_address_pairs.0.ip_address to contain a valid IP, got: 10.1.1.0/24
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform plan`

### Important Factoids
Running Openstack Queens

jtopjian:
@iwagner-inmar Thank you for reporting this. #661 will fix this. We'll get a quick release out early next week.

jtopjian:
@iwagner-inmar This has been fixed and we've also released version 0.15.1 to get this deployed. Thank you again for reporting this and I apologize for the inconvenience.

PR #661, opened 2019-02-08
Networking v2: Fix Allowed Address Pairs validation

jtopjian:
This change reverts a previous modification which restricted an allowed
address pair's IP address to be a single IP address rather than a CIDR.

For #659


kayrus:
My bad, I was confused by an `ip_address` name. But we can also add acceptance test for the CIDR case to avoid this misunderstanding in future?

In addition, can we add an extra validation function like [this](https://gist.github.com/kayrus/5006cf5e2c12b21f31c283f6e8f488ba)?

PR #662, opened 2019-02-08
Networking v2: Fix Allowed Address Pairs validation

kayrus:
An alternative fix for #661


kayrus:
@ozerovandrei @jtopjian , kindly review this PR.

jtopjian:
@kayrus This looks like a legitimate fix, but I would prefer to just revert back to the original configuration for now. I'm behind on the code cleanup and preparing for v0.12 and I don't want to minimize possibilities of further bugs right now.

kayrus:
If you don't mind, I'll keep it open for the future.

jtopjian:
Absolutely :)

PR #666, opened 2019-02-08
Compute v2: Add availability_zone_hints parameter for the instance resource

### LINKED ISSUE #658: Unable to refresh the state of the nova compute
Hi, 

Terraform is unable to refresh and pull the latest state of the compute resource, though the resource is successfully provisioned. It still destroys and re-creates on the next Run. 

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see https://www.terraform.io/community.html.

### Terraform Version
Terraform Version = 11.11.0
### Affected Resource(s)

```hcl

resource "openstack_compute_instance_v2" "baremetal" {
  count = "${length(var.nodes)}"
  name  = "${lookup(var.nodes[count.index], "name")}"
  availability_zone = "::${lookup(var.nodes[count.index], "uuid")}"
  region    = "${var.region}"
  image_id  = "${var.image}"
  flavor_id = "${var.flavor}"
  key_pair  = "${var.key_pair}"
  security_groups = "${var.security_group}"

  network
  {
   uuid = "${var.access_network}"
   access_network = "true"
   fixed_ip_v4 = "${lookup(var.nodes[count.index], "fixed_ip")}"
  }
  
  
 network
{
  uuid = "${var.backup_network}"
}

  network
{
  uuid = "${var.storage_network}"
}

  network
{
  uuid = "${var.heart_beat_network}"
}


  lifecycle {
    prevent_destroy = true
  }


 timeouts {
    create = "45m"
    update = "30m"
    delete = "30m"
  }

}

### Debug Output ###

https://gist.github.com/vikram-chandrakanth/6b3535af5b16015726afc318ff1b6279 

### Panic Output

https://gist.github.com/vikram-chandrakanth/6b3535af5b16015726afc318ff1b6279

### Expected Behavior
What should have happened?

### Actual Behavior
What actually happened?

### Steps to Reproduce
1. `terraform apply`

kayrus:
Resolves #658


kayrus:
@jtopjian ready for review

kayrus:
it looks like there should be desired (argument) and computed (attribute) AZs fields.

kayrus:
@jtopjian It appeared, that diff suppress func cannot help, because it doesn't know where the new parameter comes from (read, or terraform manifest). Adding a new parameter should allow to define AZ in a `zone:host:node` format and keep backward compatibility.

PR #667, opened 2019-02-10
Internal: fwaas resources cleanup

jtopjian:
For #456 

The FWaaS v1 API code is flaky since there were some breaking changes introduced in v1. It's also quite difficult to get a working FWaaS v1 service running in newer versions of Neutron. I've run the acceptance tests in two different environments, running two different versions of OpenStack, and tests pass in the version they were targeted for.

It's totally possible to use these resources in production, but I'm sure those who are have workarounds they are doing. This is not a fault of the resources, but of the API code.

FWaaS v1 is not tested in OpenLab.

### DIFF
--- a/openstack/fw_firewall_v1.go
+++ b/openstack/fw_firewall_v1.go
@@ -6,6 +6,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas/firewalls"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas/routerinsertion"
+
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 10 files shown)


openstack/fw_firewall_v1.go
```diff
@@ -0,0 +1,65 @@
+package openstack
+
+import (
+	"fmt"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas/firewalls"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas/routerinsertion"
+	"github.com/hashicorp/terraform/helper/resource"
```
ozerovandrei: It's better to separate this import with empty line as in `fw_policy_v1.go`.

PR #668, opened 2019-02-10
Networking v2: Deprecate openstack_networking_subnet_v2.host_routes

jtopjian:
This commit deprecates the host_routes argument of the
openstack_networking_subnet_v2 resource. Given there is a schema issue
with setting the host_routes attribute and that using both this resource
and the openstack_networking_subnet_route_v2 resource will cause diffs
to occur, it's best to have the supported solution be to use the
dedicated resource.


jtopjian:
@kayrus I see you closed #618 in favor of this, but before I merged it, just wanted to see if you had any thoughts?

kayrus:
@jtopjian, using the `openstack_networking_subnet_route_v2` is fair enough in this situation. However, if user wants to set host_routes in a specific order, there should be an explanation in the doc on how to achieve this.

PR #670, opened 2019-02-10
Internal: openstack_compute_floatingip_associate_v2 cleanup

jtopjian:
For #456

### DIFF
--- a/openstack/compute_floatingip_associate_v2.go
+++ b/openstack/compute_floatingip_associate_v2.go
@@ -40,7 +40,7 @@ func computeFloatingIPAssociateV2NetworkExists(networkClient *gophercloud.Servic
 	}
 
 	if len(allFips) > 1 {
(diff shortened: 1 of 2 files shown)


openstack/compute_floatingip_associate_v2.go
```diff
@@ -0,0 +1,98 @@
...
+		return false, err
+	}
+
+	allFips, err := nfloatingips.ExtractFloatingIPs(allPages)
+	if err != nil {
+		return false, err
+	}
+
+	if len(allFips) > 1 {
+		return false, fmt.Errorf("There was a problem retrieving the floating IP")
```
ozerovandrei: Maybe its better to make error message more specific, something like
```
fmt.Errorf("more than one floating IP with '%s' address found", floatingIP)
```

jtopjian: agreed - sounds good.

PR #671, opened 2019-02-11
Internal: Cleanup database resources

jtopjian:
For #456

### DIFF
--- a/openstack/db_configuration_v1.go
+++ b/openstack/db_configuration_v1.go
@@ -1,0 +1,55 @@
+package openstack
+
+import (
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/resource"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/db/v1/configurations"
+)
+
+func expandDatabaseConfigurationV1Datastore(rawDatastore []interface{}) configurations.DatastoreOpts {
+	v := rawDatastore[0].(map[string]interface{})
+	datastore := configurations.DatastoreOpts{
+		Version: v["version"].(string),
+		Type:    v["type"].(string),
+	}
+
+	return datastore
+}
+
+func expandDatabaseConfigurationV1Values(rawValues []interface{}) map[string]interface{} {
+	values := make(map[string]interface{})
+
+	for _, rawValue := range rawValues {
+		v := rawValue.(map[string]interface{})
+		name := v["name"].(string)
+		value := v["value"].(interface{})
+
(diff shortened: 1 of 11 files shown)


jtopjian:
recheck trove

jtopjian:
recheck trove

kiwik:
recheck trove

kiwik:
recheck trove

jtopjian:
recheck trove

jtopjian:
recheck trove

kiwik:
recheck trove

kiwik:
recheck trove

h00130372:
recheck trove

h00130372:
recheck trove

h00130372:
recheck trove

h00130372:
recheck trove

h00130372:
recheck trove

jtopjian:
recheck trove

jtopjian:
recheck trove

openstack/db_configuration_v1.go
```diff
@@ -0,0 +1,55 @@
...
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/resource"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/db/v1/configurations"
+)
+
+func expandDatabaseConfigurationV1Datastore(rawDatastore []interface{}) configurations.DatastoreOpts {
+	v := rawDatastore[0].(map[string]interface{})
```
ozerovandrei: I wanted to ask do we really need `map[string]interface{}` instead of more strict `map[string]string` given that [DatastoreOpts](https://github.com/gophercloud/gophercloud/blob/master/openstack/db/v1/configurations/requests.go#L23) has only `string` fields and `resourceDatabaseConfigurationV1` schema contains only `string` fields.

But then I read all commit so I wan't to ask the same question about all `expand*` helpers that were introduced in this PR.

Apart from this it LGTM.

jtopjian: Good point. I didn't do any modifications to the code itself, just rearranged everything since I'm not familiar with Trove. I can look into cleaning this up, though -- it should be safe to do.

jtopjian: @ozerovandrei `map[string]interface{}` is coming from Terraform (through `d.GetOk`, etc), so it's not possible to cast to `map[string]string`.

ozerovandrei: Alright, let's leave it as `map[string]interface{}` then.

PR #672, opened 2019-02-12
Switch to go modules

radeksimko:
All dependencies were semi-manually checked (I used some help from https://github.com/radeksimko/go-mod-diff) to make sure they match before and after the migration where possible.

Once this is merged I will send separate set of PRs to:
 - bump gophercloud dependencies to go-mod-aware versions
 - bump Terraform to latest `v0.11` (which is go-mod-aware)

Please run acceptance tests as I have no access to any OpenStack environment.


jtopjian:
@radeksimko OpenLab automatically runs around 2/3s of the total acceptance tests. The rest we have to run in special environments, which I can do, but I don't think it's needed. Given the diffs aren't showing any major Gophercloud changes, I think this is all good.

IMO, this can be merged. I can merge this or if someone from HC should merge it at an appropriate time, that's fine with me. I'll make sure no vendor changes are done between now and then.

jtopjian:
Also, thank you so much for doing this as well as helping with the Gophercloud work.

Issue #676, opened 2019-02-14
openstack_networking_port_v2 documentation inconsistency


hudolejev:
https://www.terraform.io/docs/providers/openstack/r/networking_port_v2.html#ip_address

says

> fixed_ip.ip_address - (Optional) <...> This field will not be populated if it is left blank or omitted.

I'd expect the OpenStack port is created with dynamically allocated IP address if `fixed_ip.ip_address` is left blank.

However if `fixed_ip.ip_address` is left blank Terraform fails with the following error:

```
Error: openstack_networking_port_v2.port: expected fixed_ip.0.ip_address to contain a valid IP, got: 
```

This code fails:

```
resource openstack_networking_port_v2 "port" {
  admin_state_up     = true
  name               = "test-port-fixed-ip-empty"
  network_id         = "<valid-subnet-id>"

  fixed_ip {
    ip_address = ""
    subnet_id  = "<valid-subnet-id>"
  }
}
```

This code works fine:

```
resource openstack_networking_port_v2 "port" {
  admin_state_up     = true
  name               = "test-port-no-fixed-ip"
  network_id         = "<valid-network-id>"

  fixed_ip {
    subnet_id  = "<valid-subnet-id>"
  }
}
```

### Affected versions/resources

 - `openstack_networking_port_v2`
 - Terraform v0.11.11
 - Terraform OpenStack provider v1.15.0, v1.15.1

Note: OpenStack provider v1.10 and v1.14.0 used to work the same way as written in the docs. Our code computes the IP address dynamically and relies on the option to provide the empty `ip_address`.

jtopjian:
@hudolejev It looks like a new validation which was added to `fixed_ip` is causing this problem. I'll get this reverted.

jtopjian:
@hudolejev This has been fixed in #678. There will be a new release later today or tomorrow.

I apologize for the inconvenience.

hudolejev:
v1.16.0 seems working fine; thanks!

PR #678, opened 2019-02-14
Networking v2: Revert validation of fixed_ip ip_address

jtopjian:
For #676


kayrus:
:man_facepalming:

jtopjian:
@kayrus It's not a big deal - it was worth a shot. :)

Issue #686, opened 2019-02-21
Proposal: Split auth logic into its own module/repository


radeksimko:
### Context

This provider (`terraform-providers/terraform-provider-openstack`) depends on `hashicorp/terraform` (like any other), because it consumes `helper/schema`. `hashicorp/terraform` also depends on `terraform-providers/terraform-provider-openstack` because it shares some auth logic in the Swift remote backend.

https://github.com/hashicorp/terraform/blob/master/backend/remote-state/swift/backend.go

This dependency cycle:

 - can cause confusion for anyone unaware (which is most people except handful of maintainers in these two repositories)
 - can cause version conflicts
    - all providers depend on `hashicorp/terraform` and therefore (as a result of the above) all providers depend on the openstack provider, so any version constraints of dependencies of this provider will constrain dependencies of other providers
 - accounts for unnecessary vendored dead code from transitive dependencies of the provider (basically everything that's unrelated to Swift) in both Terraform and all providers

### Shared auth logic

It is natural and right to have shared auth logic and not having to repeat the same code in two places. 👍 In fact we used to do the same thing with the AWS provider and S3 backend until very recently.
For the same reasons as described above though we decided to create a separate library:

https://github.com/hashicorp/aws-sdk-go-base
https://github.com/hashicorp/terraform/pull/20374 (that PR alone removed `1,3M` LOC)

### Proposal

We (_speaking on behalf of Terraform maintainers_) would love to see a similar library created for Swift/OpenStack-based auth, or any shared logic there is (and needs to be) between the Swift provider and OpenStack provider.

----

cc @jtopjian

jtopjian:
@radeksimko I agree.

To expand on the details, the swift backend is consuming the `config` struct `LoadAndValidate` method: 

* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L19
* https://github.com/hashicorp/terraform/blob/26174c814eaa38bef3de20795bb4bc24e0de16b6/backend/remote-state/swift/backend.go#L299
* https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L57
* https://github.com/hashicorp/terraform/blob/26174c814eaa38bef3de20795bb4bc24e0de16b6/backend/remote-state/swift/backend.go#L326

[...]

PR #688, opened 2019-02-23
Internal: Decouple Authentication

jtopjian:
This commit moves the core authentication logic to a separate package
housed under gophercloud/utils. This is to help the Terraform Swift
backend remove its dependency of this provider.

For #686


jtopjian:
@ozerovandrei It touches a lot of files, but other than `config.go`, it's just replacing lower-case methods with upper-case :)

Thanks! I'm going to hold off on merging this until the core side is ready.

ozerovandrei:
Closing since it was superseded by https://github.com/terraform-providers/terraform-provider-openstack/pull/930

PR #689, opened 2019-02-23
Networking V2: Update port DHCP options in one API call

kayrus:
This improvement will simplify further changes for newer port DNS (597) and port bindings extensions (665). Now port properties and DHCP options can be updated in one call.

I removed `expandNetworkingPortDHCPOptsV2Delete` func and embedded its logic into the `expandNetworkingPortDHCPOptsV2Update`. Using Sets, we can identify which options should be removed, and which options should be added, and combine this data into one `finalUpdateOpts`.

@ozerovandrei, @jtopjian please review.


kayrus:
recheck

kayrus:
recheck

kayrus:
@ozerovandrei @jtopjian , kindly ping

jtopjian:
@kayrus Ack - it's on my list.

PR #559, opened 2018-12-27
[DNM] Testing v0.12 Compatibility

jtopjian:
govendor fetch github.com/hashicorp/terraform/...@c753df6a933ef25197101664b1d7cbb5d115701d
govendor fetch github.com/hashicorp/hcl2/...
govendor fetch github.com/golang/protobuf/...@v1.2.0
govendor fetch github.com/zclconf/go-cty/...
govendor fetch github.com/hashicorp/go-plugin/...
govendor fetch github.com/hashicorp/go-version/...

PR #693, opened 2019-02-25
Networking V2: Introduce port binding support

### LINKED ISSUE #665: Networking V2: add port bindings support
This support would be appreciated for cases, when you need to create a custom port and attach it to the instance.

### Affected Resource(s)

- openstack_networking_port_v2 resource and data source

### References

https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/portsbinding

kayrus:
Depends on #689 and https://github.com/gophercloud/gophercloud/pull/1478
Resolves #665


kayrus:
@jtopjian is there a way to run tests as an admin? Looks like the port binding is not available in openlab

```json
{
  "NeutronError": {
    "message": "((rule:create_port and rule:create_port:binding:profile) and rule:create_port:fixed_ips) is disallowed by policy",
    "type": "PolicyNotAuthorized",
    "detail": ""
  }
}
```

and

```json
{
  "NeutronError": {
    "message": "(((rule:update_port and rule:update_port:binding:host_id) and rule:update_port:binding:profile) and rule:update_port:binding:vnic_type) is disallowed by policy",
    "type": "PolicyNotAuthorized",
    "detail": ""
  }
}
```

If I just add  the `testAccPreCheckAdminOnly(t)`, the tests will be skipped.

jtopjian:
> @jtopjian is there a way to run tests as an admin?

Not at this time. It might require changes to several test fixtures.

> If I just add the `testAccPreCheckAdminOnly(t)`, the tests will be skipped.

That's fine - the tests will need to be run outside of openlab, then.

kayrus:
@jtopjian @ozerovandrei, ready for review. Acceptance tests cannot be executed in openlab env, thus please try to run tests in your private env.

Since this PR is based on #689, you can take a look on the https://github.com/terraform-providers/terraform-provider-openstack/pull/693/commits/06fffe69b8133c169fc74d00a6f60abdaca8954c commit to simplify the code review.

pkqsun:
Hi @kayrus ,
Thanks for your effort working on this.
Does the new 1.17.0 version of terraform-provider-openstack will include this code merge ?

kayrus:
@jtopjian kindly ping

openstack/networking_port_v2.go
```diff
@@ -180,3 +189,139 @@ func expandNetworkingPortFixedIPToStringSlice(fixedIPs []ports.IP) []string {
...
+
+	return nil
+}
+
+func suppressDiffPortBindingProfileV2(k, old, new string, d *schema.ResourceData) bool {
+	if old == "{}" && new == "" {
+		return true
+	}
+
+	if old == "" && new == "{}" {
```
ozerovandrei: I guess that we need to remove the duplicate condition?

openstack/resource_openstack_networking_port_v2.go
```diff
@@ -263,47 +306,50 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	binding := d.Get("binding").([]interface{})
+	err = expandNetworkingPortBindingProfileCreateV2(binding, &finalCreateOpts)
```
jtopjian: I'd prefer createOpts to be handled inline of the `Create` function like all other resources. You can certainly use a function to help build the values for `createOpts`, but I'd like to see at least this inline:

```go
*opts = portsbinding.CreateOptsExt{
	CreateOptsBuilder: *opts,
	HostID:            binding["host_id"].(string),
	Profile:           profile,
	VNICType:          binding["vnic_type"].(string),
}
```

There is not a technical reason for this request. It's just to follow consistency with the code style of the rest of the resources.

openstack/resource_openstack_networking_port_v2.go
```diff
@@ -451,9 +495,20 @@ func resourceNetworkingPortV2Update(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	// Next, perform port binding option changes.
+	if d.HasChange("binding") {
+		hasChange = true
+
+		binding := d.Get("binding").([]interface{})
+		err = expandNetworkingPortBindingProfileUpdateV2(binding, &finalUpdateOpts)
```
jtopjian: Same comment here as createOpts.

openstack/networking_port_v2.go
```diff
@@ -180,3 +189,139 @@ func expandNetworkingPortFixedIPToStringSlice(fixedIPs []ports.IP) []string {
 
 	return s
 }
+
+func flattenNetworkingPortBindingV2(port portExtended, dataSource bool) interface{} {
+	var portBinding []map[string]interface{}
+	var profile interface{}
+
+	if dataSource {
```
jtopjian: So the data source's `profile` has a semi-usable data structure and the resource's `profile` has a JSON structure embedded in the string.

I understand the reasoning behind this, but it's not a consistent experience for the user. In addition, it would encourage the use-case of defining a port with the resource and then immediately using the data source to _kind of_ parse the `profile`.

I'd prefer just storing the JSON string in anticipation that Terraform will eventually support a `jsondecode`-like function. 

I can understand that this will cause the "profile" part of this feature to not be as usable as it could. But imagine if Terraform came out with `jsondecode` next week. Then it wouldn't be an issue at all to store the JSON string.

kayrus: I cannot wait for the new Terraform release, my aim is to combine multiple existing port binding profiles into on profile for the bond vlan, and if I don't store this data in a map, the particular use case, that actually was the reason for the port binding implementation, won't make sense.

What if I create and additional attribute `profile_map` for the data source?

jtopjian: I'm sorry, but this is too much of a hack. You know I'm open to supporting non-standard ways of doing things, but this is one case that I'm not comfortable with. The main reason is because this implementation would become moot the moment a function like `jsondecode` was released, yet we would still need to support and maintain the original implementation for some time.

There are a lot of other providers that are storing JSON as strings. I recommend researching how other users are parsing these strings in order to use the data in a more structured form.

The way the resource is implemented is correct. If you'd like, you can name the argument`profile_json` in anticipation for a future ability for `profile` to handle a more structured set of data (Naming arguments with `_json` is a pattern done elsewhere for these cases). 

The data source needs to be implemented the same way.

kayrus:
recheck

kayrus:
@jtopjian could you please take a look on these weird errors? they are not related to the port binding

jtopjian:
@kayrus Yes. This looks similar to what I'm seeing in #559. I'll try to figure out the root cause.

jtopjian:
@kayrus Can you rebase and try again? I think the subnet/port errors were either a bad devstack commit or a transient error. However, I found a separate failing test which has been fixed (https://github.com/terraform-providers/terraform-provider-openstack/pull/711).

kayrus:
I've no idea what is wrong. Besides port binding tests are not executed in OpenLab environment.

jtopjian:
I'm a little confused by this, but it turns out https://github.com/gophercloud/gophercloud/pull/1498 is the cause of the problem.

```diff
diff --git a/openstack/resource_openstack_networking_subnet_v2.go b/openstack/resource_openstack_networking_subnet_v2.go
index 3e55daf..e669d02 100644
--- a/openstack/resource_openstack_networking_subnet_v2.go
+++ b/openstack/resource_openstack_networking_subnet_v2.go
@@ -508,10 +508,8 @@ func waitForSubnetDelete(networkingClient *gophercloud.ServiceClient, subnetId s
                                log.Printf("[DEBUG] Successfully deleted OpenStack Subnet %s", subnetId)
                                return s, "DELETED", nil
                        }
-                       if errCode, ok := err.(gophercloud.ErrUnexpectedResponseCode); ok {
-                               if errCode.Actual == 409 {
-                                       return s, "ACTIVE", nil
-                               }
+                       if _, ok := err.(gophercloud.ErrDefault409); ok {
+                               return s, "ACTIVE", nil
                        }
                        return s, "ACTIVE", err
                }
```

I have no idea why casting to `gophercloud.ErrUnexpectedResponseCode` no longer works. It should, right? This means that `checkForRetryableError` with the 503 has never worked.

Looks like we'll need to get this patched up. I can work on a PR for this tonight.

kayrus:
@jtopjian ready

PR #695, opened 2019-02-25
Networking V2: Handle multiple dhcp options in port data source

### LINKED ISSUE #694: Networking V2: port data source doesn't print all extra_dhcp_option
### Affected data source(s)

If the port has multiple `extra_dhcp_option` configured, the data source will handle only the latest one.

### References
This bug was initially found in #626

kayrus:
Resolves #694

### DIFF
--- a/openstack/data_source_openstack_networking_port_v2.go
+++ b/openstack/data_source_openstack_networking_port_v2.go
@@ -140,7 +140,7 @@ func dataSourceNetworkingPortV2() *schema.Resource {
 			},
 
 			"extra_dhcp_option": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Computed: true,
 				Elem: &schema.Resource{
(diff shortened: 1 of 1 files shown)


kayrus:
@jtopjian the original issue in aws provider is still not answered. can you take a look on this?

kayrus:
@jtopjian kindly ping

jtopjian:
I asked about this internally and the options are to either proceed with the hashing or change it to a `TypeList`. Changing Set to List is a breaking change, but it would be fine since this is more of a bug fix.

The idea behind using `TypeList` is that since Set hashes aren't predictable or able to be easily referenced, _and_ since data sources are regenerated upon each `apply`, a Lists are more friendly to use.

IMO, let's go with the `TypeList` solution. I understand that this deviates from the resource version, but I always prefer lists over sets because they're more usable for users. Open to discussion, though.

PR #710, opened 2019-03-18
add multi-vm with corresponding floating ip example

erickeller:
This is an example how to spawn multiple openstack vm and automatically adding coresponding floating ips.

### DIFF
--- a/examples/multiple-vm-with-floating-ip/Readme.md
+++ b/examples/multiple-vm-with-floating-ip/Readme.md
@@ -1,23 +1,18 @@
-# Openstack deploy with Terraform
+# Multiple Instances with Floating IPs
 
-## Prerequists
+## Prerequisites
 
-Install Terraform
-
-```
-LATEST_VERSION=$(curl -sL https://releases.hashicorp.com/terraform/ | grep href | grep -v "\(alpha\|beta\|rc\)" | head -n2 | tail -n1 | sed -e 's#.*">.*_\(.*\)<.*#\1#g')
-wget https://releases.hashicorp.com/terraform/${LATEST_VERSION}/terraform_${LATEST_VERSION}_linux_amd64.zip
-sudo unzip terraform_${LATEST_VERSION}_linux_amd64.zip -d /usr/local/bin/
-```
+We assume you already installed terraform using the following instructions: https://www.terraform.io/downloads.html
 
 ## Getting started
 
-Source openstack configuration
+First, download an openrc file from the OpenStack dashboard. If your cloud provider does not support downloading the openrc file from a dashboard, then you can use the openrc.sample file in this directory. Rename it to openrc and modify it using a text editor of your choice.
+
+After you have modified it, source it into your shell environment:
 
 ```
-source ./my-project-openrc.sh
+source ./openrc
 ```
-Note: edit `openrc.sample` file
 
 Init providers
 
@@ -25,7 +20,10 @@ Init providers
 terraform init
 ```
 
-Plan for `n` instances
+
+## Deploy the Configuration
+
+plan the deployment
 
 ```
 terraform plan
@@ -53,7 +51,7 @@ terraform apply -auto-approve \
 
 ## Tear down
 
-cleanup
+cleanup, watch the `-force` will not ask for any confirmation.
 
 ```
 terraform destroy -force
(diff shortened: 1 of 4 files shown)


examples/multiple-vm-with-floating-ip/Readme.md
```diff
@@ -0,0 +1,67 @@
+# Openstack deploy with Terraform
```
jtopjian: Maybe "Multiple Instances with Floating IPs"?

examples/multiple-vm-with-floating-ip/Readme.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Prerequists
```
jtopjian: Prerequisites

examples/multiple-vm-with-floating-ip/Readme.md
```diff
@@ -0,0 +1,67 @@
...
+
+Install Terraform
```
jtopjian: IMO, we should either remove this section and assume the user already has Terraform installed or just link to this page: https://www.terraform.io/downloads.html

examples/multiple-vm-with-floating-ip/Readme.md
```diff
@@ -0,0 +1,67 @@
...
+
+```
+LATEST_VERSION=$(curl -sL https://releases.hashicorp.com/terraform/ | grep href | grep -v "\(alpha\|beta\|rc\)" | head -n2 | tail -n1 | sed -e 's#.*">.*_\(.*\)<.*#\1#g')
+wget https://releases.hashicorp.com/terraform/${LATEST_VERSION}/terraform_${LATEST_VERSION}_linux_amd64.zip
+sudo unzip terraform_${LATEST_VERSION}_linux_amd64.zip -d /usr/local/bin/
+```
+
+## Getting started
```
jtopjian: IMO, let's reword this section to be something like:

First, download an openrc file from the OpenStack dashboard. If your cloud provider does not support downloading the `openrc` file from a dashboard, then you can use the `openrc.sample` file in this directory. Rename it to `openrc` and modify it using a text editor of your choice.

After you have modified it, source it into your shell environment:

```
source openrc
```

examples/multiple-vm-with-floating-ip/Readme.md
```diff
@@ -0,0 +1,67 @@
...
+```
+Note: edit `openrc.sample` file
+
+Init providers
+
+```
+terraform init
+```
+
+Plan for `n` instances
```
jtopjian: This might work better under a new heading titled

```
## Deploy the Configuration
```

examples/multiple-vm-with-floating-ip/main.tf
```diff
@@ -0,0 +1,27 @@
...
+resource "openstack_networking_floatingip_v2" "fip" {
+  count = "${var.count}"
+  pool = "${var.pool}"
+}
+
+resource "openstack_compute_floatingip_associate_v2" "fip" {
+  count = "${var.count}"
+  floating_ip = "${element(openstack_networking_floatingip_v2.fip.*.address, count.index)}"
+  instance_id = "${element(openstack_compute_instance_v2.multi.*.id, count.index)}"
+}
```
jtopjian: A lot of existing examples still use `element`, but it's now better to use an index-based notation:

```hcl
resource "openstack_compute_floatingip_associate_v2" "fip" {
  count       = "${var.count}"
  instance_id = "${openstack_compute_instance_v2.multi.*.id[count.index]}"
  floating_ip = "${openstack_networking_floatingip_v2.fip.*.address[count.index]}
}
```

erickeller: Thanks for all the good comments, I updated the PR with 2 new commit which can be squashed before merge.

maybe you also have a good tipp for passing the name as variable in the following line, so it's not hardcoded to `multi`

```
name = "${format("multi-%02d", count.index+1)}"
```
Regards

jtopjian: You could add a new variable called "instance_prefix" and then do:

```
name = "${format("${var.instance_prefix}-%02d", count.index+1}}"
```

That should work - let me know if it doesn't, though.

erickeller: yeah this also worked thanks. Anything else pending for the PR?

PR #712, opened 2019-03-19
Internal: Networking V2 Router cleanup

ozerovandrei:
Cleanup messages and variables definition for the
openstack_networking_router_v2 resource.

Implement helper functions in openstack/networking_router_v2.go with
tests.


jtopjian:
I know this is marked as a Draft, but this LGTM. Just a heads up that this will conflict with https://github.com/terraform-providers/terraform-provider-openstack/pull/628.

ozerovandrei:
Let's leave it until #628 is resolved then.

PR #713, opened 2019-03-20
 Internal: Use ErrDefault409 

jtopjian:
This commit changes all 409 checks to use the new dedicated
gophercloud.ErrDefault409 type.


kiwik:
recheck

kayrus:
recheck

kiwik:
recheck

jtopjian:
I'm going to go ahead and merge this since it's just replacing one type with another. It also fixes some of the test failures happening.

Issue #714, opened 2019-03-21
Add support for cephx access type


mtneug:
The `openstack_sharedfilesystem_share_access_v2` resource is currently missing the `cephx` access type and something like `access_key` as output.

### Terraform Version

```
Terraform v0.11.13
+ provider.openstack v1.16.0
```

### Affected Resource(s)

* `openstack_sharedfilesystem_share_access_v2`

### Terraform Configuration Files

```hcl
resource "openstack_sharedfilesystem_share_access_v2" "example" {
  share_id     = "${openstack_sharedfilesystem_share_v2.share.id}"
  access_type  = "cephx"
  access_level = "rw"
  access_to    = "myuser"
}
```

jtopjian:
@mtneug Thanks!

I have opened #715 to make these changes. However, I don't have access to a Manila environment in order to verify them. Would you be able to test this?

mtneug:
Wow, that was quick.

Yes, I can test it. See comment in your PR.

jtopjian:
Implemented in #715

PR #715, opened 2019-03-22
SharedFilesystem v2: Add cephx and access_key

jtopjian:
This commit adds "cephx" as a valid access_type. This commit also adds
the access_key computed attribute.

For #714

### DIFF
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -178,7 +178,7 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 			d.Set("region", GetRegion(d, config))
 
 			// This will only be set if the Shared Filesystem environment supports
-			// microversion 2.21 or later.
+			// microversion 2.21.
 			d.Set("access_key", v.AccessKey)
 
 			return nil
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -142,13 +142,20 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 		return fmt.Errorf("Error creating OpenStack sharedfilesystem client: %s", err)
 	}
 
-	sfsClient.Microversion = minManilaMicroversion
-
 	shareID := d.Get("share_id").(string)
 
+	// Attempt to query with 2.21 for environments that support it.
+	// This is so we can retrieve the access_key.
+	sfsClient.Microversion = "2.21"
 	access, err := shares.ListAccessRights(sfsClient, shareID).Extract()
 	if err != nil {
-		return CheckDeleted(d, err, "share_access")
+		log.Printf("[DEBUG] Unable to query openstack_sharedfilesystem_share_access_v2 %s with microversion 2.21", d.Id())
+
+		sfsClient.Microversion = minManilaMicroversion
+		access, err = shares.ListAccessRights(sfsClient, shareID).Extract()
+		if err != nil {
+			return CheckDeleted(d, err, "Error retrieving openstack_sharedfilesystem_share_access_v2")
+		}
 	}
 
 	for _, v := range access {
@@ -158,8 +165,11 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 			d.Set("access_type", v.AccessType)
 			d.Set("access_to", v.AccessTo)
 			d.Set("access_level", v.AccessLevel)
-			d.Set("access_key", v.AccessKey)
 			d.Set("region", GetRegion(d, config))
+
+			if sfsClient.Microversion == "2.21" {
+				d.Set("access_key", v.AccessKey)
+			}
 
 			return nil
 		}
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -143,19 +143,26 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 		return fmt.Errorf("Error creating OpenStack sharedfilesystem client: %s", err)
 	}
 
-	shareID := d.Get("share_id").(string)
+	// Set the client to the minimum supported microversion.
+	sfsClient.Microversion = minManilaMicroversion
 
+	// Now check and see if the OpenStack environment supports microversion 2.21.
+	// If so, use that for the API request for access_key support.
 	apiInfo, err := apiversions.Get(sfsClient, "v2").Extract()
 	if err != nil {
 		return fmt.Errorf("Unable to query Shared Filesystem API endpoint: %s", err)
 	}
 
-	sfsClient.Microversion = minManilaMicroversion
-	if apiInfo.Version != "" {
-		log.Printf("[DEBUG] Setting OpenStack sharedfilesystem client to microversion %s", apiInfo.Version)
-		sfsClient.Microversion = apiInfo.Version
+	compatible, err := compatibleMicroversion("min", "2.21", apiInfo.Version)
+	if err != nil {
+		return fmt.Errorf("Error comparing microversions for openstack_sharedfilesystem_share_access_v2 %s: %s", d.Id(), err)
 	}
 
+	if compatible {
+		sfsClient.Microversion = "2.21"
+	}
+
+	shareID := d.Get("share_id").(string)
 	access, err := shares.ListAccessRights(sfsClient, shareID).Extract()
 	if err != nil {
 		return CheckDeleted(d, err, "Error retrieving openstack_sharedfilesystem_share_access_v2")
@@ -261,13 +268,6 @@ func resourceSharedFilesystemShareAccessV2Import(d *schema.ResourceData, meta in
 			log.Printf("[DEBUG] Retrieved %s share ACL: %#v", accessID, v)
 
 			d.SetId(accessID)
-			d.Set("share_id", shareID)
-			d.Set("access_type", v.AccessType)
-			d.Set("access_to", v.AccessTo)
-			d.Set("access_level", v.AccessLevel)
-			d.Set("access_key", v.AccessKey)
-			d.Set("region", GetRegion(d, config))
-
 			return []*schema.ResourceData{d}, nil
 		}
 	}
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -85,11 +85,15 @@ func resourceSharedFilesystemShareAccessV2Create(d *schema.ResourceData, meta in
 	}
 
 	sfsClient.Microversion = minManilaMicroversion
+	accessType := d.Get("access_type").(string)
+	if accessType == "cephx" {
+		sfsClient.Microversion = "2.13"
+	}
 
 	shareID := d.Get("share_id").(string)
 
 	grantOpts := shares.GrantAccessOpts{
-		AccessType:  d.Get("access_type").(string),
+		AccessType:  accessType,
 		AccessTo:    d.Get("access_to").(string),
 		AccessLevel: d.Get("access_level").(string),
 	}
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hashicorp/terraform/helper/validation"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/apiversions"
 	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/errors"
 	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
 )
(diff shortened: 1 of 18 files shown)


mtneug:
The provider now supports the new type, however I still see this error message:
```
* openstack_sharedfilesystem_share_access_v2.example: Error granting access: badRequest (400): Only 'ip', 'user' or 'cert' access types are supported.
```
I can definitely create access rules with the `manila` CLI tool:
```
$ manila access-allow --access-level rw share cephx myuser
+--------------+--------------------------------------+
| Property     | Value                                |
+--------------+--------------------------------------+
| access_key   | None                                 |
| share_id     | ...                                  |
| created_at   | ...                                  |
| updated_at   | None                                 |
| access_type  | cephx                                |
| access_to    | myuser                               |
| access_level | rw                                   |
| state        | queued_to_apply                      |
| id           | ...                                  |
| metadata     | {}                                   |
+--------------+--------------------------------------+
```

jtopjian:
> openstack_sharedfilesystem_share_access_v2.example: Error granting access: badRequest (400): Only 'ip', 'user' or 'cert' access types are supported.

That looks like an API error: https://github.com/terraform-providers/terraform-provider-openstack/blob/25fb29c43cca5954045580310b6c1acd4977f6fa/openstack/resource_openstack_sharedfilesystem_share_access_v2.go#L115

Can you run 

```
manila --debug access-allow --access-level rw share cephx myuser
```

And let me know what the actual request body is?

kayrus:
@jtopjian Cephx was added in `2.13 - Add "cephx" auth type to allow_access` microversion. Probably you have to explicitly set this microversion before sending the request body. At the moment the `minManilaMicroversion   = "2.7"`.

jtopjian:
> Probably you have to explicitly set this microversion before sending the request body. At the moment the minManilaMicroversion = "2.7"

I think we'll need to allow users to set the microversion, then.

kayrus:
https://github.com/openstack/manila/blob/a4a0f16a376033f468236844eed756be35fd536f/manila/api/common.py#L328..L342
`access_to` should not be an IP address. BTW, does openlab support ceph?

jtopjian:
@mtneug Would you be able to test with the latest commit of this PR when you have time?

mtneug:
@jtopjian I'm sorry for the delay.

Terraform now successfully creates the access rule. However, the returned `access_key` is empty (for new resources and also when importing).

kayrus:
I guess an extra microversion is required inside the Read func:

> access_key - The access credential of the entity granted share access. New in version 2.21

https://developer.openstack.org/api-ref/shared-file-system/?expanded=grant-access-detail#grant-access

jtopjian:
>  guess an extra microversion is required inside the Read func:

hah - thanks for that.

jtopjian:
This poses an interesting problem. Microversion 2.13 is supported in Mitaka, which is the unofficial minimum version Gophercloud supports.

2.21 is Newton.

So there's a chance that setting the Microversion to 2.21 during Read will break things. I'm going to attempt to do a second request just for `access_key` and silently pass if there's an error. Unfortunately it'll be hard for me to verify it works, so these notes are more for posterity.

kayrus:
@jtopjian probably it's worth to add check whether there is a `cephx` access_type set, and set the microversion correspondingly?

jtopjian:
> probably it's worth to add check whether there is a cephx access_type set, and set the microversion correspondingly?

I thought the same, but cephx is supported in mitaka. So in theory, someone could use cephx but not use the access key :/

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -150,6 +167,10 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 			d.Set("access_level", v.AccessLevel)
 			d.Set("region", GetRegion(d, config))
 
+			if sfsClient.Microversion == "2.21" {
```
jtopjian: Technically this isn't needed. If `minManilaMicroversion` was used, then `access_key` would be an empty string. However, this line serves more as internal documentation.

jtopjian:
@mtneug Thank you for your patiences 🙂 When you have time, can you test again?

jtopjian:
recheck

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
kayrus: If there is no `access_key` and there are dependencies, which use it, terraform will fail. Let's just keep `d.Set("access_key", v.AccessKey)`.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -142,13 +142,20 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 		return fmt.Errorf("Error creating OpenStack sharedfilesystem client: %s", err)
 	}
 
-	sfsClient.Microversion = minManilaMicroversion
-
 	shareID := d.Get("share_id").(string)
 
+	// Attempt to query with 2.21 for environments that support it.
+	// This is so we can retrieve the access_key.
+	sfsClient.Microversion = "2.21"
```
kayrus: It's worth using the https://godoc.org/github.com/gophercloud/gophercloud/openstack/sharedfilesystems/apiversions

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
jtopjian: I don't follow. It'll still be an empty string regardless.

jtopjian: The latest commit reverts this, but I still think it would have worked otherwise.

kayrus: Last time I got an error when I tried to access an attribute, which was not set. Probably I'm wrong...

jtopjian:
recheck

jtopjian:
recheck

mtneug:
@jtopjian now I can't get any state at all for `openstack_sharedfilesystem_share_access_v2` resources (using `terraform state show`). Import also fails saying the resource with given ID doesn't exists.

jtopjian:
@mtneug Can you run the following:

```
OS_DEBUG=1 TF_LOG=DEBUG terraform apply (or import)
```

And post a clean copy of the logs to a Github gist?

jtopjian:
Actually, the import issue is weird since it has not been modified yet. It should be working as it previously was.

jtopjian:
recheck

jtopjian:
@mtneug Actually, test results just came in and I see the error. I'll have to look into this a little more. This PR has sure escalated quickly.

jtopjian:
The local vendor change I made was a success. I'll be opening a PR in Gophercloud to get this merged properly:

```
2019/03/27 20:32:10 [DEBUG] OpenStack Request URL: GET http://192.168.1.5/share/v2/
2019/03/27 20:32:10 [DEBUG] OpenStack Request Headers:
Accept: application/json
Referer: http://192.168.1.5/share/v2
User-Agent: Terraform/0.11.12 gophercloud/2.0.0
X-Auth-Token: ***
2019/03/27 20:32:10 [DEBUG] OpenStack Response Code: 200
2019/03/27 20:32:10 [DEBUG] OpenStack Response Headers:
Content-Length: 383
Date: Wed, 27 Mar 2019 20:32:10 GMT
Server: Apache/2.4.18 (Ubuntu)
Vary: X-OpenStack-Manila-API-Version
X-Compute-Request-Id: req-0c90900e-572b-4f39-9742-0019d0a56ce8
X-Openstack-Manila-Api-Version: 2.0
2019/03/27 20:32:10 [DEBUG] OpenStack Response Body: {
  "versions": [
    {
      "id": "v2.0",
      "links": [
        {
          "href": "http://docs.openstack.org/",
          "rel": "describedby",
          "type": "text/html"
        },
        {
          "href": "http://192.168.1.5/share/v2/",
          "rel": "self"
        }
      ],
      "media-types": [
        {
          "base": "application/json",
          "type": "application/vnd.openstack.share+json;version=1"
        }
      ],
      "min_version": "2.0",
      "status": "CURRENT",
      "updated": "2015-08-27T11:33:21Z",
      "version": "2.49"
    }
  ]
}
2019/03/27 20:32:10 [DEBUG] Setting OpenStack sharedfilesystem client to microversion 2.49
```

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
jtopjian: Something was nagging me about this. I knew there was a reason I've opted not to use `apiversions` for microversion solutions and this morning I remembered: microversions can also _remove_ features.

Setting a client to the latest version could actually turn other things off.

The best solution is to poll for which versions the API service supports and then make individual API requests with the microversion set to specific versions which are known working for the feature being requested.

Granted, for the specific `access_key` feature in this PR, this situation might not ever happen, but I'll try to do it for this PR so there's an example in the repository. However, since I'm relying on OpenLab for test results, it might take a few tries.

kayrus:
@jtopjian just in case, how it is done in https://github.com/kubernetes/cloud-provider-openstack/blob/b100f24bb0e4b70c71587e24cac6ae84ddfa77d4/pkg/share/manila/client.go#L140

jtopjian:
Ah - that's _very_ close to the implementation I have worked out. Good to know I'm on the right track. Thanks!

jtopjian:
The force push was to remove my local vendor changes. The three additional commits are the new commits and can be viewed individually.

jtopjian:
@mtneug OK, I think this is ready for another test at your convenience :)

mtneug:
Import and create are now working correctly and the `access_key` output is available. Anything else I can test?

jtopjian:
> Import and create are now working correctly and the access_key output is available.

Hooray 🎉 

> Anything else I can test?

Nope. Thank you _so_ much for your patiences and for confirming functionality. I'll let the others do a code review and we'll get this merged.

mtneug:
@jtopjian I have to thank 🙂!

jtopjian:
@kayrus I'm not sure if Github notified you of a re-review, but let me know if you have any additional comments. Thank you for your help with the microversions.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -79,11 +86,15 @@ func resourceSharedFilesystemShareAccessV2Create(d *schema.ResourceData, meta in
 	}
 
 	sfsClient.Microversion = minManilaMicroversion
+	accessType := d.Get("access_type").(string)
+	if accessType == "cephx" {
+		sfsClient.Microversion = "2.13"
```
kayrus: Shall we fetch supported API versions here as well?

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -150,6 +177,10 @@ func resourceSharedFilesystemShareAccessV2Read(d *schema.ResourceData, meta inte
 			d.Set("access_level", v.AccessLevel)
 			d.Set("region", GetRegion(d, config))
 
+			// This will only be set if the Shared Filesystem environment supports
+			// microversion 2.21 or later.
```
kayrus: Nit: as we clarified, later versions can obsolete certain functions

website/docs/r/sharedfilesystem_share_access_v2.html.markdown
```diff
@@ -138,6 +142,7 @@ The following arguments are supported:
 * `access_type` - See Argument Reference above.
```
kayrus: `access_type` reference, mentioned above under the `Argument Reference` section, should be:

```
`access_type` - (Required) The access rule type. Can either be an `ip`, `user`, `cert` or `cephx`.
```

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -49,7 +50,7 @@ func resourceSharedFilesystemShareAccessV2() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
-					"ip", "user", "cert",
+					"ip", "user", "cert", "cephx",
 				}, true),
```
kayrus: Let's replace `true` with the `false`. This parameter has to be case sensitive.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
```diff
@@ -67,6 +68,12 @@ func resourceSharedFilesystemShareAccessV2() *schema.Resource {
 					"rw", "ro",
 				}, true),
```
kayrus: Let's replace `true` with the `false`. This parameter has to be case sensitive.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
jtopjian: There's no need. If someone is opting to use `cephx` it implies that they have a supported OpenStack environment. If the OpenStack environment doesn't support 2.13, an error will be returned. In the far future if/when 2.13 is dropped, this might need changed.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
jtopjian: But it's being set to 2.21 explicitly above this.

openstack/resource_openstack_sharedfilesystem_share_access_v2.go
jtopjian: You're correct, but this is a bug outside the scope of this PR. There should be a separate PR to change this to false so it can be noted in the changelog as a bug fix.

The reason being is that this `ValidateFunc` is already deployed in a released version and users might already be using values like `IP`, `USER`, etc and possibly ignoring the diff that happens.

kayrus:
LGTM

openstack/util_test.go
```diff
@@ -27,3 +27,42 @@ func TestExpandToStringSlice(t *testing.T) {
...
 }
+
+func TestCompatibleMicroversion(t *testing.T) {
+	actual, err := compatibleMicroversion("min", "2.1.0", "2.5")
+	assert.NotNil(t, err)
+
+	actual, err = compatibleMicroversion("min", "2.1", "2.5.0")
+	assert.NotNil(t, err)
+
+	actual, err = compatibleMicroversion("minn", "2.1", "2.5")
```
ozerovandrei: `minn` -> `min`

ozerovandrei: Oh, my bad, you're actually testing that `err == nil` here.

PR #718, opened 2019-03-24
Networking V2: ports data source: resolve server id by a name

### LINKED ISSUE #717: Networking V2: ports data source: resolve server id by a name
For instance, user wants to find all the ports, attached to the certain server. In case, when the server was not created by a Terraform, it requires the manual server id resolve. Terraform has neither the compute instance data source, nor possibility to import the compute instance into the Terraform state.

The fastest solution for this use case is to resolve server id by name. There is a `IDFromName` function available in [gophercloud](https://godoc.org/github.com/rackspace/gophercloud/openstack/compute/v2/servers#IDFromName) and it is simple to use it. I suggest to add the same flag, available for openstack CLI: `server`, into the port data sources.

### Affected Data Source(s)
- openstack_networking_port_ids_v2

kayrus:
Resolves #717

### DIFF
--- a/openstack/data_source_openstack_networking_port_ids_v2.go
+++ b/openstack/data_source_openstack_networking_port_ids_v2.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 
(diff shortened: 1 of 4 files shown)


kayrus:
recheck

kayrus:
Won't do

PR #738, opened 2019-04-17
SharedFilesystem v2: Fix access validation rules

jtopjian:
This commit enforces case sensitivity on access_type and access_level.

Per discussion in #715 

/cc @kayrus


kayrus:
LGTM

Issue #729, opened 2019-04-05
Networking V2: multiple allocation pools are not handled correctly for the subnet


kayrus:
### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking_subnet_v2

Since they are defined as a list, they are modified randomly:

```hcl
resource "openstack_networking_subnet_v2" "subnet" {
  name       = "subnet"
  network_id = "${openstack_networking_network_v2.subnet.id}"
  cidr       = "10.3.0.0/16"
  ip_version = 4
  dns_nameservers = []
  allocation_pools = [{ start = "10.3.0.2" end = "10.3.0.255"},{ start = "10.3.255.0" end = "10.3.255.254"},]
}
```

```
one shot:
openstack_networking_subnet_v2.subnet: Modifying... (ID: ebaef4a6-c69e-4000-8bb7-8ed6cf33ee2e)
  allocation_pools.0.end:   "10.3.255.254" => "10.3.0.255"
  allocation_pools.0.start: "10.3.255.0" => "10.3.0.2"
  allocation_pools.1.end:   "10.3.0.255" => "10.3.255.254"
  allocation_pools.1.start: "10.3.0.2" => "10.3.255.0"
other shot:
openstack_networking_subnet_v2.subnet: Modifying... (ID: ebaef4a6-c69e-4000-8bb7-8ed6cf33ee2e)
  allocation_pools.0.end:   "10.3.255.254" => "10.3.0.255"
  allocation_pools.0.start: "10.3.255.0" => "10.3.0.2"
  allocation_pools.1.end:   "10.3.0.255" => "10.3.255.254"
  allocation_pools.1.start: "10.3.0.2" => "10.3.255.0"
```

this fix did the trick:

```diff
diff --git a/openstack/resource_openstack_networking_subnet_v2.go b/openstack/resource_openstack_networking_subnet_v2.go
index dd46f63..1c5e11b 100644
--- a/openstack/resource_openstack_networking_subnet_v2.go
+++ b/openstack/resource_openstack_networking_subnet_v2.go
@@ -70,7 +70,7 @@ func resourceNetworkingSubnetV2() *schema.Resource {
                                Computed: true,
                        },
                        "allocation_pools": {
-                               Type:     schema.TypeList,
+                               Type:     schema.TypeSet,
                                Optional: true,
                                Computed: true,
                                Elem: &schema.Resource{
[...]

PR #739, opened 2019-04-17
 Networking v2: Deprecate subnet allocation_pools for allocation_pool

jtopjian:
This commit deprecates the openstack_networking_subnet_v2
allocation_pools argument in favor of allocation_pool.

allocation_pool is a singular word which makes more sense to use
declaratively.

In addition, allocation_pools was originally a TypeList, but the
OpenStack API will not respect the ordering defined by the user for
multiple pools, which triggered unneeded diffs.

allocation_pool is now a TypeSet which accounts for an unordered
response.

For #729 

/cc @kayrus

### DIFF
--- a/openstack/networking_subnet_v2_test.go
+++ b/openstack/networking_subnet_v2_test.go
@@ -1,0 +1,108 @@
+package openstack
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNetworkingSubnetV2AllocationPoolsMatch(t *testing.T) {
+	oldPools := []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
+	newPools := []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
+	same := networkingSubnetV2AllocationPoolsMatch(oldPools, newPools)
+	assert.Equal(t, same, true)
+
+	oldPools = []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+	}
+
+	newPools = []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
+	same = networkingSubnetV2AllocationPoolsMatch(oldPools, newPools)
+	assert.Equal(t, same, false)
+
+	oldPools = []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
+	newPools = []interface{}{
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
+	same = networkingSubnetV2AllocationPoolsMatch(oldPools, newPools)
+	assert.Equal(t, same, false)
+
+	oldPools = []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.10",
+			"end":   "192.168.199.150",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
+	newPools = []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
+		},
+	}
+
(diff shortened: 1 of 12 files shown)


jtopjian:
This one was interesting.

While we could have changed the `allocation_pools` from `TypeList` to `TypeSet`, I also wanted to take the opportunity to correct the naming of `allocation_pools` to `allocation_pool`.

But since we need to support _both_ arguments, a change to `allocation_pool` would still trigger a diff on `allocation_pools`. 

This is where the `CustomizeDiff` comes in. I've been meaning to learn how to use this for a long time and finally had the time tonight. It's pretty powerful and can be used to suppress/clear the diffs of other areas like this.

I'm sure there are some areas for improvement here. I'll wait until after tests pass to request an official review.

openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -431,7 +461,37 @@ func resourceNetworkingSubnetV2Delete(d *schema.ResourceData, meta interface{})
...
+// resourceSubnetAllocationPoolsCreateV2 returns a slice of allocation pools
+// when creating a subnet. It takes into account both the old allocation_pools
+// argument as well as the new allocation_pool argument.
+//
+// This can be modified to only account for allocation_pool when
+// allocation_pools is removed.
+func resourceSubnetAllocationPoolsCreateV2(d *schema.ResourceData) []subnets.AllocationPool {
+	// First check allocation_pools since that is the original,
+	// but deprecated, argument.
+	rawAPs := d.Get("allocation_pools").([]interface{})
```
ozerovandrei: Maybe it's better to check the new one first and fallback to deprecated one only in case the new one is empty?

jtopjian: Good idea.

openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -533,3 +609,43 @@ func waitForSubnetDelete(networkingClient *gophercloud.ServiceClient, subnetId s
...
 	}
 }
+
+func resourceSubnetV2AllocationPoolsCustomizeDiff(diff *schema.ResourceDiff) error {
+	if diff.Id() != "" && diff.HasChange("allocation_pools") {
+		o, n := diff.GetChange("allocation_pools")
+		oRaw := o.([]interface{})
+		nRaw := n.([]interface{})
+
+		samePools := true
```
ozerovandrei: We can simplify this code.

```suggestion
		samePools := false
```

openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -533,3 +609,43 @@ func waitForSubnetDelete(networkingClient *gophercloud.ServiceClient, subnetId s
...
+			newStart := newRawPool["start"].(string)
+			newEnd := newRawPool["end"].(string)
+
+			for _, oldRaw := range oRaw {
+				oldRawPool := oldRaw.(map[string]interface{})
+				oldStart := oldRawPool["start"].(string)
+				oldEnd := oldRawPool["end"].(string)
+
+				if oldStart == newStart && oldEnd == newEnd {
+					found = true
```
ozerovandrei:
```suggestion
					samePools = true
```

openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -533,3 +609,43 @@ func waitForSubnetDelete(networkingClient *gophercloud.ServiceClient, subnetId s
...
+				}
+			}
+
+			if !found {
```
ozerovandrei: In case of applying two previous changes we can remove the extra `found` variable from this function.

openstack/resource_openstack_networking_subnet_v2.go
jtopjian: hm... I'm not sure, but I could be wrong.

In the event that there are multiple allocation pools, setting `samePools` to `true` here would say that all pools are the same when only 1 pool matches.

My intention here is that `samePools` signifies that _all_ allocation pools are the same. `found` means that one of the pools was found. If a minimum of one pool was not found, then the two (old and new) aren't the same and we must issue a diff.

I added a new commit which breaks this code out into a dedicated function so we can unit test it.

openstack/networking_subnet_v2_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	newPools = []interface{}{
+		map[string]interface{}{
+			"start": "192.168.199.2",
+			"end":   "192.168.199.100",
+		},
+
+		map[string]interface{}{
+			"start": "10.3.0.1",
+			"end":   "10.3.0.100",
```
jtopjian: Ignore that this fixture is actually incorrect: it's not possible to have a 192 subnet and 10 subnet mixed. But they're used for visibility.

kiwik:
recheck

jtopjian:
recheck

h00130372:
recheck stable/stein

jtopjian:
OK, I think this is ready for (another) review. @kayrus any thoughts?

openstack/resource_openstack_networking_subnet_v2.go
```diff
@@ -70,7 +71,25 @@ func resourceNetworkingSubnetV2() *schema.Resource {
 				Computed: true,
 			},
 			"allocation_pools": {
-				Type:     schema.TypeList,
+				Type:       schema.TypeList,
+				Optional:   true,
+				Computed:   true,
+				Deprecated: "use allocation_pool instead",
```
kayrus: Please also add `ConflictsWith` in order to avoid double definition of the pools.

jtopjian: I'll double-check if this works. My understanding was that `ConflictsWith` only worked with strings... But it's been a while and maybe I was originally trying to do something else.

jtopjian: I was indeed mistaken. Works just fine :)

kayrus:
LGTM

jtopjian:
> Sorry for the long delay.

What delay? :)

Issue #364, opened 2018-07-03
Need resource for neutron address scopes


kschroed:
We are currently using Cisco ACI with the Neutron ACI Integration Module(AIM) and leveraging the https://github.com/kubernetes-incubator/kubespray for deploying K8S clusters on top of Openstack.  In order to stream line creating multiple K8S clusters under one Openstack project we would like to leverage the address-scope resource in neutron which currently does not have a terraform resource definition.  The terraform openstack_networking_subnetpool_v2 seems to accept address_scope_id as a valid argument but there is no resource avaialble by which to create address scopes. 

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.
Terraform v0.11.3
+ provider.null v1.0.0
+ provider.openstack v1.3.0

### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking

ozerovandrei:
Hello @kschroed.

Address-scopes resource was added in `1.15.0 (February 06, 2019)` release.
Data source will be included in the future `1.18.0` release.

PR #741, opened 2019-04-25
Networking V2: add AddressScope datasource

ozerovandrei:
Add dataSourceNetworkingAddressScopeV2 with tests and documentation.

For #364

### DIFF
--- a/openstack/data_source_openstack_networking_addressscope_v2.go
+++ b/openstack/data_source_openstack_networking_addressscope_v2.go
@@ -59,7 +59,7 @@ func dataSourceNetworkingAddressScopeV2Read(d *schema.ResourceData, meta interfa
 		listOpts.IPVersion = v.(int)
 	}
 
-	if v, ok := d.GetOk("shared"); ok {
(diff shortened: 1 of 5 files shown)


ozerovandrei:
recheck

openstack/data_source_openstack_networking_addressscope_v2.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if v, ok := d.GetOk("name"); ok {
+		listOpts.Name = v.(string)
+	}
+
+	if v, ok := d.GetOk("ip_version"); ok {
+		listOpts.IPVersion = v.(int)
+	}
+
+	if v, ok := d.GetOk("shared"); ok {
```
jtopjian: `GetOk` will only return `true` if the value is not the zero-value. This block of code will only be triggered if the user specifies `true`. If you want them to be able to specify `false`, use `GetOkExists`.

ozerovandrei:
recheck

PR #748, opened 2019-05-02
Networking v2: Make port dns_name computed

jtopjian:
The OpenStack Compute service will set a port's dns_name to the name of
the instance when the OpenStack environment has the DNS extension
enabled. Because of this behavior, the dns_name needs to be set to
Computed or else the dns_name will be changed to an empty string on the
next apply.

For #742

### DIFF
--- a/openstack/resource_openstack_networking_port_v2.go
+++ b/openstack/resource_openstack_networking_port_v2.go
@@ -256,6 +256,7 @@ func resourceNetworkingPortV2() *schema.Resource {
 			"dns_name": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 
 			"dns_assignment": {


jtopjian:
I _think_ this is the only DNS change that is needed.

I tested the DNS fields in `openstack_networking_floatingip_v2` and was unable to trigger a diff / force recreation. However, I don't have access to an OpenStack environment that can fully utilize the DNS extension.

If we get reports about that resource, then we can make the appropriate changes as a "minor" (in semver terms) bug fix since it's an API interaction bug.

Also /cc @kayrus. My comment in #742 still applies: I appreciate you questioning the correct solution for this and if you have objections, please let me know.

kayrus:
@jtopjian thanks for taking care of this, LGTM

PR #750, opened 2019-05-07
Make timeout for stop_before_destroy configurable

rnurgaliyev:
Some machines require more time to do graceful shutdown. Timeout is now hardcoded to three minutes, this PR makes this value configurable.

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -897,7 +897,7 @@ func resourceComputeInstanceV2Delete(d *schema.ResourceData, meta interface{}) e
 				Pending:    []string{"ACTIVE"},
 				Target:     []string{"SHUTOFF"},
 				Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
-				Timeout:    3 * time.Minute,
+				Timeout:    d.Timeout(schema.TimeoutDelete),
 				Delay:      10 * time.Second,
 				MinTimeout: 3 * time.Second,
 			}


openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -897,7 +902,7 @@ func resourceComputeInstanceV2Delete(d *schema.ResourceData, meta interface{}) e
 				Pending:    []string{"ACTIVE"},
 				Target:     []string{"SHUTOFF"},
 				Refresh:    ServerV2StateRefreshFunc(computeClient, d.Id()),
-				Timeout:    3 * time.Minute,
+				Timeout:    time.Duration(d.Get("stop_before_destroy_timeout").(int)) * time.Minute,
```
jtopjian: I'm hesitant to add a new argument of `*_timeout`. Can we instead use the built-in timeout? 

https://github.com/terraform-providers/terraform-provider-openstack/blob/f440b7db1fb47ae1e46dff1ac80415773fcc4d36/openstack/resource_openstack_compute_instance_v2.go#L39

rnurgaliyev: Agreed, this makes sense. I will update this PR soon!

rnurgaliyev: But in which built-in timeout this should be configured? 
```
type ResourceTimeout struct {
	Create, Read, Update, Delete, Default *time.Duration
}
```

jtopjian: I think we can just re-use "Delete" since this is happening during the delete phase. You can use it the same way as here: https://github.com/terraform-providers/terraform-provider-openstack/blob/f440b7db1fb47ae1e46dff1ac80415773fcc4d36/openstack/resource_openstack_compute_instance_v2.go#L938

rnurgaliyev: Done, now it is a one-liner

Issue #439, opened 2018-10-08
floating_ip_assignment with create_before_destroy


dave-lew99:
Hi there,

Terraform Version
Terraform v0.11.8

Affected Resource(s)
Please list the resources as a list, for example:

openstack_compute_instance_v2
openstack_compute_floatingip_associate_v2
If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

Terraform Configuration Files
```resource "openstack_compute_instance_v2" "test-server" {
    name            = "test-tf"
    flavor_id       = "003386be-0c35-427c-a0cc-0789ebb8f79b" # this is the instance type `openstack flavor list`

    block_device {
        uuid                  = "f61601e5-c9e1-4def-a107-bb082f914a61" # this is the image ID to boot from `openstack image list`
        source_type           = "image"
        volume_size           = 5
        boot_index            = 0
        destination_type      = "volume"
        delete_on_termination = true
    }

    network {
    name = "Public"
    }

    lifecycle {
        create_before_destroy = true
    }
}

resource "openstack_compute_floatingip_associate_v2" "myip" {
    floating_ip = "<public ip>"
    instance_id = "${openstack_compute_instance_v2.test-server.id}"
    lifecycle {
        create_before_destroy = true
    }
}
```

Debug Output
Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the crash.log.

Expected Behavior
When updating an openstack instance with a new image, the new instance should be provisioned then the floating the IP assignment is changed over to the new instance, before the old instance is destroyed.

Actual Behavior
The new instance is created and the floating IP is swapped over to the new instance. The apply then fails with the message 

> openstack_compute_floatingip_associate_v2.myip (deposed #0): floating ip association: Expected HTTP response code [201 202] when accessing [POST https://<openstack url>/compute/v2.1/servers/d93cb36e-fdeb-44d3-9068-0e73e2200dc9/action], but got 409 instead
{"conflictingRequest": {"message": "Floating IP <ip> is not associated with instance d93cb36e-fdeb-44d3-9068-0e73e2200dc9.", "code": 409}}

Steps to Reproduce
Please list the steps required to reproduce the issue, for example:

terraform apply
Change the image value
Run terraform apply
References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

Maybe https://github.com/terraform-providers/terraform-provider-digitalocean/issues/147

[...]

PR #761, opened 2019-06-04
Compute V2: Don't fail when disassociating floating IP from the instance when it is already not associated.

rnurgaliyev:
According to Compute API documentation, an attempt to remove floating IP association from an instance may fail with 409, when floating IP address was not associated with this instance initially. 

This patch enables "create_before_destroy" lifecycle in `openstack_compute_floatingip_associate_v2` resource. When used together with an instance, whose lifecycle also has `create_before_destroy`, this patch will remove downtime in a rolling update - floating IP address will be reassociated only when the new instance is completely provisioned.

Without this patch, setting `create_before_destroy` in lifecycle of `openstack_compute_floatingip_associate_v2` will cause terraform to fail. New VM will be provisioned before deleting the old one, then new floating ip address association will be created, re-assigning same floating IP address from the existing resource, but on the next step terraform will fail while trying to delete old association since it is not there anymore in OpenStack. However, there is no need to fail - resource is actually deleted, and consequent "terraform plan" will show no changes.

This issue is discussed in here: https://github.com/terraform-providers/terraform-provider-openstack/issues/439

### DIFF
--- a/openstack/resource_openstack_compute_floatingip_associate_v2.go
+++ b/openstack/resource_openstack_compute_floatingip_associate_v2.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/floatingips"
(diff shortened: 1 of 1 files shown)


jtopjian:
I think it would be hard to write an acceptance test for this. I have tested this offline and can confirm it works.

Issue #457, opened 2018-10-23
Compute v2: openstack_compute_server_v2 resource


jtopjian:
This issue is to track progress of a future `openstack_compute_server_v2` resource.

This resource will supersede the `openstack_compute_instance_v2` resource, though `openstack_compute_instance_v2` will still exist and will still be supported (although new features will probably not be added).

#### Why?

The `openstack_compute_instance_v2` was one of the first resources made for this provider. It's currently the largest and most complicated resource. After 4 years, it supports a wide range of features (including both nova-network and Neutron compatibility) but also has a lot of code which was implemented early-on to get around various limitations. We feel that modifying this resource to either refactor existing code or to add new features (such as importing) will cause disruption to users, and we don't want that.

Therefore, we're proposing a new resource which users can opt-in to using at their convenience. This new resource won't support certain older features (such as nova-network), so we'll be keeping the original resource around until an OpenStack Provider v2 release (which is nowhere on our radar yet).

We chose `openstack_compute_server_v2` as the name because the unified Python OpenStack SDK refers to "instances" as "servers".

#### When?

There is no ETA for this resource. Other than some discussion, no work has started. We'll update this issue as development progresses.

Comments and input are most welcomed!

jtopjian:
While developing this resource, we should consider how we can resolve issues with security groups being applied to both ports and instances, if possible.

At the moment, combining ports and instances has a [few caveats](https://www.terraform.io/docs/providers/openstack/r/compute_instance_v2.html#instances-and-ports).

PR #763, opened 2019-06-05
Enable import instance

jsm222:
Openstack does not keep track of NIC order after creation
Therefor one must set TF_MAC_IMPORT_ORDER on import in the following format
mac1_mac2P where P tells that it was a specified port on creation time

Example TF_MAC_IMPORT_ORDER='fa:16:3e:cb:c3:88_fa:16:3e:fc:ba:c9P_fa:16:3e:d8:cc:4f' ./terraform import openstack_compute_instance_v2.basic d15f7427-d2ed-4474-a07d-2c7e087ec240

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -20,12 +20,65 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/images"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
 )
 
+func resourceOpenStackComputeInstanceV2ImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	results := make([]*schema.ResourceData, 1)
+	ordered_networks := []interface{}{}
+	config := meta.(*Config)
+	err := resourceComputeInstanceV2Read(d, meta)
+	if err != nil {
+		return nil, err
+	}
+	networkClient, err := config.networkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return nil, fmt.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	for _, mac := range strings.Split(os.Getenv("TF_MAC_IMPORT_ORDER"), "_") {
+		pSpec := false
+		networks := []interface{}{}
+		networks = d.Get("network").([]interface{})
+		for _, n := range networks {
+			v := n.(map[string]interface{})
+			if strings.HasSuffix(mac, "P") {
+				mac = strings.TrimSuffix(mac, "P")
+				pSpec = true
+			}
+
+			if mac == v["mac"] {
+				if pSpec {
+					listOpts := ports.ListOpts{
+						MACAddress: mac,
+					}
+					allPages, err := ports.List(networkClient, listOpts).AllPages()
+					if err != nil {
+						return nil, fmt.Errorf("Unable to retrieve networks from the Network API: %s", err)
+					}
+
+					mPorts, err := ports.ExtractPorts(allPages)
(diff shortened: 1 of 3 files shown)


jtopjian:
> Openstack does not keep track of NIC order after creation

I'm not sure this is correct. Can you elaborate on this more?

jtopjian:
After reviewing the code, I understand what you mean about keeping track of the NIC order, but I don't see why that's necessary here.

In theory, Terraform should be able to read the instance data and set the NICs in the order that the API response sorted them. The user would then have to create a Terraform configuration that declared the nics / networks in the same order.

Ordering doesn't matter here because if a user changes any part of the `network` block, the instance will be recreated.

I think there are two options being presented:

1. Just import the NICs in the order that the API call sends and have the user create the configuration using that order.
2. Implement this `TF_MAC_IMPORT_ORDER` which allows the user to specify the order of the NICs.

It's six one way, half a dozen the other. If we were to look at merging support for Instance Importing, I'd prefer the former approach since it involves less changes upfront and doesn't require the use of an environment variable to assist with importing.

Please let me know if I'm misunderstanding something here, though.

jsm222:
I agree that [1] is cleaner. One should just manually add port id to terraform.tfstate if it is an external specified port resource, which could be noted in import documentation.
In [1] the CracyNic test fails, so the test plan must somehow be changed, or CrazyNics should just not be a test plan, since we know why it fails..  
A cinder bootet instance should be part of a test plan I think it fails at the moment, because of trouble setting the image name. I'll look into that, and hopefully come up with a modifed PR soon.

jtopjian:
This all sounds good. If you haven't determined why the CrazyNics test is failing, let me know and I'll dig into it.

If you can get this to work (with the understanding that the user must use the same ordering of NICs), I'm in favor of merging it. 

One thing to keep in mind is that we do have plans to introduce a new server resource (https://github.com/terraform-providers/terraform-provider-openstack/issues/457) in which we'll be able to use different API calls which weren't available in Gophercloud at the time the instance resource was initially created.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
```
jtopjian: We shouldn't mix API services if we can avoid it.

In this case, Block Storage v3 API is being imported, but this will now require anyone using `openstack_compute_instance_v2` to use an OpenStack cloud that has the Block Storage v3 API. That's going to cut off a lot of users.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -688,6 +648,49 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
 	if err != nil {
 		return CheckDeleted(d, err, "server")
 	}
+	computeClient.Microversion = "2.3" // Required for returning delete_on_termination
```
jtopjian: By having this line here, we now require all users to be using an OpenStack cloud that has Compute API v2.3

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -688,6 +648,49 @@ func resourceComputeInstanceV2Read(d *schema.ResourceData, meta interface{}) err
...
+	var serverWithAttachments struct {
+		VolumesAttached []map[string]interface{} `json:"os-extended-volumes:volumes_attached"`
+	}
+	raw.ExtractInto(&serverWithAttachments)
+
+	log.Printf("[DEBUG] Retrieved ServerAttachments %+v", serverWithAttachments)
+
+	bds := []map[string]interface{}{}
+	if len(serverWithAttachments.VolumesAttached) > 0 {
+		blockStorageClient, err := config.blockStorageV2Client(GetRegion(d, config))
```
jtopjian: You're creating a Block Storage v2 client, but are using the v3 API call on line 671.

jsm222:
Hi 
I took your work from https://github.com/jtopjian/terraform-provider-openstack/commit/c7802e72ef74aa22c214e7580b945ab55fcc1afd and changed some things in https://github.com/jsm222/terraform-provider-openstack/commit/817c1427cbf52ec659e99e151ea4b9b4a069c8d6
It is only briefly manually tested. It uses only blockversionclientv2 and always sets delete_on_termination to false, since it is only relevant on creation time.. then we do not need microversiobn 2.3

jsm222:
See #768

jsm222:
Closed replaced by #768

PR #766, opened 2019-06-07
Acc Tests: Update lowercase security group test

jtopjian:
A recent update to Neutron will cause an ipv6 icmp security group rule
to be returned with a protocol of ipv6-icmp.

We can't update the test to use ipv6-icmp since Nova does not allow this
protocol type.

As a workaround, we change the protocol to tcp.

/cc @kayrus


kayrus:
This change just fixes the tests. What about the transition to a newer neutron version for people, who use icmp-ipv6? Should terraform detect this change, like a custom hash func for the security group rule block?

jtopjian:
I don't think Terraform is able to do that? But maybe I'm misunderstanding. 

Basically, once the next release of OpenStack happens, users will no longer be able to use `openstack_compute_secgroup_v2` to create IPV6+ICMP rules.

That's OK, though, because the Compute Security Group API has been deprecated for 2 years now.

I'm happy to keep these resources around for older environments.

PR #768, opened 2019-06-07
Enable import instance

jsm222:
Since openstack does not keep track of NIC ordering, the user
must specify network in the order of the imported state.

### DIFF
--- a/openstack/import_openstack_compute_instance_v2_test.go
+++ b/openstack/import_openstack_compute_instance_v2_test.go
@@ -1,0 +1,55 @@
+package openstack
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccComputeV2Instance_importBasic(t *testing.T) {
+	resourceName := "openstack_compute_instance_v2.instance_1"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckComputeV2InstanceDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccComputeV2Instance_basic,
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"stop_before_destroy",
+					"force_delete",
+				},
+			},
+		},
+	})
+}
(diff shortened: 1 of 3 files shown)


jsm222:
Only fails TestAccComputeV2SecGroup_lowerCaseCIDR  fixed with #766
testing.go:568: Step 0 error: Check failed: Check 2/2 error: openstack_compute_secgroup_v2.sg_1: Attribute 'rule.3862435458.cidr' not found

jtopjian:
@jsm222 Thanks! I'll take a look at this over the weekend and test it with some of the other instance test scenarios that we have.

jsm222:
Keep in mind that I, to avoid computeclient microversion requirement of 2.3, hardcoded `"delete_on_termination": true` It should not matter much IMO as it is a on creation time thing only.

jtopjian:
@jsm222 I had some time to run a few more tests and had trouble with the following:

* testAccComputeV2Instance_bootFromVolumeImage
* testAccComputeV2Instance_blockDeviceNewVolume

I'm not sure if these are worth looking into?

I also had some trouble with `TestAccComputeV2Instance_importCrazyNICs`, but it's a problem with the testing environment that I hope to resolve soon.

jsm222:
To things about the faililng volume tests. They both have destination_type = "local" and/or non bootable disks in block_device. The import code assumes that all block_device entries are bootable volumes from images. Other volumes can manually be imported as volumes and volume attachments. Import of ephemeral root disk with specific size is not easy as nova does not expose that info after creation. Nor is import of multiple ephemeral disks. Since they will always require recreation, the value of having them in the imported state is IMO small - You could simply avoid adding them to your configuration.

(If we need the block_device of the ephemeral root disk imported, it is possible since in that case image_id is set in the response of nova api)

The TestAccComputeV2Instance_importCrazyNICs fails of course because of the imported order are by  the test system expected to be equal to the order at creation time.

joaosa:
It would be great to have this :)

jtopjian:
@jsm222 I apologize for the late reply.

That all sounds good to me, but we will need to have those caveats documented. Can you add details about importing instances to this page (https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/docs/r/compute_instance_v2.html.markdown) in a new section at the bottom called "Importing Instances"?

The section should outline exactly what you provided: the cases when instances can be imported and cannot be imported.

Once that's in place, I'll let @ozerovandrei make the call if/when he wants to merge this.

kayrus:
@jsm222 great job, thanks!

jtopjian:
@jsm222 echoing the others, thank you very much for working on this.

Issue #770, opened 2019-06-11
scheduler_hints 'query' broken


ryanmickler:
I'm trying to create new instances on a specific host, as in
```
  scheduler_hints {
    query = [ "s==","$host","os-compute-01" ]
  }
```
However, this throws an error
`"Invalid input for field/attribute query. Value: [None, None, None, 's==', '$host', 'os-compute-01'] ... is not of type 'string', 'object'`

The first three `None` interfaces come from

https://github.com/terraform-providers/terraform-provider-openstack/blob/0e66a94ba54ebe79d1c8ee5b896b225360355ec3/openstack/resource_openstack_compute_instance_v2.go#L1039

I think this should be changed to 
`query := []interface{}`

To start with an empty list.

ryanmickler:
also, can i suggest adding something like
```
JsonFilter syntax
e.g. Query:  [">=", "$free_ram_mb", "1024"],
Must be a conditional statement in the format of [op,variable,value]
```
To the documentation to explain how to use this feature

jtopjian:
Thanks for reporting this - we'll take a look. 

In the meantime, for your specific case, an alternative way of doing this should be to set the `availability_zone` to `ZONE:os-compute-01` where `ZONE` is an availability zone listed from

```
$ openstack availability zone list
```

jtopjian:
@ryanmickler This has been fixed in #771. If you're able to compile from source, please let me know if you're still having problems. Otherwise, it'll be available upon our next release (and if you're still having problems, please do let us know then, too).

PR #771, opened 2019-06-13
Compute v2: Fix scheduler hints query

jtopjian:
For #770 

Between the bug fixed in ` openstack/resource_openstack_compute_instance_v2.go` and the required vendor fix, it appears that the scheduler hints query never worked.

These two commits fix this and allow "simple" queries to work -- queries which do not require a sublist of conditionals. IIRC, this was the original design goal since it would be hard to model such syntax in terraform.


jtopjian:
recheck

kayrus:
recheck

kayrus:
recheck

jtopjian:
recheck

Issue #767, opened 2019-06-07
Removal of hashicorp/terraform/flatmap dependency


radeksimko:
👋 

We're currently in first planning phases of decoupling `hashicorp/terraform`'s `helper` packages into what we'll soon call SDK. As part of preparation work for this decoupling I have analysed all active official providers and found OpenStack to be using one particular legacy package - `flatmap`.

https://github.com/terraform-providers/terraform-provider-openstack/search?q=flatmap&unscoped_q=flatmap

There's no particular deadline we set for sunsetting that package and removing it from Terraform's core, but there are some chances it might not make it to the initial decoupled version of SDK which will otherwise be more or less copy and compatible with the API currently exposed via `helper` packages in Terraform.

It seems that this package is only used for extraction of 1 field from the `map[string]string` state exposed to the test framework. The future SDK (not the initial one) is likely going to use [cty](https://github.com/zclconf/go-cty) to represent typed data from state/configuration. Terraform 0.12 already uses cty internally, but there's more work to do in regards to exposing this fact to providers.

Therefore there isn't any more elegant solution at this point other than hard-coding the test data as `[]interface{}` - this is what pretty much all other (official) providers do today.

jtopjian:
This should be resolved in #772. As the comments in #772 states: I don't believe it's a full solution (we'll need to revisit with a more modern method of inspecting the state of test resources), but at least our dependency on flatmap is gone.

PR #772, opened 2019-06-14
Acc Tests: Remove flatmap

jtopjian:
For #767 

This commit removes the flatmap package from tests since it has been
deprecated.

I have a feeling that this is not a permanent solution. In the future, we will still need to modify the tests to account for a new way of parsing the state. But at least we have removed flatmap, which is a good first step.

It's also unfortunate that I had to disable some VPN checks, but I have no way of testing them.

I've also moved `testAccCheckNetworkingV2Tags` out of `utils.go` and into a test file since I think that's the better location.

Finally, I renamed some VPN files to follow proper naming conventions. I understand that's not part of this fix, but figured it was small enough to do here.


huangtianhua:
recheck stable/mitaka

huangtianhua:
recheck stable/stein

bzhaoopenstack:
recheck

bzhaoopenstack:
recheck stable/mitaka

bzhaoopenstack:
recheck stable/stein

Issue #760, opened 2019-06-01
QoS policies and rules support


ozerovandrei:
This provider should support OpenStack Networking V2 QoS policies, rules and `qos_policy_id` field in the network and port resources.

Upstream progress is tracked in [#1027](https://github.com/gophercloud/gophercloud/issues/1027).

ozerovandrei:
@jtopjian I was thinking about design of those resources and I think that it is better to implement it in that way:
 - `qos_policy_id` field in the existing network and port resources/datasources;
 - `qos_policy` resource/datasource that won't contain any rules given that this field is too dynamic;
 - `qos_bandwidth_limit_rule` resource/datasource;
 - `qos_dscp_marking_rule` resource/datasource;
 - `qos_minimum_bandwidth_rule` resource/datasource.

What do you think?

jtopjian:
@ozerovandrei Yep, that sounds good to me.

Just to confirm, though, the resource names will be something like `openstack_networking_qos_policy_v2`, `openstack_networking_qos_bandwidth_limit_rule_v2`, etc. I know some of these names will be quite long, but it'll be consistent with other names.

ozerovandrei:
Yeah they will have exactly those names.

PR #774, opened 2019-06-16
Networking V2: add QoS Policy resource

ozerovandrei:
Implement "openstack_networking_qos_policy_v2" resource with acceptance
tests and website documentation.

For #760

### DIFF
--- a/openstack/types.go
+++ b/openstack/types.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/qos/policies"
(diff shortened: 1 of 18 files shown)


openstack/types.go
```diff
@@ -306,3 +307,9 @@ type SiteConnectionCreateOpts struct {
 	siteconnections.CreateOpts
 	ValueSpecs map[string]string `json:"value_specs,omitempty"`
 }
+
+// QoSPolicyCreateOpts represents the attributes used when creating a new QoS policy.
+type QoSPolicyCreateOpts struct {
```
jtopjian: nit: add this to `openstack/networking_qos_policy_v2.go`. The remaining types that are in `types.go` are resources that still need to be cleaned up (I think there are a few networking ones in queue? The VPN ones are untouched).

ozerovandrei: Not really. Some of them were already refactored.
But I will have this in mind and those types will be migrated.

ozerovandrei:
recheck

ozerovandrei:
recheck

jtopjian:
recheck

PR #777, opened 2019-06-19
Readme: update Terraform and Go versions

ozerovandrei:
Use Go 1.12 and Terraform 0.12 in README.md.


kayrus:
what about the `.travis` file? is it used?

ozerovandrei:
@kayrus yeah, it should be updated in Travis config as well.
I don't think that we need `.go-version` since major popular providers (AWS, Google) don't have it. But I will ask about it in Slack.

ozerovandrei:
Official answer is:
```
tombuildsstuff [12:14 PM]
it’s an optional thing if you’re using GoEnv
it might be worth doing if you use a different version of Go on your machine, but we tend to stick to the latest (or close enough) for 1.12.x
```

I don't think that we need it for that matter.

ozerovandrei:
recheck

PR #799, opened 2019-07-16
vpnaas_site_connection: Switch admin_state_up default to true

swegener:
because false is a bad default.


jtopjian:
Another option would be to remove the default value entirely and then use `d.GetOkExists` to see if `admin_state_up` was set in the config at all. This way, unspecified `admin_state_up` should implicitly default to true. However, `createOpts` is being defined as a Builder interface which makes this method difficult to do.

@ozerovandrei I think this is the best way to go for this fix. Changing the default value might impact some users but given `true` is most likely what everyone wants, I suspect the impact will be low (or none).

Issue #657, opened 2019-02-05
Support for Managing Project Quotas


arusso:
Hello,

I would like to be able to manage OpenStack block storage and compute quotas from within Terraform. Current docs suggest this isn't currently possible (or at least I found no mention of them).

gophercloud appears to already have support for this, so I believe this is just a matter of exposing this in the provider:

- [block storage quotasets](https://godoc.org/github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/quotasets)
- [compute quotasets](https://godoc.org/github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/quotasets)

### Terraform Version

```
$ terraform -v
Terraform v0.11.10
+ provider.null v2.0.0
+ provider.openstack v1.14.0
+ provider.vault v1.5.0
```

### Terraform Configuration Files

It seems like this should reasonably be split into two new resources and a  config could look something like:

````
data "openstack_identity_project_v3" "project_1" {
  name = "demo"
}

resource "openstack_compute_quotaset_v2" "project_1-cqs" {
  project = "${data.openstack_identity_project_v3.id}"
  fixed_ips = 50
  keypairs = 100
  cores = 50
  instances = 10
  ...
}

resource "openstack_blockstorage_quotaset_v2" "project_1-bqs" {
  project = "${data.openstack_identity_project_v3.id}"
  snapshots = 20
  gigabytes = 20000
}
```

### Expected Behavior

Quotas should be applied to projects as configured.

### Actual Behavior

NA

jtopjian:
@arusso Thanks!

I think the only difficulty with implementing this is to decide what to do for the "deletion" part. One option is to consider deletion to be a "no-op" and not try to revert any of the previously modified quota items. Another option is to store the original quotas, but that might get a little verbose/ugly.

A note about the compute quotas: though it supports resources like floating IPs and fixed IPs, I don't think those are respected in newer versions of OpenStack. We can still include them to support older versions, but the documentation will need to detail that.

arusso:
I think a no-op on delete is a completely reasonable approach in this
situation.

Including the deprecated quotas is a probably good idea. I know we probably
won’t be on a version that supports the network resource quotas via the
neutron api for a few years so I’d very much like to be able to manage them
until then.

Aaron

tibers:
Please can we have this feature? I was just paging through our codebase and folks are using `local-exec` to set quotas and it's making a mess.

[...]

PR #806, opened 2019-07-26
BlockStorage V2: add Quotaset resource

ozerovandrei:
Implement the "openstack_blockstorage_quotaset_v2" resource.

For #657

### DIFF
--- a/openstack/import_openstack_blockstorage_quotaset_v2_test.go
+++ b/openstack/import_openstack_blockstorage_quotaset_v2_test.go
@@ -21,10 +21,9 @@ func TestAccBlockStorageV2Quotaset_importBasic(t *testing.T) {
 				Config: testAccBlockStorageQuotasetV2_basic,
 			},
 			{
-				ResourceName:            resourceName,
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"groups"},
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
 			},
 		},
 	})
--- a/openstack/import_openstack_blockstorage_quotaset_v2_test.go
+++ b/openstack/import_openstack_blockstorage_quotaset_v2_test.go
@@ -21,9 +21,10 @@ func TestAccBlockStorageV2Quotaset_importBasic(t *testing.T) {
 				Config: testAccBlockStorageQuotasetV2_basic,
 			},
 			{
-				ResourceName:      resourceName,
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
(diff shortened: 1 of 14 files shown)


openstack/import_openstack_blockstorage_quotaset_v2_test.go
```diff
@@ -0,0 +1,31 @@
...
+		CheckDestroy: testAccCheckIdentityV3ProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccBlockStorageQuotasetV2_basic,
+			},
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"groups"},
```
kayrus: Shouldn't `groups` be added into a gophercloud QuotaSet struct? https://github.com/gophercloud/gophercloud/blob/951d29d7ead8543b9b6e3f38b971ae952df0a527/openstack/blockstorage/extensions/quotasets/results.go#L10

website/docs/r/blockstorage_quotaset_v2.html.markdown
```diff
@@ -0,0 +1,84 @@
+---
+layout: "openstack"
+page_title: "OpenStack: openstack_blockstorage_quotaset_v2"
+sidebar_current: "docs-openstack-resource-blockstorage-quotaset-v2"
+description: |-
+  Manages a V2 quotaset resource within OpenStack.
+---
+
+# openstack\_blockstorage\_quotaset_v2
```
kayrus:
```suggestion
# openstack\_blockstorage\_quotaset\_v2
```

website/docs/r/blockstorage_quotaset_v2.html.markdown
```diff
@@ -0,0 +1,84 @@
...
+
+Manages a V2 quotaset resource within OpenStack.
```
kayrus:
```suggestion
Manages a V2 block storage quotaset resource within OpenStack.
```

website/docs/r/blockstorage_quotaset_v2.html.markdown
```diff
@@ -0,0 +1,84 @@
...
+
```
kayrus: Additionally here should be a note about the no-op deletion.

openstack/import_openstack_blockstorage_quotaset_v2_test.go
ozerovandrei: It should be and I created an issue to track it: https://github.com/gophercloud/gophercloud/issues/1667

I guess it's better to fix it upstream instead of merging PR with this workaround.

website/docs/r/blockstorage_quotaset_v2.html.markdown
ozerovandrei: You're right, thanks 👍

PR #815, opened 2019-08-01
Key manager v1: add secret data source

### LINKED ISSUE #813: Key Manager V1: Introduce secrets data source
`data.openstack_keymanager_secret_v1` is useful to get the existing secret payload, which wasn't created by terraform.

kayrus:
Resolves #813


kayrus:
@jtopjian @ozerovandrei ready for review. has to be checked in the local environment since openlab doesn't support the barbican yet. https://github.com/theopenlab/openlab/issues/324

PR #837, opened 2019-08-14
Networking V2: Mark dns attributes as computed

### LINKED ISSUE #836: openstack_networking_network_v2 dns_domain makes TF runs not idempotent and breaks on multiple runs
When creating a network, the `dns_domain` flag will be set by Openstack, if neutron is configured accordingly. It's not meant to be set by a user. If the user explicitly sets this value, an error will occure, stating `Error: Error updating openstack_networking_network_v2 VPC_ID: Bad request with: [PUT https://vpc.eu-de.otc.t-systems.com/v2.0/networks/VPC_ID], error message: {"NeutronError":{"message":"dns_domain can not be set when update network.","type":"HTTPBadRequest","detail":""}}`

However, this will ALSO occure if the network was created and terraform is applied again, as terraform tries to change the default value to null again:
```
  # module.vpc.openstack_networking_network_v2.os_network_1 will be updated in-place
  ~ resource "openstack_networking_network_v2" "os_network_1" {
        admin_state_up          = true
        all_tags                = []
        availability_zone_hints = []
      - dns_domain              = "eu-de.compute.internal." -> null
```

### Terraform Version
```
Terraform v0.12.6
+ provider.local v1.3.0
+ provider.null v2.1.2
+ provider.openstack v1.21.1
```

### Affected Resource(s)
- openstack_networking_network_v2

```hcl
resource "openstack_networking_network_v2" "os_network_1" {
  name              = "${var.project}-network"
  admin_state_up    = true
 #  dns_domain        = "eu-de.compute.internal."
}
```

### Expected Behavior
Terraform should either check the value of said field _after_ creation and include it into the networks state, or provide an option to ignore the dns_domain setting. Removing it is not an option as Openstack admins may use terraform for network management, which are allowed to set `dns_domain`. Setting it to the Openstack default value will also result in breaking as Neutron seems to deny any request where dns_domain is stated.

### Actual Behavior
Terraform tries to overwrite the default value set by Openstack and breaks

### Steps to Reproduce
1. use `openstack_networking_network_v2` to create a network on an Openstack installation with dns extension enabled and do not set `dns_domain`
2. `terraform apply` once, check network settings via `openstack network show <net_name>`
3. run `terraform apply` again and see it breaking

### References
- Looks similar to #742

kayrus:
Resolves #836

### DIFF
--- a/openstack/resource_openstack_networking_floatingip_v2.go
+++ b/openstack/resource_openstack_networking_floatingip_v2.go
@@ -102,12 +102,14 @@ func resourceNetworkingFloatingIPV2() *schema.Resource {
 			"dns_name": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 				ForceNew: true,
 			},
 
 			"dns_domain": {
 				Type:         schema.TypeString,
 				Optional:     true,
+				Computed:     true,
 				ForceNew:     true,
 				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^$|\.$`), "fully-qualified (unambiguous) DNS domain names must have a dot at the end"),
 			},
--- a/openstack/resource_openstack_networking_network_v2.go
+++ b/openstack/resource_openstack_networking_network_v2.go
@@ -157,6 +157,7 @@ func resourceNetworkingNetworkV2() *schema.Resource {
 			"dns_domain": {
 				Type:         schema.TypeString,
 				Optional:     true,
+				Computed:     true,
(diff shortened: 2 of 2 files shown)


kayrus:
@ozerovandrei @jtopjian ready for review

kayrus:
@dev00 can you compile this branch and test in your env?

dev00:
@kayrus works perfectly, thanks :)

Let's hope this gets merged soon, would bring me a great step forward!

Issue #264, opened 2018-03-12
Add ability to control caching


jtopjian:
It would be useful to allow Terraform to control the `Cache-Control` header of the Gophercloud http proxy. This would help with OpenStack environments which cache the responses, ensuring Terraform receives the most up-to-date information.

ozerovandrei:
Implemented in #849

PR #849, opened 2019-09-03
Add ability to control caching

freddebacker:
Added workaround for cases when Openstack API doesn't provide proper HTTP Cache-Control headers [GH-264]

Also see:
- https://github.com/gophercloud/gophercloud/issues/727
- http://specs.openstack.org/openstack/api-sig/guidelines/http/caching.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 1.22.1 (Unreleased)
 
+IMPROVEMENTS
+
+* Added workaround for cases when Openstack API doesn't provide proper HTTP Cache-Control headers [GH-264]
+
 BUG FIXES
 
 * Fixed the bug where `openstack_identity_auth_scope_v3` caused a panic within the domain-scope [GH-851]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 ## 1.22.1 (Unreleased)
+
+IMPROVEMENTS
+
+* Added workaround for cases when Openstack API doesn't provide proper HTTP Cache-Control headers ([#264](https://github.com/terraform-providers/terraform-provider-openstack/issues/264))
+
 ## 1.22.0 (September 05, 2019)
 
 FEATURES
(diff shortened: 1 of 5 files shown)


freddebacker:
Hi @jtopjian,

For your consideration, following up on our conversation from 1,5 year ago...

Regards,
Fred

ozerovandrei:
Hello @freddebacker,

Thank you for working on that.

I read the Gophercloud discussion and the following OpenStack resources:
https://review.opendev.org/#/c/550468/
https://bugs.launchpad.net/nova/+bug/1747935

I tend to agree that terraform-provider-openstack has to add this header because currently we not make any assertions about controlling cache.
It means that we must add the `Cache-Control: no-cache` header, like it's described in the spec that you provided.

But from the other hand this seems that we could also allow to control this header.
I think that `Cache-Control: no-cache` header needs to be added by default and we should add the possibility to remove it.

Could you please add a new field into the [LogRoundTripper](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/types.go#L28).

We then also need to add this parameter into the [Config](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L19) and pass it into the [Transport](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L193). 

It should be populated in [configureProvider](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider.go#L417).

freddebacker:
Gents,

Before making this change, could you confirm that what you want is a boolean flag that controls whether or not the `Cache-Control: no-cache` HTTP header is added in all requests? We're not talking about logic to populate the `Cache-Control` HTTP header with other possible values allowed by the HTTP specifications.

My personal opinion about this is that there is no benefit from making this configurable, as I don't see any valid use-case where you would accept a previously cached answer to an API-call. You would always want the current state of the object you're querying, not something that once used to be the state.

As stated in the discussions you mentioned, it *should* be the server (Openstack API) responsible to set proper HTTP headers, but unfortunately they don't (yet). I consider this work-around as something that must protect you from faulty API behavior and therefore I don't see any reason to make this optional.

If you disagree, I'm willing to make the changes you described...

Waiting on your confirmation...

Regards,
Fred

ozerovandrei:
> Before making this change, could you confirm that what you want is a boolean flag that controls whether or not the Cache-Control: no-cache HTTP header is added in all requests?
Yes. I guess that is better to add this header by default.

But it would be better if user could remove `Cache-Control: no-cache` header by setting `disable_no_cache_header: True` in the provider settings: https://www.terraform.io/docs/providers/openstack/#configuration-reference
It seems that it's a safe way to add this change and to leave the current behaviour for someone who needs it (who knows what kind of OpenStack installations are there).

>  We're not talking about logic to populate the Cache-Control HTTP header with other possible values allowed by the HTTP specifications.
No. As you said it should be the server duty to set proper HTTP headers.

freddebacker:
Done implementing the `disable_no_cache_header` flag...

jtopjian:
> It seems that it's a safe way to add this change and to leave the current behaviour for someone who needs it (who knows what kind of OpenStack installations are there).

Agreed 100%

It's always possible that making a blanket change like this will break _someone's_ workflow. Everyone uses Terraform differently and we need to make sure blanket changes are able to be toggled.

PR #855, opened 2019-09-06
Compute V2: Read flavor's ephemeral option

### LINKED ISSUE #852: Compute V2: Read flavor's ephemeral option
# Affected Resource(s)
- openstack_compute_flavor_v2

### Expected Behavior

Read func should fetch and set the ephemeral option

### Actual Behavior

The `d.Set("ephemeral", fl.Ephemeral)` code is commented out

kayrus:
Resolves #852


kayrus:

kayrus:
@ozerovandrei good point, but If other elements doesn't have computed attribute (except `extra_specs`), why should `ephemeral` have it?

PR #856, opened 2019-09-06
Compute V2: Fix not found logic for the flavor access

### LINKED ISSUE #853: Compute V2: Fix not found logic for the flavor access
### Affected Resource(s)
- openstack_compute_flavor_access_v2

### Expected Behavior

If the flavor access doesn't, then terraform has to use the `CheckDeleted` helper.

### Actual Behavior

If the flavor access doesn't exist anymore, it is impossible to remove it from the terraform state:

```
Error: Error removing tenant  access from flavor : Resource not found
```

kayrus:
Resolves #853


kayrus:

PR #857, opened 2019-09-06
Identity V3: Fix not found logic detection for the role assignments

### LINKED ISSUE #854: Identity V3: Fix not found logic detection for the role assignments
### Affected Resource(s)
- openstack_identity_role_assignment_v3

### Expected Behavior

If the role assignment doesn't exist, then terraform has to use the `CheckDeleted` helper.

### Actual Behavior

Terraform doesn't remove the role state, when the role is not found.

kayrus:
Resolves #854


kayrus:

PR #861, opened 2019-09-12
Identity: authenticate against identity only, when it is needed

### LINKED ISSUE #860: Auth: reduce an amount of auth requests
### Expected Behavior

Terraform should auth against the identity/keystone only when an action has to be performed.

### Actual Behavior

Each entity of the auth provider authenticates **three times**. See the debug log for a regular data source: https://gist.github.com/kayrus/bc5cee71549a15b61b5972b707b9d025

```hcl
data "openstack_identity_auth_scope_v3" "scope" {
  name = "my_scope"
}
```

### Steps to Reproduce
1. `terraform apply`

### Important Factoids

An *authentication when it is really needed* will also help to resolve the dependency problems, when you create projects and then want terraform to automatically auth inside the just created projects, e.g.

```hcl
resource "openstack_identity_project_v3" "project" {
  name      = "my-cool-project"
  domain_id = "1234567890"
}

provider "openstack" {
  alias = "project_scope"

  auth_url    = "https://identity.local/v3"
  tenant_name = "${openstack_identity_project_v3.project.name}"
}

resource "openstack_networking_network_v2" "network_1" {
  provider = "openstack.project_scope"

  name      = "my-net"
}
```

### References

This can also allow to use an individual provider for dynamic elements, like it is discussed (but not yet supported) here https://github.com/hashicorp/terraform/issues/16967

kayrus:
Resolves #860

### DIFF
--- a/openstack/config.go
+++ b/openstack/config.go
@@ -47,7 +47,12 @@ type Config struct {
 	MaxRetries                  int
 	DisableNoCacheHeader        bool
 
-	OsClient *gophercloud.ProviderClient
+	OsClient      *gophercloud.ProviderClient
+	authOpts      *gophercloud.AuthOptions
+	authenticated bool
+	authFailed    error
+	swClient      *gophercloud.ServiceClient
+	swAuthFailed  error
 }
 
 // LoadAndValidate performs the authentication and initial configuration
@@ -199,19 +204,31 @@ func (c *Config) LoadAndValidate() error {
 		},
 	}
 
-	// If using Swift Authentication, there's no need to validate authentication normally.
-	if !c.Swauth {
-		err = openstack.Authenticate(client, *ao)
-		if err != nil {
-			return err
-		}
-	}
-
 	if c.MaxRetries < 0 {
 		return fmt.Errorf("max_retries should be a positive value")
 	}
 
+	c.authOpts = ao
 	c.OsClient = client
+
+	return nil
+}
+
+func (c *Config) authenticate() error {
+	osMutexKV.Lock("auth")
+	defer osMutexKV.Unlock("auth")
+
+	if c.authFailed != nil {
+		return c.authFailed
+	}
+
+	if !c.authenticated {
+		if err := openstack.Authenticate(c.OsClient, *c.authOpts); err != nil {
+			c.authFailed = err
+			return err
+		}
+		c.authenticated = true
+	}
 
 	return nil
 }
@@ -263,6 +280,10 @@ func (c *Config) getEndpointType() gophercloud.Availability {
 // which interact with the various OpenStack services.
 
 func (c *Config) blockStorageV1Client(region string) (*gophercloud.ServiceClient, error) {
+	if err := c.authenticate(); err != nil {
+		return nil, err
+	}
+
 	client, err := openstack.NewBlockStorageV1(c.OsClient, gophercloud.EndpointOpts{
 		Region:       c.determineRegion(region),
 		Availability: c.getEndpointType(),
@@ -279,6 +300,10 @@ func (c *Config) blockStorageV1Client(region string) (*gophercloud.ServiceClient
 }
 
 func (c *Config) blockStorageV2Client(region string) (*gophercloud.ServiceClient, error) {
+	if err := c.authenticate(); err != nil {
+		return nil, err
+	}
+
 	client, err := openstack.NewBlockStorageV2(c.OsClient, gophercloud.EndpointOpts{
 		Region:       c.determineRegion(region),
 		Availability: c.getEndpointType(),
@@ -295,6 +320,10 @@ func (c *Config) blockStorageV2Client(region string) (*gophercloud.ServiceClient
 }
 
 func (c *Config) blockStorageV3Client(region string) (*gophercloud.ServiceClient, error) {
+	if err := c.authenticate(); err != nil {
+		return nil, err
+	}
+
 	client, err := openstack.NewBlockStorageV3(c.OsClient, gophercloud.EndpointOpts{
 		Region:       c.determineRegion(region),
 		Availability: c.getEndpointType(),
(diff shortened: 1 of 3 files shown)


jtopjian:
> Let me know if you have other concerns.

No new concerns - just more of the same 🙂 

> This is a pretty simple optimization

I humbly disagree. This PR is implementing locking in a position in the code that has a high amount of concurrency. This creates a risk of deadlock scenarios. Have you tested this change using Go's race detector?

In addition, have you looked at other providers for similar implementations?

> When token auth is used, the current logic still auths against the keystone to create a token for a token

This seems like the better problem to solve. I realize there might have to be Gophercloud changes (or some util function?) to help with this, but that's fine.

The result will be similar to what you're proposing here (reduced authentication calls) but decoupled from the provider (thus allowing integration with, say, Vault) and not implementing any sort of locking.

kayrus:
> I humbly disagree. This PR is implementing locking in a position in the code that has a high amount of concurrency. This creates a risk of deadlock scenarios.

I still see no risk here. Do you mean that the lock in the `authenticate` func can deadlock unpredictably? I can use a simple golang mutex if you want to avoid the `github.com/hashicorp/terraform/helper/mutexkv` package usage. It would be great if you can describe the case when it can happen. @majewsky, could you please assist me here? Probably I'm missing some details here.

> Have you tested this change using Go's race detector?

I compiled the binary with the `-race` flag and tested it.

> In addition, have you looked at other providers for similar implementations?

Other providers have different logic, they don't check the credentials during the provider initialization, but only during the resource call. This actually allows to use the code like this:

```hcl
provider "kubernetes" {
  host                   = "${azurerm_kubernetes_cluster.main.kube_config.0.host}"
  username               = "${azurerm_kubernetes_cluster.main.kube_config.0.username}"
  password               = "${azurerm_kubernetes_cluster.main.kube_config.0.password}"
  client_certificate     = "${base64decode(azurerm_kubernetes_cluster.main.kube_config.0.client_certificate)}"
  client_key             = "${base64decode(azurerm_kubernetes_cluster.main.kube_config.0.client_key)}"
  cluster_ca_certificate = "${base64decode(azurerm_kubernetes_cluster.main.kube_config.0.cluster_ca_certificate)}"
}
```

which is not available in current openstack provider.

jtopjian:
> I compiled the binary with the -race flag and tested it.

Can you try running `go test -race ...` on a handful of acceptance tests? This will show code that is susceptible to deadlocks, regardless if a deadlock happened or not.

kayrus:
> Can you try running go test -race ... on a handful of acceptance tests? This will show code that is susceptible to deadlocks, regardless if a deadlock happened or not.

I'll try to find a way to enable this option in openlab tests.

> However, I still think allowing the ability to specify an authenticated token is the best way to go. As I mentioned, this allows the provider to receive an authenticated token from another source, providing flexibility.

this won't help in situations like this:

```hcl
resource "openstack_identity_project_v3" "project" {
  name      = "my-cool-project"
  domain_id = "1234567890"
}

provider "openstack" {
  alias = "project_scope"

  auth_url    = "https://identity.local/v3"
  tenant_name = "${openstack_identity_project_v3.project.name}"
}

resource "openstack_networking_network_v2" "network_1" {
  provider = "openstack.project_scope"

  name      = "my-net"
}
```

jtopjian:
> this won't help in situations like this:

Actually, you _should_ be able to specify a project ID for all network resources and the network resource will be created under the specified project's scope. For the example you gave, you can do this today without this patch.

But for _other_ OpenStack resources (compute, block storage, etc), they do not support the per-resource project scoping - it's a limitation of the OpenStack API. And because of this, this feels like you want to bend Terraform to do something to make up for a deficiency in the OpenStack API.

> I'll try test this race enabled binary on a complex terraform script.

I recommend creating a sufficient staging environment and then running the acceptance test suite with `-race` added to `go test ...`. 

> Having an option to enable this behavior sounds like an overkill. How would you call it?

It can appear overkill, but it would be temporary. Think of it like opting into a beta/testing feature. This provides a way for a larger audience to test something potentially breaking. 

I understand that we have a difference of opinion about the possible negative effects that a change like this can create. This is a matter of risk management. My personal experience has been that as soon as locking mechanisms are added, deadlocks _will_ be reported.

kayrus:
> Actually, you should be able to specify a project ID for all network resources and the network resource will be created under the specified project's scope. For the example you gave, you can do this today without this patch.

JFYI: in certain situations you can have a domain scope only (domain admin, not the whole cluster admin) with an ability to create projects, but not creating resources inside these projects (this is allowed only to the cluster admin). Therefore you have to auth within the new project scope. The network resource was used as an example, there is a number of other openstack resources, which have to be created inside the project within one terraform shot.

jtopjian:
> The network resource was used as an example, there is a number of other openstack resources, which have to be created inside the project within one terraform shot.

I agree. My point was that this seems to be more of a limitation of the OpenStack API of effectively using an admin account or admin API for provisioning.

I feel like there are two use-cases being discussed here:

1. Reduce the amount of authentication calls. This is solvable by modifying Gophercloud or Terraform to be able to use an authenticated token further authentication (rather than providing a token which is exchanged for another token as it currently happens in Terraform).

2. The ability to scope to a new project created in the same Terraform apply. This is a worthwhile feature to have since this can help with standardized workflows for account provisioning. But could this also be achieved by creating a project, creating or obtaining a new scoped token to that new project, and then creating resources? I believe this is something you have been looking at for a while with some other ideas in the past. Unfortunately the combination of OpenStack's lack of flexible admin APIs and Terraform's current feature set seems to make this a difficult thing to implement.

jtopjian:
Also, to be clear: I am not trying to veto or dismiss this PR. All I'm trying to say is that introducing locking needs to be carefully considered. Based on my personal experience, introducing locking, no matter how simple it appears, needs to be reviewed and tested in detail. Either extensive testing needs to be done with `-race` or the ability to opt-in or opt-out of locking should be added in the event that this breaks unidentified / unknown workflows and environments.

kayrus:
@jtopjian no worries, I totally understand your concerns, and I really appreciate your opinion. I need some time to test this. I'll let you know the results as soon as possible.

kayrus:
@jtopjian my local tests showed, that there are no deadlocks or other race issues. I added the `delayed_auth` provider flag and rebased.

jtopjian:
Thank you very much for testing as well as for changing this into more of a "feature flag". This addresses all concerns I have, so I'll let @ozerovandrei take it from here 🙂 

If this does get merged, it might be a good idea to advertise this in the CHANGELOG so users can opt-in to testing it.

kayrus:
recheck

kayrus:

PR #866, opened 2019-09-17
detach ports before deleting instance

### LINKED ISSUE #858: Refreshing an instance that have an attached port (without refreshing the port) do not detach the port interface before deleting the instance
### Terraform Version
v0.12.7

### Affected Resource(s)

```tf
resource  "openstack_compute_instance_v2"  "zookeeper" {

    name           = "test"
    image_id        = "${var.image_id}"
    flavor_name       = "${var.flavor}"
    key_pair        = "${var.key_pair_name}"

  network {
    port = "${var.port_id}"
  }
  stop_before_destroy = true
}
```

### Debug Output
 Error creating OpenStack server: Expected HTTP response code [] when accessing [POST http://10.0.0.1:8774/v2.1/64b75fdb8a5b403cbfb2166d6f82c5f3/servers], but got 409 instead
{"conflictingRequest": {"message": "Port e9a37693-314f-4273-a822-a93aab3c73f4 is still in use.", "code": 409}}

### Expected Behavior
Port need to be detached correctly and reattached to the new instance

### Actual Behavior
the instance refresh delete the vm and rely on openstack to detach the port from it, this is not a good openstack practice as the port detaching api call in that case is async and could lead to an error which make the port unusable (attached to an instance that do not exist)

### Steps to Reproduce
Use an existing port (not part of the terraform definition) hardcode the port id in an instance as above
1. `terraform apply`
2. `change params in the instance definition
3 `terraform apply`

dhiaayachi:
fixes #858 detach port before deleting the instance

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -351,6 +351,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"detach_ports_before_destroy": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 			"force_delete": {
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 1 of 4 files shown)


openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -913,6 +913,27 @@ func resourceComputeInstanceV2Delete(d *schema.ResourceData, meta interface{}) e
 		}
 	}
 
+	allInstanceNetworks, err := getAllInstanceNetworks(d, meta)
+	if err != nil {
+		return CheckDeleted(d, err, "Unable to get openstack_compute_instance_v2 ports")
```
kayrus: I don't think, that you have to delete the compute resource state, when it doesn't have networks. Just print the log message. If there is no error, then you can loop over the `allInstanceNetworks`.

kayrus: @dhiaayachi my previous comment is still relevant.

kmbulebu:
Would this allow me to use the Terraform lifecycle option, `create_before_destroy`? Previously I could not use this option as the port would be seen as already in use by the existing instance.

https://www.terraform.io/docs/configuration/resources.html#create_before_destroy

dhiaayachi:
I updated the code to add it as a vendor option but the acceptance tests are timing out for an unknown reason I'm not able to understand what exactly failed.
Help with this would be appreciated.

ozerovandrei:
recheck

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -912,7 +917,34 @@ func resourceComputeInstanceV2Delete(d *schema.ResourceData, meta interface{}) e
...
+	vendorOptionsRaw := d.Get("vendor_options").(*schema.Set)
+	var detachPortBeforeDestroy bool
+	if vendorOptionsRaw.Len() > 0 {
+		vendorOptions := expandVendorOptions(vendorOptionsRaw.List())
+		detachPortBeforeDestroy = vendorOptions["detach_ports_before_destroy"].(bool)
+	}
+	if detachPortBeforeDestroy {
+		allInstanceNetworks, err := getAllInstanceNetworks(d, meta)
+		if err != nil {
+			return CheckDeleted(d, err, "Unable to get openstack_compute_instance_v2 ports")
```
ozerovandrei: I have the same concern here as @kayrus mentioned before.

We don't need to delete the compute resource state, when it doesn't have networks.

You can use the same logic as for the previous `d.Get("stop_before_destroy").(bool)` block.

So it will be something like:
```go
allInstanceNetworks, err := getAllInstanceNetworks(d, meta)
if err != nil {
    log.Printf("[WARN] Error getting openstack_compute_instance_v2 networks: %s", err)
} else {
    for _, network := range allInstanceNetworks {
         if network.Port != "" {
            ...
}
```

dhiaayachi: Change done, please check

ozerovandrei:
recheck

PR #872, opened 2019-09-18
Glance v2: Add image share resource

### LINKED ISSUE #869: Glance V2: Add image sharing resource
### Affected Resource(s)
- openstack_images_image_share_v2
- openstack_images_share_v2

```hcl
resource "openstack_images_image_v2" "rancheros" {
  name             = "RancherOS"
  image_source_url = "https://releases.rancher.com/os/latest/rancheros-openstack.img"
  container_format = "bare"
  disk_format      = "qcow2"
  visibility = "shared"

  properties = {
    key = "value"
  }
}

resource "openstack_images_image_share_v2" "rancheros" {
  image_id = "${openstack_images_image_v2.rancheros.id}"
  member_id = "%projectid%"
  status = "pending" # for the %projectid% context only
}
```

### Expected Behavior

Image owner has to have an ability to share an image with other projects/users.

### Important Factoids

User has to approve the image on his end, nevertheless user can already use the image with the `pending` status by using its id.

@ozerovandrei @jtopjian please give me a hint for the resource name:
* openstack_images_image_share_v2
* openstack_images_share_v2
* openstack_images_image_member_v2
* ???

Target user can list the shared image using the data source (probably the #139 will be helpful here as well.):

```hcl
data "openstack_images_image_v2" "ubuntu" {
  visibility = "shared"
  member_status = "pending"
}
```

then approve/decline the image membership, probably using the same terraform resource.

Target use case:

* create a project
* share an image from the source project
* auth within the target project scope
* accept the shared image

### References

https://github.com/gophercloud/gophercloud/blob/c1ae5ed189fea28b1222fb6f9d4daf80471e2855/openstack/imageservice/v2/members/results.go
#860 
https://github.com/hashicorp/terraform/issues/16967

kayrus:
Resolves #869 

@jtopjian @ozerovandrei I'd like to know whether the overall concept is ok.


website/docs/r/images_image_v2.html.markdown
```diff
@@ -6,7 +6,7 @@ description: |-
   Manages a V2 Image resource within OpenStack Glance.
 ---
 
-# openstack\_images\_image_v2
+# openstack\_images\_image\_v2
```
jtopjian: Totally minor comment and no action needs taken, just a possibly helpful piece of trivia for the future:

Technically `openstack\_images\_image_v2` is correct, but unintuitive. The last `_` does not need to be escaped because there isn't another `_` for Markdown to render the text change.

kayrus: Thanks for the note. I know that it is fine, but I wanted to align docs with common pattern.

openstack/images_image_v2.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/members"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceImagesImageV2MemberStatusFromString(v string) images.ImageMemberStatus {
+	switch v {
+	case "accepted":
```
ozerovandrei: I guess it's more reliable to check those types in the following way:
```
case string(images.ImageMemberStatusAccepted):
    return images.ImageMemberStatusAccepted
case string(images.ImageMemberStatusAccepted):
    return images.ImageMemberStatusAccepted
```

Same goes for visibility.

In that way we won't create hard-coded copies from the Gophercloud library.
What do you think?

openstack/images_image_v2.go
```diff
@@ -0,0 +1,224 @@
...
+	defer file.Close()
+
+	fstat, err := file.Stat()
+	if err != nil {
+		return -1, "", fmt.Errorf("Error reading image file %q: %s", file.Name(), err)
+	}
+
+	filesize = fstat.Size()
+	filechecksum, err = fileMD5Checksum(file)
+
```
ozerovandrei: Nit: remove this empty line.

openstack/images_image_v2.go
```diff
@@ -0,0 +1,224 @@
...
+		log.Printf("[DEBUG] OpenStack image status is: %s", img.Status)
+
+		return img, fmt.Sprintf("%s", img.Status), nil
+	}
+}
+
+func resourceImagesImageV2BuildTags(v []interface{}) []string {
+	var tags []string
+	for _, tag := range v {
+		tags = append(tags, tag.(string))
```
ozerovandrei: We can use `tags := make([]string, len(v))` here to reduce allocations.

kayrus:
@ozerovandrei fixed.

PR #892, opened 2019-10-04
Fixed names and added volume example

thobalose:
- Fixed invalid and reserved names
- Added attach, format, and mount volume example


ozerovandrei:
recheck

ozerovandrei:
recheck

Issue #722, opened 2019-04-03
Compute v2: Add tags to openstack_compute_instance_v2


jtopjian:
This issue is to track the feature request of adding `tags` to the `openstack_compute_instance_v2` resource.

npu21:
Want to know can I create PR for this?
Thanks!

f4rx:
It's a very important feature for using ansible terraform-inventory. I can't use OpenStack with Ansible without host tags (

ozerovandrei:
We need to add the following API support into the upstream Gophercloud library: https://docs.openstack.org/api-ref/compute/?expanded=#server-tags-servers-tags

Gophercloud currently only supports setting tags while creating a server or getting tags from an existing server but there is no support to perform updates on server tags.

kayrus:
#623 can be useful

ozerovandrei:
All upstream logic is ready https://github.com/gophercloud/gophercloud/issues/1669.

Since we don't set any microversions in the current instance resource implementation I can't just re-use it.

I will do the following:
 - check if `tags` are set in Create method and try to use microversion to set `tags` inside create options and return an error if instance creation fails.
 - try to Read `tags` and `all_tags` with microversion and just log errors if it wasn't possible since not all OpenStack environments support the needed microversion.

In that way we won't break the existing compatibility.

ozerovandrei:
This feature was implemented in #899 and will be available in the `1.24.0` release.

PR #899, opened 2019-10-12
Compute V2: add "tags", "all_tags" arguments

ozerovandrei:
Add "tags", "all_tags" arguments into openstack_compute_instance_v2
resource.

For #722


ozerovandrei:
Docs update is needed.

PR #900, opened 2019-10-13
Networking: cleanup subnet V2

ozerovandrei:
Cleanup openstack_networking_subnet resource and datasource.

Implement tests for networking_subnet_v2 helpers.

Change "dns_nameservers" argument form list to set to remove unnecessary
helper that verified if nameservers are distinct.

For #456


ozerovandrei:
recheck

PR #904, opened 2019-10-16
Fixed openstack_lb_loadbalancer_v2 flavor selection.

kmbulebu:
Fixes #903 

Incorporates upstream gophercloud fix.

https://github.com/gophercloud/gophercloud/issues/1743

The openstack_lb_loadbalancer_v2 flavor attribute is now
`flavor_id` to match the Neutron LBaaSv2 and Octavia
api attribute.

PR #907, opened 2019-10-19
Networking V2: add Quotas resource

ozerovandrei:
Implement "openstack_networking_quotas_v2" resource.

Add acceptance tests and documentation.

Vendor newer Gophercloud package.

For #657


ozerovandrei:
Note: tests will fail because of newer Gophercloud until https://github.com/terraform-providers/terraform-provider-openstack/pull/904 is merged.

ozerovandrei:
I tested this on devstack.
Those tests won't run on openlab, see https://github.com/theopenlab/openlab/issues/365

PR #908, opened 2019-10-19
Update lb_pool_v2.html.markdown

Prabhakar-cg:
example states that it supports cookie name for HTTP_COOKIE but terraform wont support cookie_name for HTTP_COOKIE type and the cookie_name feature supports only for APP_COOKIE.
Same has been described in bottom of this page as well.


ozerovandrei:
recheck

PR #920, opened 2019-10-30
Identivy V3: add access rules support for application credentials

### LINKED ISSUE #919: Identity V3: Add access rules field support for application credentials
There is a new feature released in Train:
https://docs.openstack.org/api-ref/identity/v3/#what-s-new-in-version-3-13-train
it allows to grant only explicit service paths and methods.

https://releases.openstack.org/train/highlights.html#keystone-identity-service

### Affected Resource(s)

- openstack_identity_application_credential_v3

### Expected Behavior

Keystone Train users should be able to create fine grained restrictions to application credentials

kayrus:
Resolves #919 
Depends on https://github.com/gophercloud/gophercloud/pull/1758


kayrus:
recheck

kayrus:
```
--- FAIL: TestAccIdentityV3ApplicationCredential_access_rules (1.95s)
    testing.go:569: Step 0 error: Check failed: Check 13/21 error: openstack_identity_application_credential_v3.app_cred_2: Attribute 'access_rules.0.method' expected "POST", got "GET"
```

kayrus:
recheck

kayrus:
recheck

kayrus:
recheck

kayrus:
recheck

kayrus:
Lists change order, when values are read from the API. Sets require custom hash func, because access rules can be set both by an existing ID, or defined as new and setting `Computed: true` for all set elements sets empty hash. Probably it's worth to use lists with custom `DiffSuppressFunc`, which keeps the initial order, defined by a user.

kayrus:
Lists change order, when values are read from the API. Sets require custom hash func, because access rules can be set both by an existing ID, or defined as new and setting `Computed: true` for all set elements sets empty hash. Probably it's worth to use lists with custom `DiffSuppressFunc`, which keeps the initial order, defined by a user.

@jtopjian I believe you already tried to solve the same problem, can you give me some hints before I proceed.

kayrus:
recheck

kayrus:

go.mod
```diff
@@ -1,8 +1,13 @@
 module github.com/terraform-providers/terraform-provider-openstack
 
 require (
+<<<<<<< HEAD
```
ricolin: I don't think you mean to add these

kayrus: it will gone, once squashed and merged

openstack/resource_openstack_identity_application_credential_v3.go
```diff
@@ -70,6 +70,41 @@ func resourceIdentityApplicationCredentialV3() *schema.Resource {
...
+							Required: true,
+							ForceNew: true,
+						},
+
+						"method": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"POST", "GET", "HEAD", "PATCH", "PUT", "DELETE",
```
ozerovandrei: What do you think about using those constants from stdlib?
https://godoc.org/net/http#pkg-constants

ozerovandrei:
@kayrus Looks fine by me. I have only a single question. But I'm happy to merge it as-is if you think that that change is unnecessary.

openstack/resource_openstack_identity_application_credential_v3.go
kayrus: Considering that we use "TCP", "UDP", etc in other resources, and a regular text is more readable than using a const. These methods won't change, but simplify grepping. I'd keep this as is.

PR #921, opened 2019-10-30
Auth: AllowReauth option should be exported

kayrus:
Follow up for the #918 
It should be possible to use this option in terraform [swift backend](https://github.com/hashicorp/terraform/blob/cd7c3e4231c3730b56812881756ac82a6a569583/backend/remote-state/swift/backend.go#L343)


kayrus:
recheck terraform-provider-openstack-acceptance-test

PR #923, opened 2019-11-11
Missing character

r0b2g1t:
Add missing character to documentation for fixing the error:
`Blocks of type "extra" are not expected here. Did you mean to define argument
"extra"? If so, use the equals sign to assign it a value.`

The example lacks an equal sign, which is the reason for the error. The solution is described in the following article. [https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types)


ozerovandrei:
recheck

Issue #925, opened 2019-11-15
Block Storage: new data resource for volume


Shonei:
Hi there,

We are looking into adding a new data resource for volumes.

### Terraform Version
Terraform v0.12.9
provider.openstack v1.23.0

### Affected Resource(s)
openstack_blockstorage_volume_v2

The argument list we are thinking of is:

- region
- name
- state
- metadata
- bootable
- volume_type

And here are the attributes that will get computed:
 - size
 - id
 - source_volume_id

Does this seem acceptable ot should we change some of the parameters.

ozerovandrei:
Any PRs are welcome.

I guess that it will be also possible to implement `openstack_blockstorage_volume_v3` in the same way you will implement `openstack_blockstorage_volume_v2` since their `List()` method and options are similar.

https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/requests.go#L113
https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/volumes/requests.go#L115

Shonei:
Should I include the v2 and v3 in a single pull request.
Also for `bootable` and `volume_type` they are not options in the API but I can filter once I get a list of volumes. Should they be options for retrieving volumes or only part of the return attributes.

ozerovandrei:
@Shonei it's better to split PRs for different datasources.

Sure you can filter all results for needed arguments. We already doing this in some other datasources like `compute_flavor_v2`.

Shonei:
As both PR have been merged I will close this.

PR #928, opened 2019-11-25
Data source for openstack_blockstorage_volume_v2

Shonei:
Implements a new data source for openstack_blockstorage_volume_v2. Here is the related issue opened for it
#925


Shonei:
The change to `openstack/resource_openstack_sharedfilesystem_sharenetwork_v2_test.go` was done to fix a failing test. Let me know if it should be separate Pull Request.

openstack/data_source_openstack_blockstorage_volume_v2.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if len(allVolumes) < 1 {
+		return fmt.Errorf("Your openstack_blockstorage_volume_v2 query returned no results.")
+	}
+
+	return dataSourceBlockStorageSnapshotV2Volume(d, allVolumes[0])
+}
+
+func dataSourceBlockStorageSnapshotV2Volume(d *schema.ResourceData, volume volumes.Volume) error {
```
ozerovandrei: I guess it's better rename this method to `dataSourceBlockStorageVolumeV2Attributes` so it will have the same pattern as other blockstorage datasources:
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_blockstorage_snapshot_v2.go#L111
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/data_source_openstack_blockstorage_snapshot_v3.go#L111

Shonei: Update function name to `dataSourceBlockStorageVolumeV2Attributes`

PR #930, opened 2019-11-26
Internal: Decouple Authentication

### LINKED ISSUE #686: Proposal: Split auth logic into its own module/repository

### LINKED ISSUE #929: Core: use GetEndpointType helper from "clientconfig"
### Expected Behavior

In order to reduce code duplication, OpenStack terraform provider should use exported helper function to determine endpoint type

### Actual Behavior

OpenStack terraform provider uses its own unexported func.

### References
- https://github.com/gophercloud/utils/pull/97

kayrus:
Resolves #686
Supersedes #688


kayrus:
recheck

PR #937, opened 2019-12-06
Add containerinfra_cluster kubeconfig attribute

andrein:
This patch adds a kubeconfig attribute to the openstack_containerinfra_cluster_v1 resource.

I'd appreciate any feedback on it.


andrein:
Looks like stack.sh is failing in openlab, can someone take a look?
```
+ ./stack.sh:main:224                      :   [[ ! xenial =~ (bionic|stretch|jessie|f29|opensuse-15.0|opensuse-15.1|opensuse-tumbleweed|rhel7) ]]
+ ./stack.sh:main:225                      :   echo 'WARNING: this script has not been tested on xenial'
WARNING: this script has not been tested on xenial
+ ./stack.sh:main:226                      :   [[ '' != \y\e\s ]]
+ ./stack.sh:main:227                      :   die 227 'If you wish to run this script anyway run with FORCE=yes'
+ functions-common:die:195                 :   local exitcode=0
+ functions-common:die:196                 :   set +o xtrace
[Call Trace]
./stack.sh:227:die
[ERROR] ./stack.sh:227 If you wish to run this script anyway run with FORCE=yesnon-zero return code
```
https://logs.openlabtesting.org/logs/37/937/def041de9d3ed49715dbf1fa0ade6e68a414934b/check/terraform-provider-openstack-acceptance-test/cc904bf/logs/devstack-early.txt.gz

kayrus:
@andrein you can take a look on how do we export kubernetes credentials in [kubernikus](https://github.com/sapcc/kubernikus) kubernetes as a service terraform resource:
* https://github.com/sapcc/terraform-provider-ccloud/blob/master/website/docs/r/kubernetes_v1.html.markdown#attributes-reference
* https://github.com/sapcc/terraform-provider-ccloud/blob/32576f06e2901074eada89924490808dc9e60a30/ccloud/resource_ccloud_kubernetes_v1.go#L442
* https://github.com/sapcc/terraform-provider-ccloud/blob/32576f06e2901074eada89924490808dc9e60a30/ccloud/ccloud_kubernetes_v1.go#L342..L427

We used the Azure concept: https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html

andrein:
Hi @ozerovandrei, thanks for the feedback. Answers to your questions below:

> 1. Why introduce changes in `resourceContainerInfraClusterV1Update()` in this PR? I don't see how it's related to the `kubeconfig` attribute.

I initially did that for backwards compatibility. If you had a cluster created with the previous version of the provider, terraform detected changes to the schema and tries to update the cluster, but the update was a no-op and the magnum api returned an error. The changes for the third question seem to address this issue as well, so I've reverted the change.

> 2. It's better to move `"RSA PRIVATE KEY"` and `"CERTIFICATE REQUEST"` values for `pem.Block` into separate constants.

Done

> 3. Why do you use `[]map[string]interface{}` for `flattenContainerInfraV1Kubeconfig()` return value and `schema.TypeList` for `kubeconfig` attribute instead of using simple `map[string]interface{}`?

I was using schema.TypeList there because of a limitation mentioned in [the guide](https://www.terraform.io/docs/extend/writing-custom-providers.html#implementing-a-more-complex-read) (https://github.com/hashicorp/terraform/issues/11115) I just realized we're not actually hitting that limitation, so I've changed it to schema.typeMap. I wonder why most of the other providers do it with schema.TypeList, though? Did other people misread the docs just as I did?

> I also think that `kubeconfig` shouldn't be introduced as an attribute of a `cluster` resource since we always create a new PKI objects: Certificate, Key and CSR.
> We don't need to create them at every `resourceContainerInfraClusterV1Read()` call.
> I think that it's better to create a separate `resource_openstack_containerinfra_kubeconfig_v1` resource. Unfortunately we should make `Delete` method of this resource as no-op operation since Magnum API can't delete the created certificates.

I would like to have them available in the same resource. That's how the google and digitalocean providers do it. We're not actually generating new certs for every read call (https://github.com/terraform-providers/terraform-provider-openstack/pull/937/commits/ce5fd39370a5ee5be367a3d02a2b5efb68d20c5d#diff-fc5879d32acf215c2b41cc440ce5c7b4R136) That check could be extended to check for certificate validity, but since magnum currently issues certificates valid for 5 years I didn't bother with it. I will revisit this when [this upstream story](https://storyboard.openstack.org/#!/story/2006991) gets implemented.

ozerovandrei:
> We're not actually generating new certs for every read call (ce5fd39#diff-fc5879d32acf215c2b41cc440ce5c7b4R136)

But in case of `d.Get("kubeconfig.client_certificate").(string) == ""` we will run the `certificates.Get()`, `rsa.GenerateKey` and all other methods in `flattenContainerInfraV1Kubeconfig()` function.

I'm also not sure what happens if we have an empty map for `kubeconfig` argument. Could terraform panic in case of the `nil` `kubeconfig` map?

andrein:
> > We're not actually generating new certs for every read call ([ce5fd39](https://github.com/terraform-providers/terraform-provider-openstack/commit/ce5fd39370a5ee5be367a3d02a2b5efb68d20c5d)#diff-fc5879d32acf215c2b41cc440ce5c7b4R136)
> 
> But in case of `d.Get("kubeconfig.client_certificate").(string) == ""` we will run the `certificates.Get()`, `rsa.GenerateKey` and all other methods in `flattenContainerInfraV1Kubeconfig()` function.

I'm assuming if the certificate isn't set, the CA and private key aren't either. Should I check every key?

> I'm also not sure what happens if we have an empty map for `kubeconfig` argument. Could terraform panic in case of the `nil` `kubeconfig` map?

It won't crash if it's empty, and it can't be nil according to https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/schema/resource_data.go#L63 (it only returns nil if the key doesn't exist in the schema, which is not the case here)

andrein:
Hi @ozerovandrei, thanks for the feedback.

I've pushed a new version with updated docs and a new attribute `raw_config` which contains the full kubeconfig file the user can use with kubectl. The implementation is mostly copied from the DigitalOcean provider. 

I've also added the safety check in `resourceContainerInfraClusterV1Update`, I must've tested a different version before, because I started getting errors again when running terraform apply on an existing cluster.

Before this is merged, I'd like to update the acceptance tests. I'm thinking of adding a few `TestMatchResourceAttr` checks for the fields I added. I'd appreciate any ideas, I'm quite new at this :)

Thanks,
Andrei

andrein:
@ozerovandrei I've rebased on master

PR #947, opened 2019-12-18
Data source for openstack_blockstorage_volume_v3

Shonei:
Implements a new data source for openstack_blockstorage_volume_v3. Here is the related issue opened for it
#925


openstack/data_source_openstack_blockstorage_volume_v2.go
```diff
@@ -5,7 +5,6 @@ import (
 	"log"
 
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes"
-
```
ozerovandrei: I guess that we need this empty line given that we use it to separate terraform and gophercloud imports in other files.

ozerovandrei: And this file is unrelated to this PR.

Shonei:
> I made some styling fixes, waiting for tests to finish.
> I don't see any more issues so I'll merge this after successful tests run.

Thanks for that. Sorry I couldn't fix it sooner.

PR #962, opened 2020-02-04
Networking V2: Fix port's fixed_ip panic, when subnet is set to an empty string

### LINKED ISSUE #961: Networking V2: Setting a port's fixed_ip subnet_id to an empty string causes a panic
### Affected Resource(s)

Initial config:

```hcl
resource "openstack_networking_port_v2" "port_1" {
  name = "port_1"
  network_id = "b9a56df5-8db1-4856-a4a6-ce77f6d15081"
  fixed_ip {
    subnet_id = "052ba0b2-dc0d-4536-a269-1340400a6e68"
  }
}
```

then change subnet to an empty string:

```hcl
resource "openstack_networking_port_v2" "port_1" {
  name = "port_1"
  network_id = "b9a56df5-8db1-4856-a4a6-ce77f6d15081"
  fixed_ip {
    subnet_id = ""
  }
}
```

### Panic Output

https://gist.github.com/kayrus/b1078ca24f3f0012faa861432eafdebe

### Expected Behavior
terraform should not crash

### Actual Behavior
terraform crashes

kayrus:
resolves #961

### DIFF
--- a/openstack/networking_port_v2.go
+++ b/openstack/networking_port_v2.go
@@ -168,12 +168,14 @@ func expandNetworkingPortFixedIPV2(d *schema.ResourceData) interface{} {
 		return nil
 	}
 
-	ip := make([]ports.IP, len(rawIP))
-	for i, raw := range rawIP {
-		rawMap := raw.(map[string]interface{})
-		ip[i] = ports.IP{
-			SubnetID:  rawMap["subnet_id"].(string),
-			IPAddress: rawMap["ip_address"].(string),
+	var ip []ports.IP
+	for _, raw := range rawIP {
+		if raw != nil {
+			rawMap := raw.(map[string]interface{})
+			ip = append(ip, ports.IP{
+				SubnetID:  rawMap["subnet_id"].(string),
+				IPAddress: rawMap["ip_address"].(string),
+			})
 		}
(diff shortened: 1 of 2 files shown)


kayrus:

ozerovandrei (dismissed):
I don't think that we should always require `subnet_id` inside of `fixed_ips` block since Neutron V2 API allows to omit it:
https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-detail#create-port
```
Each entry consists of IP address (ip_address) and the subnet ID from which the IP address is assigned (subnet_id).
If you specify both a subnet ID and an IP address, OpenStack Networking tries to allocate the IP address on that subnet to the port.
If you specify only a subnet ID, OpenStack Networking allocates an available IP from that subnet to the port.
If you specify only an IP address, OpenStack Networking tries to allocate the IP address if the address is a valid IP for any of the subnets on the specified network.
```

devops-cd-allot:
Hi
still ant create a port if leaving the subnet_id empty - TF crushes.
expecting to have it since we need the port to have multiple conditions such as IPv4 only, IPv6 only or Dual-stack.

nikParasyr (approved):
LGTM

Issue #736, opened 2019-04-16
add support for volume_type in block_device of openstack_compute_instance_v2


rptaylor:
Hello,

According to the  openstack_compute_instance_v2 documentation:
https://www.terraform.io/docs/providers/openstack/r/compute_instance_v2.html#device_type
the block_device block does not support the volume_type option.

However,  openstack_blockstorage_volume_v1 does support volume_type:
https://www.terraform.io/docs/providers/openstack/r/blockstorage_volume_v1.html#volume_type

When implicitly including a volume as part of a compute instance specification, can support for volume_type be added?

As a work around it should be possible to explicitly create a openstack_blockstorage_volume_v1 with volume_type specified, but this requires the additional step of attaching particular volumes to particular instances. It is much easier to use the inline block_device declaration.

### Terraform Version
```
Terraform v0.11.13
+ provider.null v2.1.0
+ provider.openstack v1.17.0
```

### Affected Resource(s)
openstack_compute_instance_v2
openstack_blockstorage_volume_v1 

### Terraform code
```
resource "openstack_compute_instance_v2" "k8s_node" {
  name       = "${var.cluster_name}-k8s-node-${count.index+1}"
  count      = "${var.number_of_k8s_nodes}"
  availability_zone = "${element(var.az_list_kubelet, count.index)}"
  image_name = "${var.image}"
  flavor_id  = "${var.flavor_k8s_node}"
  key_pair   = "${openstack_compute_keypair_v2.k8s.name}"

  network {
    name = "${var.network_name}"
  }

  security_groups = ["${openstack_networking_secgroup_v2.k8s.name}",
    "${openstack_networking_secgroup_v2.bastion.name}",
    "${openstack_networking_secgroup_v2.worker.name}",
    "default",
  ]

  block_device {
    uuid                  = "${var.image}"
    source_type           = "image"
    volume_size           = "${var.node_volume_size_in_gb}"
    boot_index            = 0
    destination_type      = "volume"
    delete_on_termination = true
    volume_type           = "Database"
  }

}
```

### Expected Behavior
The volume should be created with the specified type.

### Actual Behavior
The specified volume_type is ignored.

jtopjian:
> When implicitly including a volume as part of a compute instance specification, can support for volume_type be added?

Yes, it can be added.

It looks like a recent release of OpenStack will allow this: https://docs.openstack.org/nova/latest/user/block-device-mapping.html

[...]

PR #963, opened 2020-02-05
implement support for volume_type in block_device of openstack_compute_instance_v2

rptaylor:
For https://github.com/terraform-providers/terraform-provider-openstack/issues/736

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -247,6 +247,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 							Optional: true,
 							ForceNew: true,
 						},
+						"volume_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ForceNew: true,
+						},
 						"device_type": {
 							Type:     schema.TypeString,
 							Optional: true,
@@ -482,6 +487,11 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
 		createOpts = &bootfromvolume.CreateOptsExt{
 			CreateOptsBuilder: createOpts,
 			BlockDevice:       blockDevices,
+		}
+
+		specifyVolumeType := d.Get("volume_type").(bool)
+		if specifyVolumeType {
+			computeClient.Microversion = computeV2InstanceBlockDeviceVolumeTypeMicroversion
 		}
 	}
 
@@ -1055,6 +1065,7 @@ func resourceOpenStackComputeInstanceV2ImportState(d *schema.ResourceData, meta 
 				"source_type":           "image",
 				"volume_size":           volMetaData.Size,
 				"disk_bus":              "",
+				"volume_type":           "",
 				"device_type":           "",
 			}
 
@@ -1115,6 +1126,7 @@ func resourceInstanceBlockDevicesV2(d *schema.ResourceData, bds []interface{}) (
 			BootIndex:           bdM["boot_index"].(int),
 			DeleteOnTermination: bdM["delete_on_termination"].(bool),
 			GuestFormat:         bdM["guest_format"].(string),
+			VolumeType:          bdM["volume_type"].(string),
 			DeviceType:          bdM["device_type"].(string),
 			DiskBus:             bdM["disk_bus"].(string),
 		}
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -1352,7 +1352,7 @@ func checkBlockDeviceConfig(d *schema.ResourceData) error {
 				}
 			}
 
-			if vM["volume_type"] != ""  {
+			if vM["volume_type"] != "" {
 				computeClient.Microversion = computeV2InstanceBlockDeviceVolumeTypeMicroversion
 			}
 		}
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -484,6 +484,14 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
 			return err
 		}
 
+		// Check if VolumeType was set in any of the Block Devices.
+		// If so, set the client's microversion appropriately.
+		for _, bd := range blockDevices {
+			if bd.VolumeType != "" {
+				computeClient.Microversion = computeV2InstanceBlockDeviceVolumeTypeMicroversion
+			}
+		}
+
 		createOpts = &bootfromvolume.CreateOptsExt{
 			CreateOptsBuilder: createOpts,
 			BlockDevice:       blockDevices,
@@ -1350,10 +1358,6 @@ func checkBlockDeviceConfig(d *schema.ResourceData) error {
 				if vM["volume_size"] == 0 {
 					return fmt.Errorf("You must specify a volume_size when creating a blank block device")
 				}
-			}
-
-			if vM["volume_type"] != "" {
-				computeClient.Microversion = computeV2InstanceBlockDeviceVolumeTypeMicroversion
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


rptaylor:
@jtopjian Does this look right? I kind of copied what there is for device_type.

rptaylor:
Should the microversion check be something like this?
```
                        if vM["volume_type"] != "" {
                                computeClient.Microversion = computeV2InstanceBlockDeviceVolumeTypeMicroversion
                        }
```

perhaps around line 486 of resource_openstack_compute_instance_v2.go  ?
How do I get the volume_type string from blockDevices ?

jtopjian:
@rptaylor I _think_ this should work: https://github.com/jtopjian/terraform-provider-openstack/commit/6b693a4fe5c4b7748ce5079fdf20ec6c97b2abe5

The `checkBlockDeviceConfig` function is only doing a sanity check for correct combinations of block device mappings and not actually configuring a block device for creation. That's done in the `resourceInstanceBlockDevicesV2` function.

Are you able to compile this and test? Compiling is pretty easy once you have `go` installed. Just do:

```
make build
```

And the resulting binary should be in `~/go/bin/terraform-provider-openstack`. You can then copy this alongside the `terraform` binary (removing any other `terraform-provider-openstack` binary installed) and the Terraform should automatically detect it.

@ozerovandrei Thoughts on this diff (other than missing a doc entry)? The loop for setting the microversion feels a little clunky, but I can't think of a better way of doing it. As well, there might be an issue down the road if another microversion check is made _after_ this which sets the microversion lower. The logic will work for this patch, but in the future, there probably needs to be a some logic which won't lower the microversion if microversion is currently not an empty string.

ozerovandrei:
@jtopjian yeah those microversions are doing more harm than good. Never understood why even implement them on OpenStack API side.

I guess that current changes are good and can be merged. We can think about some helpers to select microversions in the future if there is more of them.

@rptaylor can you please update docs for the instance resource?
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/docs/r/compute_instance_v2.html.markdown

And is it possible to implement acceptance test for it?
https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/resource_openstack_compute_instance_v2_test.go

jtopjian:
> And is it possible to implement acceptance test for it?

An acceptance test might be a little tricky. The test environment would need to have a Volume Type created and I don't think one is created in devstack automatically. This is probably why there aren't any existing tests for volume types in the Block Storage resources.

If we can get confirmation that this works from someone testing it on their own, that might be the best we can do.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1055,6 +1068,7 @@ func resourceOpenStackComputeInstanceV2ImportState(d *schema.ResourceData, meta
 				"source_type":           "image",
 				"volume_size":           volMetaData.Size,
 				"disk_bus":              "",
+				"volume_type":           "",
```
rptaylor: Should this line be added?

rptaylor:
@jtopjian @ozerovandrei thanks a lot for your help!

I added documentation. I updated the microversion check to match what Joe suggested.
I am not sure if I should have line 1071 though.

Yes it compiles fine. I will test it tomorrow.

openstack/resource_openstack_compute_instance_v2.go
jtopjian: Yes, and I think it's safe to leave as an empty string for now. It might be possible to determine the volume type from an attached block storage volume via the Nova/Compute API, but I don't have access to determine that right now. I say leave as-is and if anyone else has additional information, they can patch this accordingly.

rptaylor:
I installed the terraform-provider-openstack that I built, and `terraform plan` was successful.
The `terraform apply` succeeded up until 

```
Error: Error creating OpenStack server: Expected HTTP response code [200 202] when accessing [POST https://cloud:8774/v2.1/os-volumes_boot], but got 406 instead
{"computeFault": {"message": "Version 2.67 is not supported by the API. Minimum is 2.1 and maximum is 2.65.", "code": 406}}

  on ../../kubespray/contrib/terraform/openstack/modules/compute/main.tf line 437, in resource "openstack_compute_instance_v2" "k8s_master_no_floating_ip_custom_volume_size":
 437: resource "openstack_compute_instance_v2" "k8s_master_no_floating_ip_custom_volume_size" {
```

which is expected because the current version of our Openstack cloud is not new enough. Hopefully by the summer it will be upgraded. I don't have any other Openstack cloud to test on at the moment...

Is the devstack cloud new enough, and does it provide any custom volume types? @ozerovandrei

jtopjian:
It turns out my dev environment is running a compatible version and I was able to get this working:

### creating a volume type

```shell
$ openstack volume type create --public foo
```

### create a basic instance with a normal ephemeral root disk and an attached block device of type "foo"

```hcl
resource "openstack_compute_instance_v2" "test" {
  name        = "test"
  flavor_name = "m1.acctest"
  image_name  = "cirros"

  network {
    name = "demo-net"
  }

  block_device {
    uuid                  = "37a45963-df17-4dcf-818d-4c0bdf3fcf56"
    source_type           = "image"
    destination_type      = "local"
    boot_index            = 0
    delete_on_termination = true
  }

  block_device {
    source_type           = "blank"
    destination_type      = "volume"
    volume_size           = 1
    boot_index            = 1
    delete_on_termination = true
    volume_type           = "foo"
  }
}
```

### debug logs showing the microversion was correctly set and the request was accepted

```
2020/02/13 03:20:28 [DEBUG] OpenStack Request URL: POST http://10.1.0.235:8774/v2.1/1e65fffb73a94e33a3547b95071abfc9/os-volumes_boot
2020/02/13 03:20:28 [DEBUG] OpenStack Request Headers:
Accept: application/json
Cache-Control: no-cache
Openstack-Api-Version: compute 2.67
User-Agent: HashiCorp Terraform/0.12.20 (+https://www.terraform.io) Terraform Plugin SDK/1.4.1 gophercloud/2.0.0
X-Auth-Token: ***
X-Openstack-Nova-Api-Version: 2.67
2020/02/13 03:20:28 [DEBUG] OpenStack Request Body: {
  "server": {
    "block_device_mapping_v2": [
      {
        "boot_index": 0,
        "delete_on_termination": true,
        "destination_type": "local",
        "source_type": "image",
        "uuid": "37a45963-df17-4dcf-818d-4c0bdf3fcf56"
      },
      {
        "boot_index": 1,
        "delete_on_termination": true,
        "destination_type": "volume",
        "source_type": "blank",
        "volume_size": 1,
        "volume_type": "foo"
      }
    ],
    "config_drive": false,
    "flavorRef": "99",
    "imageRef": "37a45963-df17-4dcf-818d-4c0bdf3fcf56",
    "name": "test",
    "networks": [
      {
        "uuid": "94b3121b-3d55-4e51-95e7-cbbab06aa4af"
      }
    ],
    "user_data": ""
  }
}
2020/02/13 03:20:30 [DEBUG] OpenStack Response Code: 202
```

### volume was successfully created

```
$ openstack volume list

+--------------------------------------+------------------------------------------------+--------+------+-------------------------------+
| ID                                   | Name                                           | Status | Size | Attached to                   |
+--------------------------------------+------------------------------------------------+--------+------+-------------------------------+
| 83b06231-4623-4fd0-890c-7fa4dbcbb896 | 53227306-ebfd-4c1c-99c8-5d96d9e77db7-blank-vol | in-use |    1 | Attached to test on /dev/vdc  |
+--------------------------------------+------------------------------------------------+--------+------+-------------------------------+
```

### and is of type "foo"

```
$ openstack volume show 83b06231-4623-4fd0-890c-7fa4dbcbb896

... snip ...

| type                           | foo

... snip ...

```

And upon `terraform delete`, the volume was cleanly deleted.

@ozerovandrei if you see any changes that need made, I'm happy to test them.

rptaylor:
Awesome, thanks @jtopjian !!

jtopjian:
LGTM from my side.

rptaylor:
Okay great, thanks! Ready to merge.

-rt

PR #978, opened 2020-03-11
Add Project Tags

randomswdev:
Add support for assigning tags to OpenStack projects. The feature is available in the OpenStack REST APIs since Queen.


ozerovandrei:
Looks fine but can you please also update website documentation in https://github.com/terraform-providers/terraform-provider-openstack/tree/master/website

You should also vendor newer Gophercloud since we use vendor for this project.

PR #984, opened 2020-03-24
Add compute_instance_v2 data source

### LINKED ISSUE #982: Feature request: create `data_source_openstack_compute_instance_v2`
Hi there,

### Terraform Version

0.12.24

### Affected Resource(s)

n/a, requesting a new feature

```hcl
data "openstack_compute_instance_v2" "instance" {
  name = "instance_name"
}
```

### Expected Behavior

n/a, feature request

### Actual Behavior
n/a, feature request

### References

The existing `resource_openstack_compute_instance_v2.go` would contain everything this needs, I think? I don't know enough golang to know if this is true, though.

aurynn:
Resolves #982 

This is largely copied from `resource.openstack_compute_instance_v2`.

I've checked the data source itself locally, but I have no way to run the full test suite, so relying on project CI.

### DIFF
--- a/openstack/data_source_openstack_compute_instance_v2.go
+++ b/openstack/data_source_openstack_compute_instance_v2.go
@@ -1,0 +1,230 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/tags"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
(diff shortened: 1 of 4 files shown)


kayrus:
Resolves #717 as well

aurynn:
Was able to validate and fix the test locally.

aurynn:
@ozerovandrei Yep, I think this is ready to be merged in. 

I definitely understand the desire to not add too much that'll have to be migrated over when you start implementing the new `openstack_compute_server_v2` API. Do you have any ideas when/if/etc you'll have the capacity to start picking up that work/maintaining this provider more?

ozerovandrei:
> Do you have any ideas when/if/etc you'll have the capacity to start picking up that work/maintaining this provider more?

Unfortunately I don't 😞

PR #986, opened 2020-03-26
Update object verisoning documentation URL

aurynn:
The old URL was no longer valid; updated to the current URL.


ozerovandrei:
recheck

PR #987, opened 2020-03-26
Add ACL examples

aurynn:
Adds some examples of how the ACLs should be used, to make it clearer what's going on.


ozerovandrei:
recheck

PR #993, opened 2020-04-13
Add SOURCE_IP_PORT load balancing method

mtneug:
The OVN load balancing provider only supports `SOURCE_IP_PORT` as load balancing method for pools.


kayrus:
@mtneug `SOURCE_IP_PORT` is available only in octavia. You need to add this note into a doc.

mtneug:
Thanks. Updated the docs.

PR #997, opened 2020-04-24
Keymanager V1: Print error log in keyManagerSecretV1GetPayload func

### LINKED ISSUE #991: Keymanager V1: keyManagerSecretV1GetPayload func doesn't print errors on failure
https://github.com/terraform-providers/terraform-provider-openstack/blob/25ef7986c56a9bf2d106bfcf59cc18c311a375df/openstack/keymanager_secret_v1.go#L99..L105

There should be a log message or an explicit error return with further acknowledge logic.

```
diff --git a/openstack/keymanager_secret_v1.go b/openstack/keymanager_secret_v1.go                                     
index 9426058b..3869b0b9 100644                                                                                        
--- a/openstack/keymanager_secret_v1.go                                                                                
+++ b/openstack/keymanager_secret_v1.go                                                                                
@@ -99,7 +99,7 @@ func keyManagerSecretMetadataV1WaitForSecretMetadataCreation(kmClient *gopherclo                     
 func keyManagerSecretV1GetPayload(kmClient *gophercloud.ServiceClient, id string) string {                            
        payload, err := secrets.GetPayload(kmClient, id, nil).Extract()                                                
        if err != nil {                                                                                                
-               fmt.Errorf("Could not retrieve payload for secret with id %s: %s", id, err)                            
+               log.Printf("[DEBUG] Could not retrieve payload for secret with id %s: %s", id, err)                    
        }                                                                                                              
        return string(payload)                                                                                         
 }
```

kayrus:
Resolves #991

### DIFF
--- a/openstack/keymanager_secret_v1.go
+++ b/openstack/keymanager_secret_v1.go
@@ -99,7 +99,7 @@ func keyManagerSecretMetadataV1WaitForSecretMetadataCreation(kmClient *gopherclo
 func keyManagerSecretV1GetPayload(kmClient *gophercloud.ServiceClient, id string) string {
 	payload, err := secrets.GetPayload(kmClient, id, nil).Extract()
 	if err != nil {
-		fmt.Errorf("Could not retrieve payload for secret with id %s: %s", id, err)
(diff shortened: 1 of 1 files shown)


kayrus:

PR #1024, opened 2020-07-01
Add missing floating_ip_enabled and merge_labels option to openstack_containerinfra_cluster_v1

brtknr:


brtknr:
FYI @openstacker

openstack/resource_openstack_containerinfra_cluster_v1.go
```diff
@@ -189,6 +189,20 @@ func resourceContainerInfraClusterV1() *schema.Resource {
...
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
+
+			"merge_labels": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
```
ozerovandrei, line 203: It should not be computed since it can't be populated from API responses in the `resourceContainerInfraClusterV1Read `.

brtknr: Thanks for the review, I have pushed a new commit with the requested amendments.

brtknr:
Yes I think so

Issue #1025, opened 2020-07-08
Feature proposal: allow sharing a network based on a network search


randomswdev:
# Feature proposal

### Terraform Version
v0.12.24

### Affected Resource(s)
openstack_networking_rbac_policy_v2

### Terraform Configuration Files
```hcl
resource "openstack_networking_rbac_policy_v2" "rbac_policy_1" {
  action        = "access_as_shared"
  object_search  {
     name = "network-*-24"
     already_shared = false
  }
  object_type     = "network"
  target_tenant = "20415a973c9e45d3917f078950644697"
}
```

### Expected Behavior
The expression above should allow searching for a network whos ename matches the glob expression included in object_search.name and that is not shared yet. This network is then shared to the target tenant.

This would allow selecting and sharing a network from  a set of already allocated networks.

What do you think? I can contribute the code if you think this could be an itneresting feature to add.

PR #1026, opened 2020-07-08
Add the ability to search ofr network to share when creating an RBAC

randomswdev:
Pull request to implement the proposal in the issue https://github.com/terraform-providers/terraform-provider-openstack/issues/1025

Refer to the doc changes for an high level view of the changes to the rbac resource.


kayrus:
@randomswdev can the data source be used for the object_id search?

kayrus:
It would be better to implement such functionality with data sources.

Issue #895, opened 2019-10-10
Identity V3: Group Resource


LmV4corbeau:
Hi, people,

thank you so much for the great work.

Would it be possible to create a resource for Openstack Identity Groups to create, update and delete groups?

### Terraform Version
Terraform v0.12.9
+ provider.openstack v1.23.0

### Affected Resource(s)
openstack_identity_group_v3

ozerovandrei:
It's possible and upstream Gophercloud library already provides all needed handlers for it: https://github.com/gophercloud/gophercloud/tree/master/openstack/identity/v3/groups

ozerovandrei:
Implemented in #1028

PR #1028, opened 2020-07-10
feat: add openstack_identity_group_v3 resource

c35sys:
This PR allows to create openstack_identity_group_v3 resource, referring to #895

### DIFF
--- a/openstack/resource_openstack_identity_group_v3.go
+++ b/openstack/resource_openstack_identity_group_v3.go
@@ -1,0 +1,125 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/groups"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceIdentityGroupV3() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceIdentityGroupV3Create,
+		Read:   resourceIdentityGroupV3Read,
+		Update: resourceIdentityGroupV3Update,
+		Delete: resourceIdentityGroupV3Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"domain_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceIdentityGroupV3Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	createOpts := groups.CreateOpts{
+		DomainID: d.Get("domain_id").(string),
+		Name:     d.Get("name").(string),
+	}
+
+	log.Printf("[DEBUG] openstack_identity_group_v3 create options: %#v", createOpts)
+	group, err := groups.Create(identityClient, createOpts).Extract()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


openstack/resource_openstack_identity_group_v3_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckIdentityV3GroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccIdentityV3Group_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3GroupExists("openstack_identity_group_v3.group_1", &group),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_group_v3.group_1", "name", &group.Name),
```
ozerovandrei, line 27: It's better to check if the name is "group_1" here.

openstack/resource_openstack_identity_group_v3_test.go
```diff
@@ -0,0 +1,105 @@
...
+				),
+			},
+			{
+				Config: testAccIdentityV3Group_update,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3GroupExists("openstack_identity_group_v3.group_1", &group),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_group_v3.group_1", "name", &group.Name),
```
ozerovandrei, line 35: It's better to check if the name is "group_2" here so we will be sure that name was changed.

openstack/resource_openstack_identity_group_v3_test.go
c35sys: It is the case with `Config: testAccIdentityV3Group_basic` which is using `const testAccIdentityV3Group_basic`, `name` is the group name attribute.

openstack/resource_openstack_identity_group_v3_test.go
c35sys: Same thing it is the case with `Config: testAccIdentityV3Group_update` which is using `const testAccIdentityV3Group_update`.

openstack/resource_openstack_identity_group_v3.go
```diff
@@ -0,0 +1,125 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"domain_id": {
```
kayrus, line 29: According to keystone api ref

> The ID of the new domain for the group. The ability to change the domain of a group is now deprecated, and will be removed in subsequent release. It is already disabled by default in most Identity service implementations.

therefore please add `ForceNew: true,` to `domain_id`.

openstack/resource_openstack_identity_group_v3.go
```diff
@@ -0,0 +1,125 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
```
kayrus, line 39: Can you add an optional description field?

openstack/resource_openstack_identity_group_v3.go
c35sys: done

openstack/resource_openstack_identity_group_v3.go
c35sys: done

kayrus: description is not reflected in the update function.

c35sys: done

website/docs/r/identity_group_v3.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+resource "openstack_identity_group_v3" "group_1" {
+  name = "group_1"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - The name of the group.
```
kayrus, line 28: I'm sorry to bother you again, but could you please reflect all the code changes you made by my request in the documentation? Thanks for your patience :)

c35sys: no worries, I'll do that

c35sys: done

Issue #1016, opened 2020-05-30
Support retrieving EC2 credentials


ninja-:
It would be nice to get support like when using CLI

```
openstack ec2 credentials create
```

This would allow to use S3-compatible API on Swift with proper credentials and inject the credentials into other parts of the project...

kayrus:
@ninja- gophercloud supports ec2 credentials: https://pkg.go.dev/github.com/gophercloud/gophercloud@v0.11.0/openstack/identity/v3/extensions/ec2credentials?tab=doc
You can make a terraform resource following the application credentials resource code: https://github.com/terraform-providers/terraform-provider-openstack/blob/242ffd8f8b807092dbf8a53cff6b47f97cb0cc33/openstack/resource_openstack_identity_application_credential_v3.go

ninja-:
Are you saying it's supported or just pointing me in direction on how to add support by modifying source code?

kayrus:
@ninja- i'm pointing you to the direction to add the support following the similar code, which is already implemented.

ninja-:
Alright, thanks

kayrus:
@ninja- @ozerovandrei this issue can be closed

ozerovandrei:
@ninja- `identity_ec2_credential` resource was added and it's available starting from the `1.31.0` release of the provider.
@pbasov thank you for implementing that.

PR #1033, opened 2020-07-28
Add identity_ec2_credential resource

pbasov:
* allows creating EC2 credentials for S3-compatible Swift/RadosGW endpoints

### DIFF
--- a/openstack/resource_openstack_identity_ec2_credential_v3.go
+++ b/openstack/resource_openstack_identity_ec2_credential_v3.go
@@ -1,0 +1,238 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/identity/v2/tenants"
+	tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/extensions/ec2credentials"
+	tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceIdentityEc2CredentialV3() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceIdentityEc2CredentialV3Create,
+		Read:   resourceIdentityEc2CredentialV3Read,
+		Delete: resourceIdentityEc2CredentialV3Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"access": {
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"secret": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+				ForceNew:  true,
+			},
+
+			"project_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"user_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"trust_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceIdentityEc2CredentialV3Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	user, project, err := GetTokenInfo(identityClient)
+	if err != nil {
+		return fmt.Errorf("Error getting token info: %s", err)
+	}
+
+	var tenantID string
+	// Set tenant to user value if provided, get one from token otherwise
+	if definedProject, ok := d.GetOk("project_id"); ok {
+		tenantID = definedProject.(string)
+	} else {
+		tenantID = project
+	}
+
+	createOpts := ec2credentials.CreateOpts{
+		TenantID: tenantID,
+	}
+
+	var userID string
+	// Set userid to defined tf value, use one from token otherwise
+	if definedUser, ok := d.GetOk("user_id"); ok {
+		userID = definedUser.(string)
+	} else {
+		userID = user
+	}
+
+	log.Printf("[DEBUG] openstack_identity_ec2_credential_v3 create options: %#v", createOpts)
+
+	ec2Credential, err := ec2credentials.Create(identityClient, userID, createOpts).Extract()
+
+	if err != nil {
+		if v, ok := err.(gophercloud.ErrDefault404); ok {
+			return fmt.Errorf("Error creating openstack_identity_ec2_credential_v3: %s", v.ErrUnexpectedResponseCode.Body)
+		}
+		return fmt.Errorf("Error creating openstack_identity_ec2_credential_v3: %s", err)
+	}
+
+	d.SetId(ec2Credential.Access)
+
+	d.Set("access", ec2Credential.Access)
+	d.Set("secret", ec2Credential.Secret)
+	d.Set("user_id", ec2Credential.UserID)
+	d.Set("project_id", ec2Credential.TenantID)
+	d.Set("trust_id", ec2Credential.TrustID)
+
+	return resourceIdentityEc2CredentialV3Read(d, meta)
+}
+
+func resourceIdentityEc2CredentialV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	user, _, err := GetTokenInfo(identityClient)
+	if err != nil {
+		return fmt.Errorf("Error getting token info: %s", err)
+	}
+
+	var userID string
+	// Set userid to defined tf value, use one from token otherwise
+	if definedUser, ok := d.GetOk("user_id"); ok {
+		userID = definedUser.(string)
+	} else {
+		userID = user
+	}
+
+	ec2Credential, err := ec2credentials.Get(identityClient, userID, d.Id()).Extract()
+	if err != nil {
+		return CheckDeleted(d, err, "Error retrieving openstack_identity_ec2_credential_v3")
+	}
+
+	log.Printf("[DEBUG] Retrieved openstack_identity_ec2_credential_v3 %s: %#v", d.Id(), ec2Credential)
+
+	d.Set("secret", ec2Credential.Secret)
+	d.Set("user_id", ec2Credential.UserID)
+	d.Set("project_id", ec2Credential.TenantID)
+	d.Set("access", ec2Credential.Access)
+	d.Set("trust_id", ec2Credential.TrustID)
+
+	return nil
+}
+
+func resourceIdentityEc2CredentialV3Delete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	user, _, err := GetTokenInfo(identityClient)
+	if err != nil {
+		return fmt.Errorf("Error getting token info: %s", err)
+	}
+
+	var userID string
+	// Set userid to defined tf value, use one from token otherwise
+	if definedUser, ok := d.GetOk("user_id"); ok {
+		userID = definedUser.(string)
+	} else {
+		userID = user
+	}
+
+	err = ec2credentials.Delete(identityClient, userID, d.Id()).ExtractErr()
+	if err != nil {
+		err = CheckDeleted(d, err, "Error deleting openstack_identity_ec2_credential_v3")
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func GetTokenInfo(sc *gophercloud.ServiceClient) (string, string, error) {
+	r := sc.ProviderClient.GetAuthResult()
+	switch r := r.(type) {
+	case tokens2.CreateResult:
+		user, project, err := GetTokenInfoV2(r)
+		if err != nil {
+			return "", "", err
+		}
+		return user, project, nil
+	case tokens3.CreateResult:
+		user, project, err := GetTokenInfoV3(r)
+		if err != nil {
+			return "", "", err
+		}
+		return user, project, nil
+	default:
+		return "", "", fmt.Errorf("got unexpected AuthResult type %t", r)
+	}
+}
+
+func GetTokenInfoV3(t tokens3.CreateResult) (string, string, error) {
+	user, err := t.ExtractUser()
+	if err != nil {
+		return "", "", err
+	}
+	project, err := t.ExtractProject()
+	if err != nil {
+		return "", "", err
+	}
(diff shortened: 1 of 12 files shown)


pbasov:
Pull request to close #1016

pbasov:
ec2credentials v3 extension seems to be missing in vendor:
https://github.com/terraform-providers/terraform-provider-openstack/tree/master/vendor/github.com/gophercloud/gophercloud/openstack/identity/v3/extensions

Definitely there in gophercloud [v0.12](https://github.com/gophercloud/gophercloud/tree/v0.12.0/openstack/identity/v3/extensions)

kayrus:
> What's the best way forward to fix this?

you have to run `go mod vendor`, then add new vendor dir to git

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+			"access": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"secret": {
+				Type:      schema.TypeString,
+				Optional:  true,
```
kayrus, line 32: user cannot set the secret, therefore please remove `Optional: true`

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+func resourceIdentityEc2CredentialV3() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceIdentityEc2CredentialV3Create,
+		Read:   resourceIdentityEc2CredentialV3Read,
+		Delete: resourceIdentityEc2CredentialV3Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
```
kayrus, line 22: please specify the region, like it is done in other resources.

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+		return CheckDeleted(d, err, "Error retrieving openstack_identity_ec2_credential_v3")
+	}
+
+	log.Printf("[DEBUG] Retrieved openstack_identity_ec2_credential_v3 %s: %#v", d.Id(), ec2Credential)
+
+	d.Set("secret", ec2Credential.Secret)
+	d.Set("user_id", ec2Credential.UserID)
+	d.Set("tenant_id", ec2Credential.TenantID)
+	d.Set("access", ec2Credential.Access)
+	d.Set("trust_id", ec2Credential.TrustID)
```
kayrus, line 136: trust ID is not specified in the schema definition. please add it, and add it in the create function as well.

website/docs/r/identity_ec2_credential_v3.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+### EC2 credential in pre-define project scope
+```hcl
+resource "openstack_identity_ec2_credential_v3" "ec2_key1" {
+    project_id = "f7ac731cc11f40efbc03a9f9e1d1d21f"
+}
+```
+
+## Arguments Reference
+* `project_id` - The ID of the project the EC2 credential is created
```
kayrus, line 34: add a note that this field is optional

website/docs/r/identity_ec2_credential_v3.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+    be scoped to.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `access` - contains an EC2 credential access UUID
+* `secret` - contains an EC2 credential secret UUID
+* `user_id` - contains a User ID of the EC2 credential owner
+* `project_id` - contains an EC2 credential project scope
```
kayrus, line 45: add `trust_id` field here as well

website/docs/r/identity_ec2_credential_v3.html.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+## Example Usage
+
+### EC2 credential in current project scope
+
+```hcl
+resource "openstack_identity_ec2_credential_v3" "ec2_key1" {}
+```
+
+### EC2 credential in pre-define project scope
```
kayrus, line 26:
```suggestion
### EC2 credential in pre-defined project scope
```

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func resourceIdentityEc2CredentialV3Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	token := tokens.Get(identityClient, config.OsClient.TokenID)
```
kayrus, line 61: there is no need to request the token info from keystone every time, this info is already stored in provider client. you can extract it from the provider client, see the doc: https://pkg.go.dev/github.com/gophercloud/gophercloud?tab=doc#AuthResult
you may create an extra helper function, therefore we can optimize existing terraform resources logic, and remove redundant calls

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+		Read:   resourceIdentityEc2CredentialV3Read,
+		Delete: resourceIdentityEc2CredentialV3Delete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"access": {
+				Type:     schema.TypeString,
+				Optional: true,
```
kayrus, line 25: user cannot set the access, therefore please remove `Optional: true`

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func resourceIdentityEc2CredentialV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	token := tokens.Get(identityClient, config.OsClient.TokenID)
```
kayrus, line 115: redundant API call to keystone

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func resourceIdentityEc2CredentialV3Delete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	token := tokens.Get(identityClient, config.OsClient.TokenID)
```
kayrus, line 148: redundant API call to keystone

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+	if token.Err != nil {
+		return token.Err
+	}
+
+	user, err := token.ExtractUser()
+	if err != nil {
+		return err
+	}
+
+	var tenantId string
```
kayrus, line 71:
```suggestion
	var tenantID string
```

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,166 @@
...
+			},
+
+			"project_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"user_id": {
```
kayrus, line 45: cloud admins may create ec2 credentials for any user in the openstack cloud, mark this optional and add a note in readme that this property is available only for admins.

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,191 @@
...
+	token, ok := identityClient.ProviderClient.GetAuthResult().(tokens.CreateResult)
+	if !ok {
+		return fmt.Errorf("Unexpected auth error")
+	}
+
+	if token.Err != nil {
+		return token.Err
+	}
+
+	user, err := token.ExtractUser()
```
kayrus, line 82: `user_id` attribute here is not recognized here, when an admin wants to create ec2 credentials for a user.

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,191 @@
...
+
+func resourceIdentityEc2CredentialV3Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	token, ok := identityClient.ProviderClient.GetAuthResult().(tokens.CreateResult)
```
kayrus, line 73: keystone v2 is not dead, even when it is deprecated. I suggested to create a helper func to assert auth result from both tokens2 or tokens3.

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func resourceIdentityEc2CredentialV3Read(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	token, ok := identityClient.ProviderClient.GetAuthResult().(tokens.CreateResult)
```
kayrus, line 132: If admin creates a resource for another user, the resource cannot be retrieved here, because this logic uses admin's token and admin's user ID. See an example on how this case is solved in lbaas l7 rules: https://github.com/terraform-providers/terraform-provider-openstack/blob/152286f9ae3d016cb38d53ae2c0d1920d967d368/openstack/resource_openstack_lb_l7rule_v2.go#L202

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func resourceIdentityEc2CredentialV3Delete(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	identityClient, err := config.IdentityV3Client(GetRegion(d, config))
+	if err != nil {
+		return fmt.Errorf("Error creating OpenStack identity client: %s", err)
+	}
+
+	token, ok := identityClient.ProviderClient.GetAuthResult().(tokens.CreateResult)
```
kayrus, line 169: add an ability to delete credentials for any `user_id`. Example: https://github.com/terraform-providers/terraform-provider-openstack/blob/152286f9ae3d016cb38d53ae2c0d1920d967d368/openstack/resource_openstack_lb_l7rule_v2.go#L324

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,191 @@
...
+
+	err = ec2credentials.Delete(identityClient, user.ID, d.Id()).ExtractErr()
+	if err != nil {
+		err = CheckDeleted(d, err, "Error deleting openstack_identity_ec2_credential_v3")
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
kayrus, line 191: For admin purpose please specify an extra Import function, which will handle `access[/user_id]` (user_id should be optional, make a note for this in docs), like here: https://github.com/terraform-providers/terraform-provider-openstack/blob/152286f9ae3d016cb38d53ae2c0d1920d967d368/openstack/resource_openstack_lb_l7rule_v2.go#L372

website/docs/r/identity_ec2_credential_v3.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+
+* `access` - contains an EC2 credential access UUID
+* `secret` - contains an EC2 credential secret UUID
+* `user_id` - contains a User ID of the EC2 credential owner
+* `project_id` - contains an EC2 credential project scope
+* `trust_id` - contains an EC2 credential trust ID scope
+
+## Import
+
+EC2 Credentials can be imported using the `access`, e.g.
```
kayrus, line 50:
```suggestion
EC2 Credentials can be imported using the `access[/user_id]`, e.g.
```

openstack/resource_openstack_identity_ec2_credential_v3.go
pbasov: I've implemented it in the new commit, but honestly I'm not sure it's worth making so much mess over v2 token that might be stored in the providerclient at that point. 

Please check it out, but it's not pretty. I couldn't find a single example use of GetAuthResult() in the whole provider, so that's a first, I guess?

website/docs/r/identity_ec2_credential_v3.html.markdown
pbasov: Let me think about this some more, what's the usecase here? Just sticking to the "optional" tag in the doc?

kayrus: usecase is the ability to import the resource without recreation. For instance, when end user (e.g. admin, who creates EC2 credentials for unprivileged user) lost tfstate file and wants to restore it.

openstack/resource_openstack_identity_ec2_credential_v3.go
kayrus: > I couldn't find a single example use of GetAuthResult() in the whole provider, so that's a first, I guess?

you're correct. I asked to introduce this func, because there are openstack environments, which have a limit for keystone requests. I already significantly reduced the amount of redundant requests in my previous PRs. We can use 
 a new `GetTokenInfo` func in other data sources/resources like:
* `openstack_identity_auth_scope_v3`
* `openstack_identity_project_v3`
* `openstack_identity_application_credential_v3`

openstack/resource_openstack_identity_ec2_credential_v3.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+		return user, project, nil
+	case tokens3.CreateResult:
+		user, project, err := GetTokenInfoV3(r)
+		if err != nil {
+			return "", "", err
+		}
+		return user, project, nil
+	default:
+		panic(fmt.Sprintf("got unexpected AuthResult type %t", r))
```
kayrus, line 205: I'd not recommend to use `panic`, this can lead to terraform status loose.

website/docs/r/identity_ec2_credential_v3.html.markdown
pbasov: It should be already possible to import a resource like that by defining `user_id` in the resource definition though.

kayrus: yes, you're right. sorry for a confusion.

pbasov:
recheck

pbasov:
Hmm, I vendored the ec2cred extension (both sources and modules.txt change are included in the commit) and yet the [test](https://logs.openlabtesting.org/logs/33/1033/de14da0f500e4bc272beb4bddc2011ef1b93bac5/check/terraform-provider-openstack-unittest/5ec39df/job-output.txt.gz) seems to fail anyway. 

Am I doing something wrong? `make test` passes on my machine.

pbasov:
Looks like a false negative with blockstorage checks failures, do we recheck?

kayrus:
recheck

pbasov:
recheck

pbasov:
Shall we stop burning the trees in the name of green checkmarks?

kayrus:
recheck

pbasov:
It hasn't passed once from the end of July
https://status.openlabtesting.org/builds?job_name=terraform-provider-openstack-acceptance-test

pbasov:
recheck

wangxiyuan:
recheck

pbasov:
recheck

ozerovandrei:
I'm going to merge this since I don't know when CI will be fixed and changes in this PR a unrelated to those failures.

PR #1034, opened 2020-07-31
Add "allowed_cidrs" option for Octavia listeners

### LINKED ISSUE #1029: Allowed CIDR is not exposed for loadbalancer listeners
### Affected Resource(s)
- openstack_lb_listener_v2

### Expected Behavior

I should be able to set the equivalent of `--allowed-cidr` for `openstack loadbalancer listener create`.

### Actual Behavior

There is no exposed option that allow me to set the allowed CIDR for a listener.

### References

- https://docs.openstack.org/python-octaviaclient/latest/cli/index.html#loadbalancer-listener-create

dave2:
This adds support for Octavia's "allowed CIDRs" option on load balancer listeners. Gophercloud already supported the API, so this is just wiring up support in the provider side. It accepts a list of CIDR blocks as strings.

Fixes #1029

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -117,6 +117,14 @@ func chooseLBV2ListenerCreateOpts(d *schema.ResourceData, config *Config) (neutr
 
 		opts.InsertHeaders = headers
 
+		if raw, ok := d.GetOk("allowed_cidrs"); ok {
+			allowedCidrs := make([]string, len(raw.([]interface{})))
+			for i, v := range raw.([]interface{}) {
+				allowedCidrs[i] = v.(string)
+			}
+			opts.AllowedCIDRs = allowedCidrs
+		}
+
 		createOpts = opts
 
 		return createOpts, nil
@@ -236,6 +244,17 @@ func chooseLBV2ListenerUpdateOpts(d *schema.ResourceData, config *Config) (neutr
 			}
 
 			opts.InsertHeaders = &headers
+		}
+
+		if d.HasChange("allowed_cidrs") {
+			hasChange = true
+			var allowedCidrs []string
+			if raw, ok := d.GetOk("allowed_cidrs"); ok {
+				for _, v := range raw.([]interface{}) {
+					allowedCidrs = append(allowedCidrs, v.(string))
+				}
+			}
+			opts.AllowedCIDRs = &allowedCidrs
 		}
 
 		if hasChange {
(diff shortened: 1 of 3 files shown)


openstack/lb_v2_shared.go
```diff
@@ -117,6 +117,14 @@ func chooseLBV2ListenerCreateOpts(d *schema.ResourceData, config *Config) (neutr
 
 		opts.InsertHeaders = headers
 
+		if raw, ok := d.GetOk("allowed_cidrs"); ok {
+			var allowedCidrs []string
+			for _, v := range raw.([]interface{}) {
+				allowedCidrs = append(allowedCidrs, v.(string))
```
ozerovandrei, line 123: We don't really need to use append here since we already know the final slice size.
It's better to use the following syntax:
```go
allowedCidrs := make([]string, len(raw))
for i, v := range ...
    allowedCidrs[i] = v.(string)
...
```

openstack/lb_v2_shared.go
```diff
@@ -238,6 +246,17 @@ func chooseLBV2ListenerUpdateOpts(d *schema.ResourceData, config *Config) (neutr
 			opts.InsertHeaders = &headers
 		}
 
+		if d.HasChange("allowed_cidrs") {
+			hasChange = true
+			if raw, ok := d.GetOk("allowed_cidrs"); ok {
+				var allowedCidrs []string
+				for _, v := range raw.([]interface{}) {
+					allowedCidrs = append(allowedCidrs, v.(string))
```
ozerovandrei, line 254: Same issue with append.

openstack/lb_v2_shared.go
ozerovandrei, line 123: We don't really need to use append here since we already know the final slice size.
It's better to use the following syntax:
```go
allowedCidrs := make([]string, len(raw))
for i, v := range raw.([]interface{}) {
    allowedCidrs[i] = v.(string)
...
```

dave2: I've changed this as suggested

openstack/lb_v2_shared.go
dave2: In this case, we actually need to pass an empty array to AllowedCIDRs when we have no value but used to. This was preventing updates that removed the attribute entirely. I've reworked it so that happens correctly, but it still uses `append()`.

dave2:
I'll squash the commits once all reviews pass

dave2:
Should have docs as well now, making clear not providing `allowed_cidrs` is the same as allow-all.

dave2:
> Can you please also provide a website documentation updates https://github.com/terraform-providers/terraform-provider-openstack/blob/master/website/docs/r/lb_listener_v2.html.markdown

Missed this, but should be resolved now.

dave2:
It's good to go. I haven't added money-spending tests, but it has been tested a fair bit on my employer's public cloud based on OpenStack.

PR #1039, opened 2020-08-06
Fix merge labels

brtknr:
v1.30.0 release breaks backward compatibility by sending merge_labels.
Only add this to createOpts when true.

### DIFF
--- a/openstack/resource_openstack_containerinfra_cluster_v1.go
+++ b/openstack/resource_openstack_containerinfra_cluster_v1.go
@@ -268,7 +268,6 @@ func resourceContainerInfraClusterV1Create(d *schema.ResourceData, meta interfac
 
 	// Get boolean parameters that will be passed by reference.
 	floatingIPEnabled := d.Get("floating_ip_enabled").(bool)
-	mergeLabels := d.Get("merge_labels").(bool)
 
 	createOpts := clusters.CreateOpts{
 		ClusterTemplateID: d.Get("cluster_template_id").(string),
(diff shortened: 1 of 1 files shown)


openstack/resource_openstack_containerinfra_cluster_v1.go
```diff
@@ -305,6 +303,11 @@ func resourceContainerInfraClusterV1Create(d *schema.ResourceData, meta interfac
 		createOpts.NodeCount = &nodeCount
 	}
 
+	mergeLabels := d.Get("merge_labels").(bool)
+	if mergeLabels == true {
```
ozerovandrei, line 307: It's better to use a simpler form:
```go
if mergeLabels {
    createOpts.MergeLabels = &mergeLabels
}
```

brtknr: Sounds good, the reason I did not initially implement it like this was because this seem to imply that null values are emitted in request: https://github.com/terraform-providers/terraform-provider-openstack/blob/master/vendor/github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clusters/requests.go#L29

Is that not the case?

kayrus: From my perspective the fix has to be made in gophercloud, not in terraform provider. Since `merge_labels` is an optional parameter, it should be a regular boolean, not a pointer. Additionally unit tests for [gophercloud/openstack/containerinfra/v1/clusters](https://github.com/gophercloud/gophercloud/blob/51f8fa152459ae60d3b348023ad79f850db3a931/openstack/containerinfra/v1/clusters/testing/fixtures.go#L99) are not full and don't include `th.TestJSONRequest` calls.

brtknr: The pointer pattern comes from cluster template and how bools are implemented there.

ozerovandrei:
recheck

ozerovandrei (approved):
I guess it should be implemented as a pointer since old Magnum API won't vaildate a request with the additional merge_labels field.

LGTM.

I'm going to merge this since test failures are unrelated to changes in this PR.

PR #1070, opened 2020-09-21
compute/instance: add different_cell scheduler hint

cardoe:
While the code base supported the target_cell scheduler hint to get
instances to build in a specified cell, there was no support to skip
cells if you were attempting to skip a specific cell.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/terraform-provider-openstack/terraform-provider-openstack
 go 1.14
 
 require (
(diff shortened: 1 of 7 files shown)


ozerovandrei (changes requested):
Looks good but needs to be rebased.

cardoe:
Thanks for the review. Rebased and should pass tests.

ozerovandrei (approved):

cardoe:
Thank you.

PR #1073, opened 2020-09-25
compute_instance_v2: handle special case of 2 default networks

### LINKED ISSUE #1071: network ordering appears to want to rebuild compute instance
Creating a number of nodes using `count` and depending on the ordering of the network it will decide that it needs to rebuild a node.

### Terraform Version
$ terraform -v
Terraform v0.12.29
+ provider.openstack v1.30.0.1

### Affected Resource(s)

```hcl
resource "openstack_compute_instance_v2" "node" {
  count     = var.worker_count
  name      = format("node%02d.%s.mydomain.com", count.index + 1, var.region)
  image_id  = var.image_id
  flavor_id = var.flavor_id

  network {
    uuid = "00000000-0000-0000-0000-000000000000"
    name = "public"
  }

  network {
    uuid = "11111111-1111-1111-1111-111111111111"
    name = "private"
  }

  metadata = {
    role = "node"
    env  = var.region
  }

}
```

### Debug Output
I'll have to get a sanitized output as the debug contains a lot of creds in it. But a plan output of one of the baddies

```
  # openstack_compute_instance_v2.node[9] must be replaced
-/+ resource "openstack_compute_instance_v2" "node" {
      ~ access_ip_v4        = "10.190.37.33" -> (known after apply)
      + access_ip_v6        = (known after apply)
      ~ all_metadata        = {
          - "env"  = "myregion"
          - "role" = "node"
        } -> (known after apply)
      + all_tags            = (known after apply)
      + availability_zone   = (known after apply)
        flavor_id           = "7"
      ~ flavor_name         = "15GB Standard Instance" -> (known after apply)
      + force_delete        = false
      ~ id                  = "4854412f-ed08-4cb4-8fc4-e8ccf92b5f7c" -> (known after apply)
        image_id            = "d511a1dc-76c1-4091-9b56-be2ea3aa01cb"
      ~ image_name          = "Ubuntu 18.04 LTS (Bionic Beaver) (PVHVM)" -> (known after apply)
        metadata            = {
            "env"  = "myregion"
            "role" = "node"
        }
        name                = "node10.myregion.mydomain.com"
        power_state         = "active"
      + region              = (known after apply)
      ~ security_groups     = [] -> (known after apply)
      + stop_before_destroy = false

      ~ network {
            access_network = false
          ~ fixed_ip_v4    = "10.190.37.33" -> (known after apply)
          + fixed_ip_v6    = (known after apply)
          + floating_ip    = (known after apply)
          + mac            = (known after apply)
          ~ name           = "private" -> "public" # forces replacement
          + port           = (known after apply)
          + uuid           = (known after apply)
        }
      ~ network {
            access_network = false
          ~ fixed_ip_v4    = "10.69.213.181" -> (known after apply)
          + fixed_ip_v6    = (known after apply)
          + floating_ip    = (known after apply)
          + mac            = (known after apply)
          ~ name           = "public" -> "private" # forces replacement
          + port           = (known after apply)
          + uuid           = (known after apply)
        }

      - timeouts {}
    }
```

### Expected Behavior
It doesn't want to recreate these instances.

### Actual Behavior
It wants to recreate the instances.

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Rackspace OpenStack

cardoe:
When you query for the details on a server in OpenStack, you get a list
of addresses assigned to that server which is keyed by pool. This
unfortunately causes problems becauses addresses are actually a JSON
object keyed by their network name. JSON objects are unordered by spec
and Go uses map[string]interface{} for JSON objects which is also
unordered. However this provider keeps the networks in a TypeList which
is order dependent. If you import a server then the networks will come
in a random order and terraform will insist it must rebuild the server.
Unfortunately there's no way to ensure that the ordering matches how it
came back from OpenStack due to this. This is only an issue when there
are two default networks such as for a provider with "public" and
"private" as default networks. Without a way to generically fix the
ordering of networks for imported servers, this fix just special cases
that situation to workaround the issue. fixes #1071

### DIFF
--- a/openstack/compute_instance_v2.go
+++ b/openstack/compute_instance_v2.go
@@ -341,7 +341,35 @@ func getInstanceNetworkInfoNeutron(
 func getInstanceAddresses(addresses map[string]interface{}) []InstanceAddresses {
 	var allInstanceAddresses []InstanceAddresses
 
-	for networkName, v := range addresses {
+	// Addresses includes a list of all IP addresses assigned to the server,
+	// keyed by pool. This unfortunately causes problems because addresses are a
+	// JSON object which means that they are unordered because gophercloud
+	// uses a map[string]interface{} to hold them and maps are unordered.
+	// However OpenStack uses OrderedDict to return this data and this
+	// provider uses a List type for the networks which is an ordered type,
+	// this means that importing resources into terraform will result in
+	// networks import out of order which causes terraform to say it has to
+	// rebuild the instance. Unfortunately there is no way to ensure the
+	// ordering is correct, lastly this issue occurs only when there's
+	// "public" and "private" as the only networks so account for that
+	// case specially. Create a list of the network names and if it's
+	// the special case override it.
+	networkNames := make([]string, len(addresses))
+	i := 0
+	for k := range addresses {
+		networkNames[i] = k
+		i++
+	}
+
+	if len(networkNames) == 2 {
+		if networkNames[0] == "private" && networkNames[1] == "public" {
+			networkNames[0] = "public"
+			networkNames[1] = "private"
+		}
+	}
+
+	for _, networkName := range networkNames {
+		v, _ := addresses[networkName]
 		instanceAddresses := InstanceAddresses{
 			NetworkName: networkName,
 		}


ozerovandrei (approved):
LGTM

Thank you for debugging this and for the great explanation @cardoe !

cardoe:
Thank you for merging this.

gvcgael:
@cardoe I'm trying to make sense of this pull request.

```

	if len(networkNames) == 2 {
		if networkNames[0] == "private" && networkNames[1] == "public" {
			networkNames[0] = "public"
			networkNames[1] = "private"
		}
	}
```

What happens for instances which have two networks but are not named `private` or `public` ? Why does this bug only occurs for these names, not for `some_main_network` and `some_secondary_network` ?

Thanks in advance

cardoe:
This bug will occur for anyone importing any instance with more than 1 network. As I tried to explain in the commit, this provider made a mistake by storing this data in TypeList (of terraform) which is an ordered list type. The data is returned by OpenStack in a JSON object which is an unordered type. The provider walks each key in the object and converts it to a list, the result is this provider is taking an unordered type and making it ordered internally. So every time that it runs the order can vary resulting in the provider wanting to rebuild the instance at random times. The correct fix is that this provider should be using TypeSet, which is an unordered list. However that's stored completely differently in the terraform state so there's absolutely no way to migrate everyone's state nicely without invalidating everyone's state which seemed like a poor action to take.

At this time, I'm aware of this affecting instances built and imported on Rackspace's OpenStack as they have a "public" and "private" network on each instance. So without a clean way of fixing the provider without hurting everyone, this adds a special exception for that case to stop causing this provider to behave badly on Rackspace's OpenStack.

I'm game for fixing it generically but I couldn't figure out how to convert TypeList to TypeSet without major issues for people.

gvcgael:
Okay thank you.

As a workaround, adding the networks in the ignored properties for the lifecycle changes seems to be a good fix as a  first-approach, what do you think ?

cardoe:
I don't think it is. Because people will actually want to change the networks on the instance and have those network changes take effect.

kayrus:
I don't think this fix is a proper one. See similar fixes in k8s openstack cloud controller manager:
* https://github.com/kubernetes/cloud-provider-openstack/pull/884
* https://github.com/kubernetes/cloud-provider-openstack/pull/1041

Issue #1063, opened 2020-09-14
Adding Golanci-lint Github Action and refactoring code for linters


ozerovandrei:
We should use http://golangci-lint.run to check the provider code against lots of useful Go linters.
This project is used by lots of other providers as well as other Go projects.

We can add the Github Action workflow: https://golangci-lint.run/usage/install#github-actions

Lots of code should be refactored to fix most linters issues.

dkder3k:
Due to deprecation of maligned and interfacer linters should we disable them?
Latest versions of golangci-lint show corresponding warnings.
```
WARN [runner] The linter 'maligned' is deprecated due to: The repository of the linter has been archived by the owner. Use govet 'fieldalignment' instead. 
WARN [runner] The linter 'interfacer' is deprecated due to: The repository of the linter has been archived by the owner.
```
There is [a suggestion](https://github.com/golangci/golangci-lint/issues/541#issuecomment-790929096) to replace maligned with fieldalignment but for interfacer there is no any replacement afaik

ozerovandrei:
> Due to deprecation of maligned and interfacer linters should we disable them?
I don't see why we should use deprecated linters.
We can also run `golangci-lint linters` in the new version to see which linters do we need or not.
I guess that we need most of them to simplify PR reviews.

Issue #1058, opened 2020-09-09
Core: terraform mutexkv is deprecated


kayrus:
https://github.com/hashicorp/terraform-plugin-sdk/pull/438

> Deprecated: This will be removed in v2 without replacement. If you need its functionality, you can copy it or reference the v1 package

PR #1078, opened 2020-10-08
Moved global mutexKV to Config structure

dkder3k:
So this global mutex kv struct available everywhere in provider without
being a global variable.

For #1063

### DIFF
--- a/openstack/provider.go
+++ b/openstack/provider.go
@@ -9,13 +9,11 @@ import (
 	"github.com/gophercloud/utils/terraform/auth"
 )
 
-// This is a global MutexKV for use within this plugin.
-var osMutexKV = mutexkv.NewMutexKV()
-
 // Use openstackbase.Config as the base/foundation of this provider's
-// Config struct.
+// Config struct and contains a global MutexKV.
 type Config struct {
 	auth.Config
+	*mutexkv.MutexKV
 }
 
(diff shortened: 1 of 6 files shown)


openstack/provider.go
```diff
@@ -9,13 +9,11 @@ import (
 	"github.com/gophercloud/utils/terraform/auth"
 )
 
-// This is a global MutexKV for use within this plugin.
-var osMutexKV = mutexkv.NewMutexKV()
-
 // Use openstackbase.Config as the base/foundation of this provider's
-// Config struct.
+// Config struct and contains a global MutexKV.
```
ozerovandrei, line 13: Can we rephrase it like: "Config contains auth parameters and also provides an instance of MutexKV." ?

ozerovandrei (changes requested):

kayrus:
According to #1058 let's just move the `github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv` package under the https://github.com/gophercloud/utils/blob/master/terraform/auth/config.go

dkder3k:
@kayrus this will be done during next pr (after this one will be merged). The purpose of this pr is just to get rid of global variable for a mutex

ozerovandrei (approved):

PR #1080, opened 2020-10-11
Copy MutexKV to internal package

dkder3k:
Because MutexKV will be removed from terraform-plugin-sdk package copy
it to local internal/mutexkv package.

For #1058

### DIFF
--- a/openstack/internal/mutexkv/mutexkv.go
+++ b/openstack/internal/mutexkv/mutexkv.go
@@ -1,0 +1,51 @@
+package mutexkv
+
+import (
+	"log"
+	"sync"
+)
+
+// MutexKV is a simple key/value store for arbitrary mutexes. It can be used to
+// serialize changes across arbitrary collaborators that share knowledge of the
+// keys they must serialize on.
+//
+// The initial use case is to let aws_security_group_rule resources serialize
+// their access to individual security groups based on SG ID.
+type MutexKV struct {
+	lock  sync.Mutex
+	store map[string]*sync.Mutex
+}
+
+// Locks the mutex for the given key. Caller is responsible for calling Unlock
+// for the same key.
+func (m *MutexKV) Lock(key string) {
+	log.Printf("[DEBUG] Locking %q", key)
+	m.get(key).Lock()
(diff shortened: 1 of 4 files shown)


ozerovandrei (approved):

kayrus:
@dkder3k as I mentioned in https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1078#issuecomment-706187999, a better place for a mutexkv is in https://github.com/gophercloud/utils/tree/master/terraform

dkder3k:
@kayrus  afaik, gophercloud by itself doesn't use this MutexKV structure and I'm not sure that common utils repo is a better place. Afaik this structure is mostly used by tf providers (like this one, [aws provider](https://github.com/terraform-providers/terraform-provider-aws/tree/master/aws/internal/mutexkv), [selectel provider](https://github.com/selectel/terraform-provider-selectel/tree/master/selectel/internal), etc.) so my humble opinion is that it is better keep it here and not make utils repo a all-in-one place for not related things. But if there is any point (maybe there is a need for structure like this in gophercloud or any other related projects) lets think about it.

kayrus:
@dkder3k gophercloud/utils/terraform is used by this provider and other openstack extension providers like https://github.com/selectel/terraform-provider-selectel/tree/master/selectel/internal or https://github.com/sapcc/terraform-provider-ccloud. So I'd prefer to move mutexkv into the `gophercloud/utils/terraform`.
@ozerovandrei can you share your feedback on this?

ozerovandrei:
@kayrus I guess it's possible to extend `gophercloud/utils` to include a separate `mutexkv` package inside it.
In that case it should also include the [hashcode](https://github.com/hashicorp/terraform-plugin-sdk/tree/v1.16.0/helper/hashcode) helper that was also removed from the Terraform SDK V2.

Maybe `gophercloud/utils/terraform/auth` it's not a good place for those features. What about `gophercloud/utils/terraform/mutexkv` and `gophercloud/utils/terraform/hashcode` ?

kayrus:
> What about `gophercloud/utils/terraform/mutexkv` and `gophercloud/utils/terraform/hashcode` ?

+1

Issue #1082, opened 2020-10-12
Port binding.host_id set to null? 


GregoryVds:
Hello,

If I run `terraform plan` immediately after the deployment of my infrastructure, without any change, I see that the port `binding.host_id` would be set to `null`. I would expect this sub-subsequent `plan` to be give me a no-op.

I only want to specify the `vnic_type` for the port, but I don't want to specify the `host_id` and I left this empty in my Terraform resource definition. 

Why is it resetting the `host_id`? 
(If I apply this plan, the `binding.host_id` is set to null, my port enters `DOWN` status and I loose connectivity).

Thanks in advance,
Gregory

### Terraform Version

Terraform v0.13.4
Terraform Orovider Openstack v1.32.0

### Affected Resource(s)
- openstack_networking_port_v2

``` 
resource "openstack_networking_port_v2" "node_main" {
  name           = "node-main"
  admin_state_up = "true"
  port_security_enabled = "false"
  network_id = var.network.main_id 

  binding {
    vnic_type = var.perf ? "direct" : "normal"
  }
}

resource "openstack_networking_port_v2" "node_int" {
  name           = "node-int"
  admin_state_up = "true"
  port_security_enabled = "false"
  network_id = var.network.int_id

  binding {
    vnic_type = var.perf ? "direct" : "normal"
  }
}

resource "openstack_networking_port_v2" "node_slave" {
  name           = "node-slave"
  admin_state_up = "true"
  port_security_enabled = "false"
  network_id = var.network.slave_id

  binding {
    vnic_type = var.perf ? "direct" : "normal"
  }
}
``` 

### Debug Output

https://gist.github.com/GregoryVds/5e02af04cea43598235e5670dc504eb0

### Expected Behavior

I expect this `plan` to be a no-op and the ports to be untouched. My infrastructure has just been freshly deployed and I didn't make any config change.

### Actual Behavior

It seems the port `binding.host_id` is set to `NULL`. 
Port then enters status `DOWN` and connectivity is lost.

GregoryVds:
I did some more testing, and the problem does arise if I remember the whole `binding` section in the Terraform resource definition.

ozerovandrei:
Fixed in #1084

PR #1084, opened 2020-10-14
Networking V2: fix 'binding.host_id' reset (#1082)

GregoryVds:
Hi,

This is linked to my issue #1082.

I don't know if this is the right fix but this seems to fix my issue (and maybe it would prevent an explicit setting of the host_id in the Terraform file?). It seems to be set as "Computed " as well in openstack/data_source_openstack_networking_port_v2.go but I am really not aware of the difference in purpose between these 2 files.

Thanks for the feedback,
Gregory

----

If the binding section is defined in the ressource definition (e.g.
just with "binding.vnic_type" set), but "binding.host_id" is left empty,
the port is created correctly and "binding.host_id" populated with the
id of the selected hypervisor. However, upon a subsequent plan/apply,
Terraform seems to reset "binding.host_id" to NULL. The port enters
status DOWN and connectivity is lost. Not sure this is the right fix,
but it seems the "host_id" should not be overriden in this case.

Signed-off-by: Gregory Vander Schueren <gregory.vanderschueren@tessares.net>

### DIFF
--- a/openstack/resource_openstack_networking_port_v2.go
+++ b/openstack/resource_openstack_networking_port_v2.go
@@ -224,6 +224,7 @@ func resourceNetworkingPortV2() *schema.Resource {
 						"host_id": {
 							Type:     schema.TypeString,
 							Optional: true,
+							Computed: true,
 						},
 						"profile": {
 							Type:             schema.TypeString,


ozerovandrei (approved):
LGTM

PR #1085, opened 2020-10-16
Use mutexkv from gophercloud/utils

dkder3k:
* Remove internal mutexkv package
* Update gophercloud/utils dependency

After [this pr](https://github.com/gophercloud/utils/pull/127) was merged it is possible to use this structure from gophercloud/utils dependency as it was discussed in #1080


ozerovandrei (approved):
Thank you!

PR #1086, opened 2020-10-17
add address_scope, security_group and subnetpool to rbac types

tobikris:
Openstack release ussuri added some types for RBAC policies. This allows them to be managed by the provider.


ozerovandrei (approved):
Thank you!

Issue #1038, opened 2020-08-05
Attach volume incorrect time report


jichenjc:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

master version 
### Affected Resource(s)
Please list the resources as a list, for example:
openstack_compute_volume_attach_v2

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Debug Output
Please provide a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. 

In addition, running Terraform with the `OS_DEBUG=1` environment variable set will print additional information specific to the OpenStack provider which will be helpful for troubleshooting.

Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
What should have happened?

we are attaching 2 Volumes to VM and the 
openstack_compute_volume_attach_v2.db2: Creating...
  device:      "" => "<computed>"
  instance_id: "" => "9ebed95a-81fa-4a9b-aa70-73829b85015e"
  region:      "" => "<computed>"
  volume_id:   "" => "cb8e2089-550d-4036-9f4e-595076d8d643"
openstack_compute_volume_attach_v2.db: Creating...
  device:      "" => "<computed>"
  instance_id: "" => "9ebed95a-81fa-4a9b-aa70-73829b85015e"
  region:      "" => "<computed>"
  volume_id:   "" => "0a0d4d79-e20a-43e6-a59f-b3d8d2616c22"
openstack_compute_volume_attach_v2.db2: Creation complete after 6s (ID: 9ebed95a-81fa-4a9b-aa70-73829b85015e/cb8e2089-550d-4036-9f4e-595076d8d643)
openstack_compute_volume_attach_v2.db: Still creating... (10s elapsed)
[...]

PR #1106, opened 2020-12-01
Check volume status for 'in-use' before return 'ATTACHED'

jichenjc:


jichenjc:
@fgimenezm this is the PR that for #1038 , I didn't test yet ,just push here so that you might have rough idea on what I am proposing

openstack/compute_volume_attach_v2.go
```diff
@@ -33,11 +35,20 @@ func computeVolumeAttachV2AttachFunc(computeClient *gophercloud.ServiceClient, i
...
+			return va, "ATTACHING", nil
+		}
+
 		return va, "ATTACHED", nil
 	}
 }
 
-func computeVolumeAttachV2DetachFunc(computeClient *gophercloud.ServiceClient, instanceID, attachmentID string) resource.StateRefreshFunc {
+func computeVolumeAttachV2DetachFunc(
+	computeClient *gophercloud.ServiceClient, instanceID, attachmentID string) resource.StateRefreshFunc {
```
ozerovandrei, line 51: No need to move it to different line.

openstack/compute_volume_attach_v2.go
```diff
@@ -23,7 +24,8 @@ func computeVolumeAttachV2ParseID(id string) (string, string, error) {
 	return instanceID, attachmentID, nil
 }
 
-func computeVolumeAttachV2AttachFunc(computeClient *gophercloud.ServiceClient, instanceID, attachmentID string) resource.StateRefreshFunc {
+func computeVolumeAttachV2AttachFunc(
```
ozerovandrei, line 27: Don't need to move it to a different line.

jichenjc: updated, thanks

openstack/compute_volume_attach_v2.go
```diff
@@ -33,6 +34,14 @@ func computeVolumeAttachV2AttachFunc(computeClient *gophercloud.ServiceClient, i
 			return va, "", err
 		}
 
+		v, err := volumes.Get(blockStorageClient, volumeID).Extract()
+		if err != nil {
+			return va, "", err
+		}
+		if v.Status != "in-use" {
```
ozerovandrei, line 41: What if the API responds with 200 status code but volume status is "error"? I guess we need to handle that case and return some error from `computeVolumeAttachV2AttachFunc` instead of returning the "ATTACHING" status.

jichenjc: yes, updated a `error` check ,thanks for point out

ozerovandrei (approved):
LGTM

Let me know if it's ready to be merged or if you want to make any other changes.

jichenjc:
I think it's ready to be merged now .. thanks

PR #1110, opened 2020-12-03
Upgrade golangci-lint-action for new github actions env changes

bpetermannS11:
Github deprecated some env things:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

New golangci-lint looks like it fixes this

https://github.com/golangci/golangci-lint-action/issues/127#issuecomment-729139550


ozerovandrei (approved):
Thank you @bpetermannS11

PR #1117, opened 2020-12-16
Docs: remove mention of `floating_ip` option in `openstack_compute_instance_v2`.

### LINKED ISSUE #1115: Docs: `floating_ip` options in `openstack_compute_instance_v2`.
Hello,
I have found something a bit inconsistent/confusing in the documentation:

The incipit of https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/master/website/docs/r/compute_floatingip_associate_v2.html.markdown says "Associate a floating IP to an instance. This can be used instead of the `floating_ip` options in `openstack_compute_instance_v2`."

However on the page for the `compute_instance_v2`, the only time `floating_ip` is mentioned is with an example of how to use `floatingip_associate_v2`.

This leads me to believe that one of the following is true:

- There is no "`floating_ip` options in `openstack_compute_instance_v2`" and thus the only way to associate a floating IP to an instance is through `floatingip_associate`
or

- There is a `floating_ip` option in `openstack_compute_instance_v2` which is however undocumented

If someone familiar with the project could confirm either of the two cases or if I'm missing something I would be happy to update the documentation accordingly.

Thank you for this awesome provider!

aledegano:
Remove mention of `floating_ip` option in `openstack_compute_instance_v2`
from the documentation of `openstack_compute_floatingip_associate_v2`
resource as that option was removed with 3d426884761ca431da5ed361bf59f9c4a26e1d90.

Fix #1115.

Signed-off-by: Alessandro Degano <a.degano@gmail.com>

### DIFF
--- a/website/docs/r/compute_floatingip_associate_v2.html.markdown
+++ b/website/docs/r/compute_floatingip_associate_v2.html.markdown
@@ -8,8 +8,7 @@ description: |-
 
 # openstack\_compute\_floatingip_associate_v2
 
-Associate a floating IP to an instance. This can be used instead of the
-`floating_ip` options in `openstack_compute_instance_v2`.
+Associate a floating IP to an instance.
 
 ## Example Usage
 


ozerovandrei (approved):
LGTM

ozerovandrei:
recheck

ozerovandrei:
recheck

PR #1118, opened 2020-12-16
Properly read fixed_ip property for network attachment

MatMaul:
Currently if `fixed_ip` is specified idempotence is not respected because the property is not read on refresh.


ozerovandrei:
recheck

ozerovandrei (approved):
LGTM

Note: API parameters are explained here: [port-interfaces-servers-os-interface](https://docs.openstack.org/api-ref/compute/?expanded=create-interface-detail#port-interfaces-servers-os-interface)

Issue #1119, opened 2020-12-21
terraform-provider-openstack 1.34.0 forces recreation of openstack_compute_flavor_v2


nikParasyr:
### Terraform Version
```
Terraform v0.12.21
+ provider.openstack v1.34.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_compute_flavor_v2

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    openstack = "= 1.33.0"
  }
}

provider openstack {
}

resource "openstack_compute_flavor_v2" "m1-large" {
  name      = "m1.large"
  ram       = "16384"
  vcpus     = "8"
  disk      = "50"
  swap      = "0"
  is_public = "true"
  extra_specs = {
    "storage_type" = "ceph"
  }
}
```

### Expected Behavior
Changing versions should not force recreation of flavor resources.

### Actual Behavior
All flavor resources need to be recreated. Tried to "fix" it by deleting the resource and importing it but it still "complains" about the `flavor_id` to `id` change and triggers a recreate.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` using 1.33.0 version of the provider
2. Switch to 1.34.0 and init again
3. terraform plan => shows that the resources needs to be recreated with this change:
```
      - flavor_id    = "1202e6e0-75b0-4fd1-9d3e-9aaa11173dce" -> null # forces replacement
      ~ id           = "1202e6e0-75b0-4fd1-9d3e-9aaa11173dce" -> (known after apply)
```

### Important Factoids
Openstack Stein release. No changes to upstream nova.

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1107/files This could be related.

nikParasyr:
The problem persists even if you create the resource on 1.34.0 version and then run a plan again, it still shows that the resource needs to be recreated

ozerovandrei:
Could you please check if problem still exists on v1.34.1 version?

nikParasyr:
Tested it and the problem is resolved. Thanks for the fast reply :) 

Closed by #1120

PR #1120, opened 2020-12-21
Flavor V2: make "flavor_id" computed

ozerovandrei:
We set it on Read() call so it should be computed and it shouldn't
create plan changes in case of an empty value.

For #1119


ftarasenko (approved):
My fault, think we need more test cases for resource.

PR #1124, opened 2021-01-07
add: ignore image stores propertie

fitbeard:
In newer OpenStack releases (Ussuri) glance adds read-only image propertie _stores_.
Screenshot 2021-01-07 at 23 45 56
It need to be excluded like _direct_url_


ozerovandrei:
recheck

ozerovandrei (approved):
LGTM

PR #1125, opened 2021-01-08
Add image_id to openstack_images_image_v2 resource creation options

fitbeard:
OpenStack and GopherCloud have ability to set Glance image UUID while creating image.
Replicating this functionality in Terraform. Docs and tests also altered.
Regex taken from Glance API response.
This is useful with images dedicated to Ironic when creating bundles of AKI/AMI and QCOW2 and in QCOW2 properties info about _ramdisk_id_ and _kernel_id_ can be filled with AKI and AMI UUID's.

### DIFF
--- a/openstack/resource_openstack_images_image_v2.go
+++ b/openstack/resource_openstack_images_image_v2.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
@@ -61,6 +62,14 @@ func resourceImagesImageV2() *schema.Resource {
 			"file": {
 				Type:     schema.TypeString,
 				Computed: true,
+			},
+
+			"image_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^([0-9a-fA-F]){8}-([0-9a-fA-F]){4}-([0-9a-fA-F]){4}-([0-9a-fA-F]){4}-([0-9a-fA-F]){12}$`), "image_id must be valid UUID"),
 			},
 
 			"image_cache_path": {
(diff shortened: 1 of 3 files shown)


openstack/resource_openstack_images_image_v2_test.go
```diff
@@ -367,6 +383,19 @@ const testAccImagesImageV2Basic = `
       }
   }`
 
+const testAccImagesImageV2BasicWithID = `
+  resource "openstack_images_image_v2" "image_1" {
+	  name   = "Rancher TerraformAccTest"
+	  image_id = "c1efdf94-9a1a-4401-88b8-d616029d2551"
```
ozerovandrei, line 389: Indentation is bad for `name` and `image_id`.

openstack/resource_openstack_images_image_v2.go
```diff
@@ -63,6 +64,14 @@ func resourceImagesImageV2() *schema.Resource {
 				Computed: true,
 			},
 
+			"image_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^([0-9a-fA-F]){8}-([0-9a-fA-F]){4}-([0-9a-fA-F]){4}-([0-9a-fA-F]){4}-([0-9a-fA-F]){12}$`), "image_id must be valid UUID"),
```
ozerovandrei, line 72: Can we use a validation function from SDK here? https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.16.0/helper/validation/uuid.go#L10

ozerovandrei (changes requested):

openstack/resource_openstack_images_image_v2.go
fitbeard: Changed to SDK: `invalid value for image_id (image_id must be valid UUID)` now even default message from function is enough informative

openstack/resource_openstack_images_image_v2_test.go
fitbeard: My bad. Fixed.

ozerovandrei (approved):
LGTM

PR #1126, opened 2021-01-09
Hypervisor data source support

rudexi:
Support for Hypervisor data source.

Due to the nature of hypervisors (only LIST is available in the API), it's only possible to make data sources for hypervisors.
This data source aim is to provide information about cpu/memory/disk, so it's possible to auto-calculate quotas by listing every hypervisors in the system.

This backend is, as far as I know, only available for the admin user, so the test is restricted accordingly.

### DIFF
--- a/openstack/data_source_openstack_compute_hypervisor_v2.go
+++ b/openstack/data_source_openstack_compute_hypervisor_v2.go
@@ -1,0 +1,98 @@
+package openstack
+
+import (
+	"fmt"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/hypervisors"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceComputeHypervisorV2() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceComputeHypervisorV2Read,
+		Schema: map[string]*schema.Schema{
+			"hostname": {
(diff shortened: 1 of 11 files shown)


openstack/data_source_openstack_compute_hypervisor_v2.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+func dataSourceComputeHypervisorV2() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceComputeHypervisorV2Read,
+		Schema: map[string]*schema.Schema{
+			"hostname": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
ozerovandrei, line 17: Could you please separate attributes with a blank line?

ozerovandrei (changes requested):

openstack/data_source_openstack_compute_hypervisor_v2.go
rudexi: Done

ozerovandrei: I propose to add a blank line separator the same way as https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1121#discussion_r555937346

ozerovandrei (changes requested):

ozerovandrei (approved):
LGTM

PR #1128, opened 2021-01-13
Bump Gophercloud

pierreprinetti:
Remove a dependency on a deprecated package: utils now uses "imageservice"
instead of "compute" to resolve image IDs.


pierreprinetti:
This issue is actually less anodyne than it looks. Because of how Minimal version selection works, the backwards-incompatible change introduced in utils is breaking `go mod` in dependant packages.

This patch changes `terraform-provider-openstack` so that it does not depend any more on a removed `utils` function.

mandre (commented):
Thanks :+1:

mandre (approved):

Fedosin (approved):
LGTM

ozerovandrei (approved):
LGTM

PR #1135, opened 2021-01-21
doc: use local provider

pierrchen:
openstack provider has moved to terraform-provider-openstack namespace,
and we can't use that as the local name directly, instead we define a
local name in the provider configuration block. This is required for newer terraform
version (maybe > 0.13.0).

[1]https://www.terraform.io/docs/configuration/provider-requirements.html#local-names


ozerovandrei (approved):
LGTM

PR #1141, opened 2021-01-22
Only use old client micro-version for old policies

elafontaine:
This should allow custom policies to be applied by using the more recent version of nova API which allow different policies then anti-affinity and affinity.

See https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1134 for more details

### DIFF
--- a/openstack/compute_servergroup_v2.go
+++ b/openstack/compute_servergroup_v2.go
@@ -8,8 +8,6 @@ import (
 const (
 	antiAffinityPolicy = "anti-affinity"
 	affinityPolicy     = "affinity"
-	softAntiAffinityPolicy = "soft-anti-affinity"
-	softAffinityPolicy     = "soft-affinity"
 )
 
 // ServerGroupCreateOpts is a custom ServerGroup struct to include the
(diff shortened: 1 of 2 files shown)


elafontaine:
Hi, this is a PR just to share what I believe the code could look like which would preserve the old behavior and allow the new.

dhiaayachi:
Hello
Modified the fix to ensure behavior for affinity and anti-affinity policies is unchanged. Also added/modified unit tests accordingly.
Please review

elafontaine:
Thanks @dhiaayachi !
@ozerovandrei , could you review and let us know what you think ? (Or send someone to review and merge ?)

We're starting to be dependent on this to be able to promote terraform internally :) .

ozerovandrei (approved):
LGTM

Note that this won't work when Compute API adds some policies that are available from some new 2.xx microversion.

elafontaine:
Hi @ozerovandrei , it's important to note that after that micro-version, the confirmation is done on the server side (If I recall right), so the client side version won't really matter for custom policies until Openstack gets some new fields on the server groups API.

elafontaine:
Thanks a lot for merging :).

PR #1152, opened 2021-01-29
Networking: expose subnets in network data source

kayrus:
Can be used to solve the #1150 issue. Less efficient way to list and filter external network subnets:

```hcl
variable "pattern" {}
  
data "openstack_networking_network_v2" "network" {
  name = "external-network"
}

data "openstack_networking_subnet_v2" "subnet" {
  count     = length(data.openstack_networking_network_v2.network.subnets)
  subnet_id = data.openstack_networking_network_v2.network.subnets[count.index]
}

output "subnet_ids" {
  value = [for x in data.openstack_networking_subnet_v2.subnet : x.id if length(regexall(var.pattern, x.name)) > 0]
}

output "subnet_names" {
  value = [for x in data.openstack_networking_subnet_v2.subnet : x.name if length(regexall(var.pattern, x.name)) > 0]
}
```

PR #1153, opened 2021-01-29
Networking: introduce openstack_networking_subnet_ids_v2 data source

kayrus:
This is an efficient way (comparing to #1152) to filter out network subnet IDs by a certain criteria. See a #1150  for a reference.

P.S. Fixed a couple of tiny bugs and docs typos while worked on the PR. I also deprecated `dhcp_disabled` in `data.openstack_networking_subnet_v2`, because it was introduced before Terraform SDK started to support the `GetOkExists` method.


ozerovandrei (approved):
LGTM

PR #1155, opened 2021-01-30
Enable stylecheck linter

dkder3k:
For now ST1005 (capitalized errors) check is disabled as it is required to
change almost every error message. Will be done in different PR.
For #1063

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -38,7 +38,7 @@ linters:
 #    - scopelint
     - staticcheck
     - structcheck
-#    - stylecheck
+    - stylecheck
     - typecheck
     - unconvert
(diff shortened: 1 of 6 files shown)


ozerovandrei:
> Will be done in different PR.

I guess that we can also create some constants to wrap all errors since most of them differ only by resource/datasource name.

dkder3k:
> 
> 
> > Will be done in different PR.
> 
> I guess that we can also create some constants to wrap all errors since most of them differ only by resource/datasource name.

Agreed. Can do this but it'll take some time.

ozerovandrei (approved):
👍

PR #1157, opened 2021-02-01
Image: add basic auth support to download source image

pierrchen:


kayrus:
@pierrchen thanks for the PR. Did you try to set a password in a URL, e.g. `https://username:password@example.com/file.img`?

pierrchen:
@kayrus  It doesn't seems work for me and if the password contains an `@` the url became invalid. Is that expected to be a legit way to set the basic Auth for you?

kayrus:
According to https://github.com/golang/go/issues/24572 it should work. What kind of error do you receive?

kayrus:
See also https://golang.org/src/net/http/client.go#L239
https://golang.org/pkg/net/url/#Userinfo
Go automatically parses Username and Password from the URL.

pierrchen:
@kayrus sorry, I made some mistake in my experiment and I can confirm (with a standalone go programe) putting user name and password in url do works (even with `@` in the password - `curl` will complain about this).

However, I have some questions:

1. Do you think provide separate `username` and `password` will be better alternative? IMO having the url and password in the URL is a bad practice and will cause some confusions especially in the Terraform user cases I will mention below.
2. If we have the credential in the url, when credential changes, does Terraform consider it is a new resource and thus take actions it shouldn't?
3. Putting credential in the url, will cause the url to be sensitive, and thus be mask out in the `plan` output, which make it hard for review, say if the url is correctly set.

And in either case, we need additional check the response status, since an non-2xx response won't cause error be return, and thus it will an error json file be uploaded as an image.

Let me know how you want to proceed:
A. Have separate user name and password config as this patch does.
B. Putting url and password in the url, but add the response error check.

Thanks,

openstack/images_image_v2.go
```diff
@@ -100,10 +100,28 @@ func resourceImagesImageV2File(d *schema.ResourceData) (string, error) {
...
+			resp, err := client.Do(request)
 			if err != nil {
 				return "", fmt.Errorf("Error downloading image from %q", furl)
 			}
+
+			// check for credential error among other errors
+			if resp.StatusCode != http.StatusOK {
+				return "", fmt.Errorf("Error downloading image from %q, statusCode is %d", furl, resp.StatusCode)
+			}
+
```
pierrchen, line 124: @kayrus  this is the required error check I mentioned.

pierrchen:
@ozerovandrei would you like to have a review on this? thanks

openstack/images_image_v2.go
```diff
@@ -100,10 +100,28 @@ func resourceImagesImageV2File(d *schema.ResourceData) (string, error) {
 				return "", fmt.Errorf("Error creating file %q: %s", filename, err)
 			}
 			defer file.Close()
-			resp, err := http.Get(furl)
+			client := http.DefaultClient
```
kayrus, line 103: I'd prefer to use the provider's http client. In future we will control it's transport timeouts.

openstack/resource_openstack_images_image_v2.go
```diff
@@ -84,6 +84,20 @@ func resourceImagesImageV2() *schema.Resource {
...
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"local_file_path"},
+			},
+
+			"image_source_password": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
```
kayrus, line 97: Add `Sensitive: true` flag.

website/docs/r/images_image_v2.html.markdown
```diff
@@ -48,6 +48,10 @@ The following arguments are supported:
    being uploaded to Glance.
    Conflicts with `local_file_path`.
 
+* `image_source_username` - (Optional) The username of basic auth to download `image_source_url`.
+
+* `image_source_password` - (Optional) The password of basic auth to download `image_source_url`.
```
kayrus, line 53: Add a note that a password will be kept in a plain text. See https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/7a0e3834382cde6e494cc14c6ebc1adc5aba623b/website/docs/r/identity_application_credential_v3.html.markdown#L13 for example

openstack/resource_openstack_images_image_v2.go
```diff
@@ -84,6 +84,20 @@ func resourceImagesImageV2() *schema.Resource {
 				ConflictsWith: []string{"local_file_path"},
 			},
 
+			"image_source_username": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
```
kayrus, line 90: ForceNew will recreate a previous image, when the credentials are changed.

openstack/resource_openstack_images_image_v2.go
kayrus: ForceNew will recreate a previous image, when the credentials are changed.

kayrus (changes requested):
> Do you think provide separate username and password will be better alternative? IMO having the url and password in the URL is a bad practice and will cause some confusions especially in the Terraform user cases I will mention below.

Yes, I agree.

> If we have the credential in the url, when credential changes, does Terraform consider it is a new resource and thus take actions it shouldn't?

Yes it will consider it as a new resource.

> Putting credential in the url, will cause the url to be sensitive, and thus be mask out in the plan output, which make it hard for review, say if the url is correctly set.

Agree. See my comments.

> And in either case, we need additional check the response status, since an non-2xx response won't cause error be return, and thus it will an error json file be uploaded as an image.

Agree.

> A. Have separate user name and password config as this patch does.
> B. Putting url and password in the url, but add the response error check.

Both is fine.

openstack/images_image_v2.go
pierrchen: will do.

openstack/resource_openstack_images_image_v2.go
pierrchen: done.

website/docs/r/images_image_v2.html.markdown
pierrchen: done.

openstack/resource_openstack_images_image_v2.go
pierrchen: done

pierrchen:
Not familiar with how to interpreter the acceptance test failure so not sure if it is related with the my change or it is false positive?

website/docs/r/images_image_v2.html.markdown
kayrus: Please use a special syntax at the beginning of the doc like in the example.

openstack/images_image_v2.go
```diff
@@ -100,10 +100,28 @@ func resourceImagesImageV2File(d *schema.ResourceData) (string, error) {
 				return "", fmt.Errorf("Error creating file %q: %s", filename, err)
 			}
 			defer file.Close()
-			resp, err := http.Get(furl)
+			client := &http.Client{}
```
kayrus, line 103: you haven't used the provider's http client. You need to pass the `imageClient` pointer to the func, then re-use its client like:
```suggestion
			client := serviceClient.ProviderClient.HTTPClient
```

kayrus (changes requested):

pierrchen:
Not entirely sure about this [comments](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1157#discussion_r570765797) and it was marked as resolved so I didn't make any change. Let me know if I should.

openstack/images_image_v2.go
pierrchen: done.

website/docs/r/images_image_v2.html.markdown
```diff
@@ -48,6 +48,11 @@ The following arguments are supported:
...
+   Warning: the password will be stored in the Terraform state as plain-text.
```
kayrus, line 54: Please use the syntax below and put it in top.

```
~> **Note:** All arguments including the source image URL password will be
stored in the raw state as plain-text. [Read more about sensitive data in
state](https://www.terraform.io/docs/language/state/sensitive-data.html).
```

pierrchen: done

kayrus (approved):
LGTM
Let's wait for openlab issues to be fixed.

jeblair:
recheck

kayrus:
@pierrchen merged, thank you

PR #1159, opened 2021-02-02
Core: extend provider "max_retries" argument

### LINKED ISSUE #1156: Core: provider should honor the 429 response code with a Retry-After header
### Affected Resource(s)
- all

### Expected Behavior
When OpenStack responses with a 429 response code and a `Retry-After` header, terraform should not fail, but sleep for a `Retry-After` value and retry.

### Actual Behavior
When OpenStack responses with a 429 response code, terraform fails.

### Steps to Reproduce

n/a

### Important Factoids

We need to introduce a new provider option, or reuse already existing `max_retries` with an extra help in documentation.

### References
- https://github.com/gophercloud/gophercloud/pull/2097

cc @dkistner @RaphaelVogel

kayrus:
This PR allows to configure terraform to detect the 429 response code with a `Retry-After` header and sleep to reduce the load on API.

Fixes #1156


kayrus:

ozerovandrei (approved):
LGTM

Issue #1104, opened 2020-11-30
Networking quota cannot be managed by region


rodcloutier:
Networking Quotas are bound to project and region. The same project can have different quotas in each region.

Different clients witll be returned for each region, as presented here.

https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/e33bbfb06108f6b3538ee8d61a97524149ede439/openstack/resource_openstack_networking_quota_v2.go#L101

Currently, each quota managed returns the same id (the project id) which makes it impossible to manage quotas for the same project across multiple regions.

https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/e33bbfb06108f6b3538ee8d61a97524149ede439/openstack/resource_openstack_networking_quota_v2.go#L134

We can try to import the current quotas but each region returns the same id even though the quota values differ.

ozerovandrei:
It can be managed but it should be done in separated directories (so Terraform will use separated states).

But yeah, ID can be changed to use this form for example: `id_region`.
This change should also be backward compatible so we should parse `id` or `id_region` forms of quota ID so current users won't be affected by that change.

PR #1169, opened 2021-02-06
Add resource openstack_lb_quota_v2

### LINKED ISSUE #1168: [New resource] Octavia quota
Octavia allows to set quotas per project. Add support for this resource.
This is part of [Gophercloud](https://github.com/gophercloud/gophercloud/tree/v0.15.0/openstack/loadbalancer/v2/quotas)
[Api docs](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=list-quota-detail#show-project-quota)

nikParasyr:
Close #1168 

Adds support for octavia quota.

In my understanding this cannot be implemented for neutron LBaaS

### DIFF
--- a/openstack/resource_openstack_lb_quota_v2.go
+++ b/openstack/resource_openstack_lb_quota_v2.go
@@ -1,0 +1,248 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/quotas"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceLoadBalancerQuotaV2() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceLoadBalancerQuotaV2Create,
+		Read:   resourceLoadBalancerQuotaV2Read,
+		Update: resourceLoadBalancerQuotaV2Update,
+		Delete: schema.RemoveFromState,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Update: schema.DefaultTimeout(10 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"project_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"loadbalancer": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+
+			"listener": {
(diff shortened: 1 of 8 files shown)


nikParasyr:
Some notes
- Not sure if im handling microversions correctly. L7_rule and l7_policy were added in 2.19 as seen [here](https://github.com/openstack/octavia/blob/master/octavia/api/root_controller.py#L124). and also how to have acceptance tests it in different microversions (if that is required).
- In my understanding this is not implemented in neutron LBaas that is why i return error if not `use_octavia`
- I haven't added `region` as a attribute but maybe I should(?). If this is the case should I also append it in the resource id to avoid future issues such as https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1104?

jeblair:
recheck https://github.com/theopenlab/openlab/issues/681

jeblair:
recheck
https://github.com/theopenlab/openlab/issues/681

jeblair:
recheck

nikParasyr:
recheck lbaas

ozerovandrei:
> I haven't added region as a attribute but maybe I should(?). If this is the case should I also append it in the resource id to avoid future issues such as #1104?

I think that region is needed since there can be different quota for a single project in different regions.
Adding it in ID is fine. I think we can then do a refactoring of the existing quota resources in a similar way.

nikParasyr:
@ozerovandrei Added region as an attribute and also appended on the resource ID.

website/docs/r/lb_quota_v2.html.markdown
```diff
@@ -0,0 +1,96 @@
...
+layout: "openstack"
+page_title: "OpenStack: openstack_lb_quota_v2"
+sidebar_current: "docs-openstack-resource-lb-quota-v2"
+description: |-
+  Manages a V2 quota resource within OpenStack.
+---
+
+# openstack\_lb\_quota\_v2
+
+Manages a V3 block storage quota resource within OpenStack.
```
ozerovandrei, line 11: Not `V3 block storage quota` 😉

openstack/resource_openstack_lb_quota_v2.go
```diff
@@ -0,0 +1,262 @@
...
+		_, err = quotas.Update(lbClient, projectID, updateOpts).Extract()
+		if err != nil {
+			return fmt.Errorf("Error updating openstack_lb_quota_v2: %s", err)
+		}
+	}
+
+	return resourceLoadBalancerQuotaV2Read(d, meta)
+}
+
+func parseLBQuotaID(id string) (string, string, error) {
```
ozerovandrei, line 251: Could you move this function to file with other helpers https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/lb_v2_shared.go and implement an unit test for it.

ozerovandrei (changes requested):

website/docs/r/lb_quota_v2.html.markdown
nikParasyr: :innocent: fixed :)

openstack/resource_openstack_lb_quota_v2.go
nikParasyr: done as well

ozerovandrei (approved):
LGTM

PR #1171, opened 2021-02-09
networking v2: test scenario for neutron 400 ExternalIpAddressExhausted response code

kayrus:
testing a fix for https://github.com/theopenlab/openlab/issues/683


kayrus:
@ozerovandrei for some reason the FIP allocation started to fail in openlab, and neutron returns 400 (`ExternalIpAddressExhausted`) code instead of 409 (`IpAddressGenerationFailure`).

In my local environment I get 409. This patch handles 400 error code as well.

ozerovandrei (approved):

PR #1177, opened 2021-02-22
networking_quota_v2: Manage quota by region

### LINKED ISSUE #1104: Networking quota cannot be managed by region

nikParasyr:
Close #1104 

Appends region to networking_quota_v2 resource ID. This allows to manage quotas across regions for the same project.

### DIFF
--- a/openstack/networking_v2_shared.go
+++ b/openstack/networking_v2_shared.go
@@ -2,7 +2,9 @@ package openstack
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -72,4 +74,17 @@ func decodeNeutronError(body []byte) (*neutronError, error) {
 	}
 
 	return &e.NeutronError, nil
+}
+
+func parseNetworkingQuotaID(id string) (string, string, error) {
+	// Use SplitN as it is possible for a region name to contain "/"
+	idParts := strings.SplitN(id, "/", 2)
+	if len(idParts) < 2 {
+		return "", "", fmt.Errorf("Unable to determine networking quota ID %s", id)
+	}
+
+	projectID := idParts[0]
+	region := idParts[1]
+
+	return projectID, region, nil
 }
--- a/openstack/resource_openstack_networking_quota_v2.go
+++ b/openstack/resource_openstack_networking_quota_v2.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/quotas"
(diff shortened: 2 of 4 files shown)


openstack/resource_openstack_networking_quota_v2.go
```diff
@@ -140,19 +143,33 @@ func resourceNetworkingQuotaV2Create(d *schema.ResourceData, meta interface{}) e
...
-	networkingClient, err := config.NetworkingV2Client(GetRegion(d, config))
+	region := GetRegion(d, config)
+	networkingClient, err := config.NetworkingV2Client(region)
 	if err != nil {
 		return fmt.Errorf("Error creating OpenStack networking client: %s", err)
 	}
 
-	q, err := quotas.Get(networkingClient, d.Id()).Extract()
+	// update resource ID with region to allow multi-region quota management
+	if !strings.Contains(d.Id(), region) {
```
nikParasyr, line 153: Not sure if this is the preferred way. Changing the id happens "silently" for the user.

A different approach would be to change `parseNetworkingQuotaID` to allow both project_id and project_id/region but this will lead to inconsistent resource ids as the ones created before this release will have project_id and after this release project_id/region.

nikParasyr:
I'll make similar PRs for blockstorage_quota_v2 and v3 after this is merged

openstack/networking_v2_shared.go
```diff
@@ -73,3 +75,16 @@ func decodeNeutronError(body []byte) (*neutronError, error) {
 
 	return &e.NeutronError, nil
 }
+
+func parseNetworkingQuotaID(id string) (string, string, error) {
```
nikParasyr, line 79: If we go for the same setup for all quotas this func should be moved to a more generic file if possible and be used by all quota resources

openstack/resource_openstack_networking_quota_v2.go
ozerovandrei: I guess having a single preferred way is better.
Also we shouldn't create any issues by changing ID implicitly since this resource won't be deleted because deletion of a quota resource is a no-op action.

ozerovandrei (approved):
LGTM.

Please let me know if it's ready to be merged.

nikParasyr:
@ozerovandrei  Please merge

I'll prep later this week subsequent MRs for using the same func, blockstorage_quota_v2, v3 and compute_quota

PR #1180, opened 2021-02-24
Refactor quota resources to use the same parseQuotaID

nikParasyr:
This PR changes lb_quota and networking_quota to use the same func to to parse their resource ID.
Follow up from https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1177.

Additionally updateQuotaID was added on util.go with tests, and networking_quota updated to use this also during update
Close #1184  

This will be used as requirement for follow-up PRs for blockstorage and compute quotasets

PR #1182, opened 2021-02-24
blockstorage_quotaset_v2: Add multi-region support

nikParasyr:
Similar to #1177 and #1180 this adds support for blockstorage_quota_v2


ozerovandrei (approved):
LGTM

PR #1185, opened 2021-02-24
blockstorage_quotaset_v3: Add volume type quota support

nikParasyr:
Related to: https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/979

Gophercloud added support for volume_type quota with: https://github.com/gophercloud/gophercloud/pull/2109

This PR requires both #1180 and #1183

### DIFF
--- a/openstack/resource_openstack_blockstorage_quotaset_v3.go
+++ b/openstack/resource_openstack_blockstorage_quotaset_v3.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"log"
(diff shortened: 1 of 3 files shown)


nikParasyr:
I have tested this on a test cluster and it works. I cannot update the tests to include this functionality as:
- we dont have a resource to create volume_types (I might look to add it later on)
- the existing volume types change for each cloud, so i dont want to hardcode anything that only works for openlab

ozerovandrei (approved):
LGTM

Issue #1193, opened 2021-03-10
Can't set blockstorage quota per volume type


fitbeard:
### Terraform Versions
Terraform: v0.14.7
OpenStack provider: v1.39.0

### Affected Resource(s)
blockstorage_quotaset_v2
blockstorage_quotaset_v3

### Terraform Configuration Files
```hcl
resource "openstack_blockstorage_quotaset_v3" "quotaset_1" {
  project_id = "88b769ef67994cd2910f143c87283904"
  region = "region-1"
  volumes   = 10
  snapshots = 10
  gigabytes = 20
  per_volume_gigabytes = -1
  backups = 0
  backup_gigabytes = 0
  groups = 10
  volume_type_quota = {
    volumes_standard = 10
    gigabytes_standard = 20
    snapshots_standard = 10
  }
}
```

### Debug Output
https://gist.github.com/fitbeard/0aa89097ecfe42d6f00648499ed4616c

### Expected Behavior
Volume type quotas created/updated

### Actual Behavior
Getting `"badRequest": {"code": 400, "message": "The server could not comply with the request since it is either malformed or otherwise incorrect."`

### Steps to Reproduce
Describe volume type quotas
1. `terraform apply`

### Important Factoids
OpenStack release: Ussuri
Cinder version: 16.2.1
Cinder V2 API: 2.x?
Cinder v3 API: 3.60

fitbeard:
@nikParasyr

nikParasyr:
@fitbeard can i get a gist with debug level logs?

nikParasyr:
> @fitbeard can i get a gist with debug level logs?

TF_LOG=DEBUG terraform apply

Also can you ensure that the project in question has access to the "standard" volume type? Although in that case i think the error message is different

fitbeard:
@nikParasyr by the way. This happens with application credentials and old style creds .

nikParasyr:
@fitbeard thanks for the info.

Unfortunately when adding this feature I couldn't add tests as we lack a volume_type resource --which i aim to add soon. That being said i tested it locally back then and also now, with your exact snippet and it works fine. terraform version and provider same as yours, on openstack stein clusters though although that shouldn't be an issue *in theory*....

> This happens with application credentials and old style creds .

This shouldn't be affecting it. By old style creds i guess you mean typical openrc file sourced? 

The logs unfortunately are not super helpful. Could you check a thing for me if possible:
[...]

PR #1199, opened 2021-03-12
blockstorage_quotaset_v2/3: volume_type_quota to TypeInt

nikParasyr:
Elements of volume_type_quota are TypeInt. This ensures that
the request sent to Openstack is not malformed.
Read func on both resources has been updated to set volume_type_quota
only when user is defining them. This is to avoid breaking_changes and
unnecessary updates during apply.

Unortunately, when a user defines volume_type_quota then he needs to define all keys for all volume types to avoid unnecessary 
updates. This is a bit unfriendly. I'll upload a POC with explicit type conversion that doesnt introduce this.

For: #1193

### DIFF
--- a/openstack/resource_openstack_blockstorage_quotaset_v2.go
+++ b/openstack/resource_openstack_blockstorage_quotaset_v2.go
@@ -84,6 +84,7 @@ func resourceBlockStorageQuotasetV2() *schema.Resource {
 
 			"volume_type_quota": {
 				Type:     schema.TypeMap,
+				Elem:     &schema.Schema{Type: schema.TypeInt},
 				Optional: true,
 			},
 		},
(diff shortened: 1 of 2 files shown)


nikParasyr:
@ozerovandrei Could you have an initial look here and on #1200 which is a POC? I wanted to use DiffSupressFunc for this but its only used for strings. Related ticket: https://github.com/hashicorp/terraform-plugin-sdk/issues/477

- This approach has the disadvantage of forcing the user to define all volume_type_quota and not a subset, and also read needs to be updates
- POC doesnt have these issues but does the conversion explicitly

nikParasyr:
recheck

nikParasyr:
Closing this. #1200 implements the required functionality

PR #1221, opened 2021-03-30
Allow to disable status check in resource_openstack_dns_recordset_v2

ya-isakov:
This change is similar to a498d03c3a50d9ac986010a34411655d15b4db9c.

If recordset is created in Designate, and even all the backends are down,
Designate will try to recover. If only one of the backends is down, the
recordset will be in ERROR state, but will be served by other backends.
In this case, waiting for ACTIVE state also doesn't make sense.


ya-isakov:
recheck designate

ozerovandrei (approved):
LGTM

ya-isakov:
@ozerovandrei Could we have a new release, please?

ozerovandrei:
@ya-isakov Hello. Release is published!

ya-isakov:
Thank you!

PR #1225, opened 2021-04-09
lb_loadbalancer_v2: Add availability_zone

### LINKED ISSUE #1090: port_id and availability_zone arguments on openstack_lb_loadbalancer_v2 resource
Hi there,

Can we add port_id and availability_zone arguments to openstack_lb_loadbalancer_v2 resource. The provider gives us information on ports which we have to use. Currently we create lb resource and import the created resource into TF. It would be tremendously useful if we can give port_id and availability_zone as arguments in openstack_lb_loadbalancer_v2.

Thanks

nikParasyr:
Add availability_zone attribute to loadbalancer_v2

Close #1090

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -12,7 +12,6 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 
 	octavialisteners "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
-	octavialoadbalancers "github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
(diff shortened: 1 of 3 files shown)


nikParasyr:
recheck lbaas

nikParasyr:
recheck lbaas

nikParasyr:
@ozerovandrei This is ready from my side. A couple of notes:
- I had to remove the `chooseLBV2LoadBalancerCreateOpts` helper function and move the logic within the create func. Reason was that the return was of type `neutronloadbalancers.CreateOptsBuilder` and it was barking as AvailabilityZone is not part of it. I can create 2 helper functions if you prefer it that way
- I haven't updated the tests as it is quite hard. It requires a nova aggregate with az, an octavia az-profile and then an octavia az. We don't have resources for most of these.

nikParasyr:
recheck

ozerovandrei (approved):
LGTM

Issue #1235, opened 2021-04-29
Database instance creation fails from Urssi release


dave2:
https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/2665d4c052f3224481919071355d7cdd94dd92f5/openstack/resource_openstack_db_instance_v1.go#L225

This will fail to create the database as the final state is `HEALTHY` starting with Urssi release, per the notes here:
https://docs.openstack.org/releasenotes/trove/ussuri.html

Testing for `ACTIVE` will be unreliable as the resource could go from `BUILD` to `HEALTHY` between polls.

This results in the following error:

```
Error: Error waiting for openstack_db_instance_v1 e9395b74-0b32-4b62-ab0f-49e5537eaaf0 to become ready: unexpected state 'HEALTHY', wanted target 'ACTIVE'. last error: %!s(<nil>)
```

Tested against Victoria release of Trove and provider version 1.41.0.

PR #1241, opened 2021-05-02
Test for HEALTHY not ACTIVE on database instance create

dave2:
This is needed since Urssi release of Trove.

Note: tests I have run against Urssi also highlight that the acceptance tests need some fixes (eg, flavor appears to be mandatory)

Fixes: #1235


dave2:
Instead of just breaking everything before Urssi, we now test for either target state (pre-Urssi and post-Urssi) which should be more stable. Trove doesn't have any versioning hints I can find to tell which state is correct as completed.

dave2:
Anything else I need to do for this to be merged? It's breaking on the public cloud of my employer :)

PR #1243, opened 2021-05-10
networking_rbac_policy_v2: Add address_group as object_type

nikParasyr:
Similar to #1086 this adds support for address_group.
Update docs are well

Also added `.terraform.lock.hcl` to `.gitignore` for when testing with tf >= 0.14


ozerovandrei (approved):
LGTM

PR #1248, opened 2021-05-18
Add nil check to server group hint creation

jhg03a:
I ran into a provider panic while trying to unassign a server group from a server.  By deferring to the default behavior, the nil check is sufficient to effectively unassign a server group.

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -1403,15 +1403,15 @@ func getFlavorID(computeClient *gophercloud.ServiceClient, d *schema.ResourceDat
 
 func resourceComputeSchedulerHintsHash(v interface{}) int {
 	var buf bytes.Buffer
-	
+
 	m, ok := v.(map[string]interface{})
 	if !ok {
-  		return hashcode.String(buf.String())
+		return hashcode.String(buf.String())
 	}
 	if m == nil {
-  		return hashcode.String(buf.String())
+		return hashcode.String(buf.String())
 	}
-	
+
 	if m["group"] != nil {
 		buf.WriteString(fmt.Sprintf("%s-", m["group"].(string)))
 	}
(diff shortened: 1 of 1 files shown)


jhg03a:
The test failures appear to be unrelated to the change itself but rather constructing the devstack.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1403,31 +1403,31 @@ func getFlavorID(computeClient *gophercloud.ServiceClient, d *schema.ResourceDat
...
-
-	if m["group"] != nil {
-		buf.WriteString(fmt.Sprintf("%s-", m["group"].(string)))
-	}
-
-	if m["target_cell"] != nil {
-		buf.WriteString(fmt.Sprintf("%s-", m["target_cell"].(string)))
-	}
+	if v != nil && v.(map[string]interface{}) != nil {
+		m := v.(map[string]interface{})
```
ozerovandrei, line 1407: I guess it's better not to nest all the logic under a single `if {}` and we can also cast just a single time like this:
```go
m, ok := v.(map[string]interface{})
if !ok {
  return hashcode.String(buf.String())
}
if m == nil {
  return hashcode.String(buf.String())
}

// other stuff

return hashcode.String(buf.String())
```

ozerovandrei (changes requested):

openstack/resource_openstack_compute_instance_v2.go
jhg03a: Addressed with commit

jhg03a:
The CI failures appear to be unrelated to the change not being able to find a golang compiler.

ozerovandrei:
@jhg03a could you please also resolve `golangci-lint` issue?
I can merge this without `openlab/check` since those changes looks safe.

ozerovandrei (approved):
LGTM but we need tests to be fixed first #1253

jhg03a:
Those acceptance test failures are not related to the PR, but rather https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1253

nikParasyr:
recheck

ozerovandrei (approved):

PR #1254, opened 2021-05-21
Add SUDO_PROJECT_ID for openstack_dns_recordset_v2

sebastienmusso:
Add the capability to change recordset project ID for openstack_dns_recordset_v2

### DIFF
--- a/openstack/dns_zone_v2.go
+++ b/openstack/dns_zone_v2.go
@@ -2,10 +2,9 @@ package openstack
 
 import (
 	"fmt"
-	"log"
-
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"log"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/dns/v2/zones"
@@ -70,7 +69,7 @@ func dnsClientSetAuthHeader(resourceData *schema.ResourceData, dnsClient *gopher
 
 	if len(headers) != 0 {
 		dnsClient.MoreHeaders = headers
-		log.Printf("[DEBUG] openstack_dns_zone_v2 request headers set: %#v", headers)
+		log.Printf("[DEBUG] request headers set: %#v", headers)
 	}
 
 	return nil
--- a/openstack/dns_zone_v2.go
+++ b/openstack/dns_zone_v2.go
@@ -2,9 +2,10 @@ package openstack
 
 import (
 	"fmt"
+	"log"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
-	"log"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/dns/v2/zones"
--- a/openstack/resource_openstack_dns_recordset_v2.go
+++ b/openstack/resource_openstack_dns_recordset_v2.go
@@ -88,6 +88,13 @@ func resourceDNSRecordSetV2() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+
+			"project_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -97,6 +104,10 @@ func resourceDNSRecordSetV2Create(d *schema.ResourceData, meta interface{}) erro
 	dnsClient, err := config.DNSV2Client(GetRegion(d, config))
 	if err != nil {
 		return fmt.Errorf("Error creating OpenStack DNS client: %s", err)
+	}
+
+	if err := dnsClientSetAuthHeader(d, dnsClient); err != nil {
+		return fmt.Errorf("Error setting dns client auth headers: %s", err)
 	}
 
 	records := expandDNSRecordSetV2Records(d.Get("records").([]interface{}))
@@ -116,6 +127,7 @@ func resourceDNSRecordSetV2Create(d *schema.ResourceData, meta interface{}) erro
 
 	zoneID := d.Get("zone_id").(string)
 	n, err := recordsets.Create(dnsClient, zoneID, createOpts).Extract()
+
 	if err != nil {
 		return fmt.Errorf("Error creating openstack_dns_recordset_v2: %s", err)
 	}
@@ -136,7 +148,6 @@ func resourceDNSRecordSetV2Create(d *schema.ResourceData, meta interface{}) erro
 				"Error waiting for openstack_dns_recordset_v2 %s to become active: %s", d.Id(), err)
 		}
 	}
-
 	id := fmt.Sprintf("%s/%s", zoneID, n.ID)
 	d.SetId(id)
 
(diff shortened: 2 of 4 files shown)


openstack/dns_zone_v2.go
```diff
@@ -70,7 +70,7 @@ func dnsClientSetAuthHeader(resourceData *schema.ResourceData, dnsClient *gopher
 
 	if len(headers) != 0 {
 		dnsClient.MoreHeaders = headers
-		log.Printf("[DEBUG] openstack_dns_zone_v2 request headers set: %#v", headers)
+		log.Printf("[DEBUG] request headers set: %#v", headers)
```
ozerovandrei, line 73: Why this message is changed?

openstack/resource_openstack_dns_recordset_v2.go
```diff
@@ -116,6 +127,7 @@ func resourceDNSRecordSetV2Create(d *schema.ResourceData, meta interface{}) erro
 
 	zoneID := d.Get("zone_id").(string)
 	n, err := recordsets.Create(dnsClient, zoneID, createOpts).Extract()
+
```
ozerovandrei, line 130: I don't think that separating `err` from `err != nil` is a good style.

openstack/resource_openstack_dns_recordset_v2_test.go
```diff
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"fmt"
+	//"log"
```
ozerovandrei, line 5: This should be removed.

openstack/resource_openstack_dns_recordset_v2_test.go
```diff
@@ -181,13 +212,18 @@ func testAccCheckDNSV2RecordSetExists(n string, recordset *recordsets.RecordSet)
 		if rs.Primary.ID == "" {
 			return fmt.Errorf("No ID is set")
 		}
+		//log.Printf("[DEBUG] 2 ----------------------- %#v", rs.Primary.Attributes["project_id"])
```
ozerovandrei, line 215: This should be removed.

ozerovandrei (changes requested):

openstack/dns_zone_v2.go
sebastienmusso: Besause same function is used by resource of type recordset, in order to not being specific to dns zone resources only

openstack/resource_openstack_dns_recordset_v2_test.go
sebastienmusso: Fixed

openstack/resource_openstack_dns_recordset_v2_test.go
sebastienmusso: Fixed

openstack/resource_openstack_dns_recordset_v2.go
sebastienmusso: You mean have same error message for both dns client configuration ans dns header ?

ozerovandrei: No I meant the empty line. Same error message is fine :)

sebastienmusso: Yes right done in the last commit

ozerovandrei:
Looks fine but I want to run `recheck designate` in this PR after https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1253 is resolved.

nikParasyr:
recheck

nikParasyr:
recheck designate

sebastienmusso:
> Build succeeded.
> 
> - [terraform-provider-openstack-unittest-foo ](https://logs.openlabtesting.org/logs/54/1254/651571abb8abeadf19a67af12d57227ec52cb0e2/check/terraform-provider-openstack-unittest-foo/b82bd9b/) : SUCCESS in 4m 19s
> - [terraform-provider-openstack-acceptance-test-foo ](https://logs.openlabtesting.org/logs/54/1254/651571abb8abeadf19a67af12d57227ec52cb0e2/check/terraform-provider-openstack-acceptance-test-foo/f2f848c/) : SUCCESS in 2h 15m 18s
> - [terraform-provider-openstack-acceptance-test-admin-foo ](https://logs.openlabtesting.org/logs/54/1254/651571abb8abeadf19a67af12d57227ec52cb0e2/check/terraform-provider-openstack-acceptance-test-admin-foo/e931b38/) : SUCCESS in 49m 45s
> 

Yes !!! Here we go

nikParasyr:
> Build failed.
> 
>     * [terraform-provider-openstack-acceptance-test-designate-foo ](https://logs.openlabtesting.org/logs/54/1254/651571abb8abeadf19a67af12d57227ec52cb0e2/recheck-designate/terraform-provider-openstack-acceptance-test-designate-foo/770c676/) : FAILURE in 39m 38s

@sebastienmusso This test-set is more important and it is failing. I haven't had the time to go through the logs etc but by searching for `FAIL` you should be able to see which tests fail.

sebastienmusso:
Hi,

it seems to miss some permission within openstack api:

```
POST http://192.168.0.21/identity/v3/projects], error message: {"error":{"code":403,"message":"You are not authorized to perform the requested action: identity:create_project.","title":"Forbidden"}}
```

In the test, I create an openstack project and a dns record inside it.
The test user does't seems to have sufficient permissions.

nikParasyr:
@sebastienmusso sorry for the late reply. After having a look, it is a bit harder to properly fix the test scenario. I'll give some context:

In general a job(and all tests in that job) run as a single openstack user, who is either a `member` (the typical openstack `user` role) or an `admin`. Some tests need to be run on a job that is admin and some on the one that is a member. Unfortunately currently we don't have a more dynamic way of changing the permissions of a user within a job.  For that reason our typical check has 2 jobs(excluding the unit tests), one for the admin tests and one for the user tests.

The designate job runs as a `member`, which means you don't have permissions to create a project (which is required for your test case). Some of the options:
1. You don't create a project but you get the current project_id that the user belongs to (basically the env var OS_PROJECT_ID, a helper function could do that and then you pass it as a var in your`testAccDNSV2RecordSetDifferentProject`) and use that (i'm not 100% sure this will be enough).
2. We create a new job for designate where you have admin credentials therefore you can create a project. In this case and based on your docs(`The ID of the project DNS zone is created
  for, sets X-Auth-Sudo-Tenant-ID header (requires an assigned 
  user role in target project)`) you should also add a role association to that project which is currently missing.
  
  
My opinion is that if case 1 works then that's better cause we don't add yet another job. Otherwise for the option 2 i can create an MR with the new admin job, unfortunately it might take some time for our test infrastructure to be updated due to how its being managed. After that you can rebase and update your test accordingly. I'd like @ozerovandrei opinion on this. It's quite some work to cover one test case unfortunately.

Hopefully this explains some things to you.

ozerovandrei:
It seems that the first way is easier if it worked.

sebastienmusso:
I simplify test scenario. No more project are created and dns entry is created with "project_id" argument

nikParasyr:
recheck designate

sebastienmusso:
recheck designate

sebastienmusso:
Here we go !

sebastienmusso:
recheck designate

ozerovandrei (approved):
LGTM

Thank you @sebastienmusso

PR #1260, opened 2021-06-07
Expose database instance IP addresses

dave2:
When creating a database instance with only a network to attach to, Trove will automatically create a port on that network and have an IP address assigned. However, this can only be see on the query API in the `ip` or `addresses` attributes, you cannot see the network port uuid. (`ip` is deprecated, but the only attribute GopherCloud supports).

Expose the assigned addresses as a new exported list `addresses` which is just a straight copy of the `ip` attribute on the query response.

Tested against Trove Victoria.


ozerovandrei (approved):
LGTM

Issue #1253, opened 2021-05-21
Acc Tests: OpenLab Failure


nikParasyr:
Acceptance tests are failing. I speculate the reason is that ubuntu bionic support was [dropped](https://opendev.org/openstack/devstack/commit/7ad4cd07c8bf4f302acc4fc6684e362309332c9d) recently from devstack.

I'll have a look to update the openlab jobs to use `focal` instead. 

Another issue worth noting, coming from devstack:
Devstack has dropped support for blockstorage v2 api as seen [here](https://opendev.org/openstack/devstack/commit/35cec0d7c0857d76d3ea0b52b97f2a166c04c13e). I'm not sure how we can handle the tests for the blockstorage v2 api

ozerovandrei:
I guess we should disable V2 tests and add a new testing environment variable that can enable them.
And not set that environment in the Devstack.

nikParasyr:
@ozerovandrei sorry for the delay, had some IRL stuff to take care of. Some notes:

- I'm pushing to get the focal PR merged, hopefully this will be done soon
- For the BlockstorageV2 tests as you mention we should disable them by default and only enable in a specific env. We should pin both openstack version and devstack version. I'll have a look after the focal PR goes through
- In general i know that gophercloud has some of its zuul config to define jobs etc in the gophercloud repo, and that allows them to make changes such as the focal etc easier and faster. Is it something we should/could do as well, so that we have more control on the test env?

ozerovandrei:
@nikParasyr Thank you.
As for the last question. If you have the time to improve testing jobs/env you can do it of course :)

nikParasyr:
Unfortunately, the old jobs run for new PRs. I expect this is similar to what happened when the `admin` job was enabled. Zuul required a restart to pick up the new job configs (see https://github.com/theopenlab/openlab/issues/717). I have opened a new ticket for this on openlab requesting a restart of zuul (https://github.com/theopenlab/openlab/issues/777). I hope it will be seen and done soon.

nikParasyr:
@ozerovandrei openlab is doing its magic again :)
I have kicked most pipelines on PRs.

There are still some to-dos to resolve this issue fully:
- [ ] double-check that the job for blockstorage v2 works as expected
- [ ] remove the job definition from openlab (this might take a while)
- [ ] remove the `-foo` from the jobs defined here (this might require a zuul restart once again, so we need to find an appropriate window for that)

[...]

PR #1263, opened 2021-06-21
Add ci configuration for openlab zuul

nikParasyr:
For #1253 

Copy the zuul job configurations from openlab to allow
easier/faster updates.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,10 +1,220 @@
+# Terraform-provider-openstack jobs
+- job:
+    name: terraform-provider-openstack-unittest-foo
+    parent: golang-test-main
+    description: |
+      Run terraform-provider-openstack unit test
+    run: .zuul/playbooks/terraform-provider-openstack-unittest/run.yaml
+    nodeset: ubuntu-xenial-ut
+    vars:
+      go_version: '1.14'
+
+## ========== terraform-provider-openstack + openstack jobs ==============
+
+### parent job
+- job:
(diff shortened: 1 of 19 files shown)


nikParasyr:
recheck-bsv2

nikParasyr:
recheck bsv2

nikParasyr:
recheck bsv2-admin

nikParasyr:
recheck lbaas

nikParasyr:
recheck lbaas

nikParasyr:
@ozerovandrei Couple of things:
- I copied the playbooks and added the job definitions based on what was on openlab. I had to change the name of the jobs (appended a `-foo`) because they were duplicate. If we go forward with this I can make a PR to openlab to remove the definitions from there and then i can remove `-foo` from here.
- Added 2 new jobs for BlockStorageV2 tests and updated all the tests accordingly. The jobs are based a bit on https://review.opendev.org/c/openstack/tempest/+/792025/.

Currently i'm facing 2 problems that I'm not sure what the reason is.
- The jobs do run and finish on openlab (i can see/find the logs etc ) but the PR is not updated with the result. Not sure why. Wonder if it is a permissions thingy or something like that that i dont have permissions to check.
- The 2 new jobs that i have added do not get triggered by comments. Not sure why. There was a similar issue when i added the admin job and a restart of zuul was required.

nikParasyr:
recheck bsv2

ozerovandrei (approved):
Thank you @nikParasyr !

PR #1267, opened 2021-06-28
Check CIDR value in resource_openstack_networking_subnet_v2

pbasov:
This check catches terraform state mismatch with Openstack subnet in case of input error

Consider the following example:
```
resource "openstack_networking_subnet_v2" "subnet" {
  name             = "subnet"
  network_id    = openstack_networking_network_v2.network.id
  cidr                = "192.168.1.0/20"
  ip_version      = 4
  tenant_id       = openstack_identity_project_v3.project.id
}
```

Terraform plans subnet with a `cidr` attribute value of `192.168.1.0/20`. This adress is a part of the `192.168.0.0/20` subnet, so Openstack Neutron automatically assumes it to be `192.168.0.0/20`, which is a valid CIDR. Subnet creates successfully.

The problem is, that terraform will store the information for the `networking_subnet_v2` resource with `cidr` attribute value set to `192.168.1.0/20`, but every following terraform executions will read `192.168.0.0/20`from Neutron API, since it is the correct value and have been automatically corrected by Neutron API, creating a mismatch with the state. Hence all the following runs will try to re-create the subnet as `ForceNew` is set for `cidr` attribute.

This commit prevents erroneous by checking if input `cidr` attribute matches the network address and throws an error otherwise.

### DIFF
--- a/openstack/resource_openstack_networking_subnet_v2.go
+++ b/openstack/resource_openstack_networking_subnet_v2.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"log"
+	"net"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/customdiff"
@@ -50,6 +51,10 @@ func resourceNetworkingSubnetV2() *schema.Resource {
 				Optional:      true,
 				Computed:      true,
 				ForceNew:      true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.IsCIDR,
+				},
 			},
 
 			"prefix_length": {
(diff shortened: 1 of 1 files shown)


ozerovandrei (changes requested):
It's better to use the SDK helper function that is available in v1 and in v2 versions: https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.17.2/helper/validation/network.go#L99

pbasov:
> It's better to use the SDK helper function that is available in v1 and in v2 versions: https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.17.2/helper/validation/network.go#L99

I can add `IsCidr` validation function, to skip the error check for the subnet calc, but that func doesn't solve the described problem by itself. I see there's also a `IsCIDRNetwork` func, but I'm not sure we have the input for it:

https://github.com/hashicorp/terraform-plugin-sdk/blob/2c03a32a9d1be63a12eb18aaf12d2c5270c42346/helper/validation/network.go#L123

ozerovandrei (approved):

PR #1275, opened 2021-08-09
Data source for openstack_identity_domain_v3

ftarasenko:
Implements a new data source for openstack_identity_domain_v3.
Required to get ID of openstack keystone's domain.

WIP: Needs documentation and review.

### DIFF
--- a/openstack/data_source_openstack_identity_domain_v3.go
+++ b/openstack/data_source_openstack_identity_domain_v3.go
@@ -1,0 +1,88 @@
+package openstack
+
+import (
+	"fmt"
+	"log"
+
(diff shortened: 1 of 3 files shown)


ftarasenko:
_Projects may, in addition to acting as containers for OpenStack resources, act as a domain (by setting the attribute is_domain to true), in which case it provides a namespace in which users, groups and other projects can be created. In fact, a domain created using the POST /domains API will actually be represented as a project with is_domain set to true with no parent (parent_id is null)._
Not required, **openstack_identity_project_v3** with **is_domain = true** parameter should be used.

PR #1297, opened 2021-10-06
Correctly delete openstack_networking_router_interface_v2 resources

pascal-hofmann:
When creating multiple `openstack_networking_router_interface_v2`
resources for different IPv6 subnets in the same network, openstack will
attach these to the same port on the router.

On destroy the current implementation will pass the `port_id` to
`routers.RemoveInterface`, which will trigger the deletion of the whole
port, including other `openstack_networking_router_interface_v2` resources
on the same port.

This commit changes this behaviour (so that only the correct router interface
gets deleted) and adds a test case that ensures this does not break
again in the future.

Note: We were affected by this bug. In our case we deleted an `openstack_networking_router_interface_v2` via `terraform destroy -target …`, but terraform-provider-openstack silently deleted the whole port including another, important `openstack_networking_router_interface_v2` resource which was crucial for our production setup.

### DIFF
--- a/openstack/networking_router_interface_v2.go
+++ b/openstack/networking_router_interface_v2.go
@@ -38,6 +38,12 @@ func resourceNetworkingRouterInterfaceV2DeleteRefreshFunc(networkingClient *goph
 			PortID:   d.Get("port_id").(string),
 		}
 
+		if removeOpts.SubnetID != "" {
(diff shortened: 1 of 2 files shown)


pascal-hofmann:
No idea why the status checks fail. I'm lost in the thousands of log lines. Can someone help with this?

PR #1299, opened 2021-10-13
image: set forcenew=false on min_disk_gb, min_ram_mb, protected

### LINKED ISSUE #1287: openstack_images_image_v2 forces replacement when changing protected property
### Terraform Version
```
Terraform v1.0.5
on linux_amd64
+ provider registry.terraform.io/terraform-provider-openstack/openstack v1.43.0
```

### Affected Resource(s)
- openstack_images_image_v2

```hcl
resource "openstack_images_image_v2" "images" {
  name             = "my-image-name"
  local_file_path  = "my-image.qcow2"
  container_format = "bare"
  disk_format      = "qcow2"
  protected        = true
}
```

### Expected Behavior
When `protected` changes, I expect terraform to change the value in-place. I can do it manually via the openstack web interface.

### Actual Behavior
Forces a replacement

```
~ protected        = false -> true # forces replacement
```

### Steps to Reproduce
1. Deploy an image using terraform, without setting protected value. Default is false
2. Add `protected = true`, and run apply

rossengeorgiev:
On images allow setting these properties in-place: `min_disk_gb`, `min_ram_mb`, `protected`

Closes #1287

⚠ I'm not sure whether this is all I need to do to fix this. There maybe code and tests that need to be modified

### DIFF
--- a/openstack/resource_openstack_images_image_v2.go
+++ b/openstack/resource_openstack_images_image_v2.go
@@ -109,7 +109,6 @@ func resourceImagesImageV2() *schema.Resource {
 			"min_disk_gb": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				ForceNew:     true,
 				ValidateFunc: validation.IntAtLeast(0),
 				Default:      0,
 			},
@@ -117,7 +116,6 @@ func resourceImagesImageV2() *schema.Resource {
 			"min_ram_mb": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				ForceNew:     true,
 				ValidateFunc: validation.IntAtLeast(0),
 				Default:      0,
 			},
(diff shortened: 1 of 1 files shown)


openstack/resource_openstack_images_image_v2.go
```diff
@@ -109,15 +109,15 @@ func resourceImagesImageV2() *schema.Resource {
 			"min_disk_gb": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				ForceNew:     true,
+				ForceNew:     false,
```
ozerovandrei, line 112: We usually don't add `ForceNew` field if it doesn't need to be set to `true` since it's `false` by default.
Could you just remove those lines?

ozerovandrei (changes requested):

openstack/resource_openstack_images_image_v2.go
rossengeorgiev: Done. Does any code need to be modified beyond that to address the issue?

ozerovandrei (approved):
LGTM, thank you @rossengeorgiev

PR #1303, opened 2021-10-25
Retry reading DNS zone and Compute instance state after creation

pbasov:
We had some issues with our cloud which buckled under a high load, which resulted in a bunch of 504 errors.

Provider didn't handle the errors during async operations with DNS zones and Compute instances, which resulted in mismatched states when cloud resources were actually created, but didn't go into ACTIVE state before an intermittent API error.

This change let's the provider keep refreshing the state through 502 and 504 errors, to prevent missing resources in state, that are already present on the cloud.


ozerovandrei:

ozerovandrei (approved):
LGTM

PR #1304, opened 2021-10-28
Prevent not specified quota values are being set to 0

randomswdev:
Fixes the issue https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/973

### DIFF
--- a/openstack/resource_openstack_compute_quotaset_v2.go
+++ b/openstack/resource_openstack_compute_quotaset_v2.go
@@ -139,36 +139,64 @@ func resourceComputeQuotasetV2Create(ctx context.Context, d *schema.ResourceData
 	}
 
 	projectID := d.Get("project_id").(string)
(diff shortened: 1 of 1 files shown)


ozerovandrei:

StopMotionCuber:
I'm adapting this provider for our division; from the [docs](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_quotaset_v2) I get the impression that setting not specified quota values to 0 is actually the behavior that is documented, as there is "Note: This resource has all-in creation so all optional quota arguments that were not specified are created with zero value."

Would I be correct that this comment now is outdated for `resource_openstack_compute_quotaset_v2` with this merge?

I still observe the behavior of quota values being set to 0 for `openstack_blockstorage_quotaset_v3` but could work around that by fetching the value before and setting it to the value I just fetched.

PR #1305, opened 2021-10-29
Flag tempurl url attribute as sensitive

wilas:
Tempurl URL shouldn't be visible in plaintext by default.

URL should be known only by those who received it explicit rather than everyone with ability to see terraform plan.
Tempurl might point to object with sensitive content.

`nonsensitive()` function can be used in case someone prefer display tempurl in plaintext in terraform plan/output.


ozerovandrei:

ozerovandrei (approved):
LGTM thank you @wilas

PR #1309, opened 2021-11-04
r/d/networking_network_v2: add providerext attributes

yanndegat:
* `physical_network` - The physical network where this network is implemented.
* `segmentation_id` - An isolated segment on the physical network.
* `network_type` - The type of physical network.

### DIFF
--- a/openstack/data_source_openstack_networking_network_v2.go
+++ b/openstack/data_source_openstack_networking_network_v2.go
@@ -113,6 +113,38 @@ func dataSourceNetworkingNetworkV2() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"network_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"physical_network": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"segmentation_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"segments": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"network_type": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"physical_network": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"segmentation_id": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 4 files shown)


TomEros (approved):
Looks cool

yanndegat:
hi @jtopjian 

i have an issue with the acctests on this PRs.

i'm not sure the failing tests are due to my PRs.
they are related to the `openstack_networking_trunk_v2` resource and according tests. but my PR only adds attributes to a network resource.

here's the error raised by the test on my PR:
 Error: Error creating openstack_networking_trunk_v2: Resource not found with openstack_networking_trunk_v2.trunk_1,

& when i locally run the same test, on the "main" branch, i have the same error.
do you have any clue on this ?

thanks a lot

jtopjian:
@yanndegat "Resource not found" might be that the trunk extension is not enabled - maybe removing it or not enabling it was a recent change in the upstream OpenStack code?

yanndegat:
hi @jtopjian 

thanks for your feedback. Is there anything i should do ?

yanndegat:
hi @ozerovandrei

i have CI issues on this PR, do you know what i should do to get it fixed ?

thanks a lot

openstack/data_source_openstack_networking_network_v2.go
```diff
@@ -245,6 +277,21 @@ func dataSourceNetworkingNetworkV2Read(ctx context.Context, d *schema.ResourceDa
 	d.Set("mtu", network.MTU)
 	d.Set("dns_domain", network.DNSDomain)
 	d.Set("region", GetRegion(d, config))
+	d.Set("network_type", network.NetworkType)
+	d.Set("physical_network", network.PhysicalNetwork)
+	d.Set("segmentation_id", network.SegmentationID)
+
+	// Set the segments
+	segments := make([]map[string]interface{}, len(network.Segments))
```
ozerovandrei, line 285: Shouldn't it be
```go
segments := make([]map[string]interface{}, 0, len(network.Segments))
```

Otherwise we will get segments slice with some empty elements at the beginning: https://go.dev/play/p/HAEUmMzyJ56

ozerovandrei:
Trunk tests will be disabled https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1323

openstack/data_source_openstack_networking_network_v2.go
yanndegat: updated. thanks

website/docs/d/networking_network_v2.html.markdown
```diff
@@ -67,3 +67,7 @@ are exported:
   is enabled
 * `subnets` - A list of subnet IDs belonging to the network.
 * `all_tags` - The set of string tags applied on the network.
+* `physical_network` - The physical network where this network is implemented.
```
ozerovandrei, line 70: I guess that datasource docs should also have `segments` in attributes.

ozerovandrei (changes requested):

website/docs/d/networking_network_v2.html.markdown
yanndegat: done. thanks for pointing this one

ozerovandrei (approved):
LGTM

PR #1314, opened 2021-11-16
ComputeV2: add "device_name" field

krembu:
Add "device_name" field into the "block_device"
attribute of the Compute V2 Instance resource.


kayrus:
According to the nova API note:

> Nova libvirt driver no longer honors a user-supplied device name

I'm not sure about other hypervisors, but usually disk lookup refers to openstack disk UUID, e.g. `/dev/disk/by-???`. @ozerovandrei @nikParasyr  I think this PR can be closed.

PR #1316, opened 2021-11-18
fixing network quota resource

n-marton:
this change is required to make networking_quota_v2 resource to work properly

this problem with the origin solution, is that it doesn't "honor" the omitempty setting of keys in the underlaying lib

.Get() defaults to 0 if there is no key, so even undefined keys were passed to openstack api with 0 value

with the .GetOk() approach it is possible to change each quota individually without affecting any other, as the underlaying lib and the api supports


nikParasyr:
recheck

ozerovandrei (approved):
LGTM thank you @n-marton

n-marton:
awesome!

PR #1326, opened 2021-12-03
Octavia by default as Neutron LBAAS is deprecated

TomEros:
The Neutron LBAAS is deprecated :
https://wiki.openstack.org/wiki/Neutron/LBaaS/Deprecation
It has been replaced by Octavia, so set Octavia to the default one

This can be confusing since the error returned by the provider is not
very clear even if you put the provider in TRACE ( it will try to
contact neutron for creation the loadbalancer )


ozerovandrei (approved):
LGTM

PR #1334, opened 2021-12-11
Remove `ForceNew` from network segments

ozerovandrei:
Use the same logic as in `physical_network`, `network_type`,
`segmentation_id` arguments.


ozerovandrei:
recheck

PR #1386, opened 2022-06-21
Fix fetching fixed_ip attribute

### LINKED ISSUE #1384: When port is created, fixed_ip is empty if we have more than one subnet (ipv4 and ipv6)
Hi there,

When I create a port using the resource openstack_networking_port_v2, I am unable to get information inside the fixed_ip (empty array).
This only happens if I have more than one subnet. Typically in my case I have two subnets IPV4 and IPV6.

This is the output when I do openstack port show :
```
| fixed_ips               | ip_address='X.X.X.X', subnet_id='667a3609-961b-4ecf-a027-7c5d9da8b7bf'                                                                                                                     
|                         | ip_address='eeee:yyyy:bbb:aaa::zzzz', subnet_id='d05d870d-64fa-427e-ba9d-61d08a5d2eac'
```

PS: one thing I noticed is that on the openstack api the attribute is called fixed_ip**s** which is not the case on the resource attribute

### Terraform Version
1.1.5

### Affected Resource(s)

```hcl
resource "openstack_networking_port_v2" "ext_port" {
  count = var.nbinstances

  name               = local.hostnames[count.index]
  network_id         = var.ext_net_id
  admin_state_up     = "true"
  security_group_ids = [var.secgroup_id]

  lifecycle {
    ignore_changes = [device_id, device_owner, qos_policy_id, tags]
  }
}
```

### Expected Behavior
I should have got an array inside the fixed_ip attribute with the ip and subnet_id for each entry

### Actual Behavior
fixed_ip is an empty array

### Steps to Reproduce
1. create a network with two subnets ipv4 and ipv6
2. create a port on the network created previously
3. try to get the content of fixed_ip attributes
4. do openstack port show to make sure that fixed_ip has both subnets and IPs

othman-essabir:
This PR fixes #1384


othman-essabir:
@ozerovandrei lint errors are fixed. May please approve the workflow ?
Thanks

kayrus (changes requested):
Thanks for you PR. Could you adjust docs correspondingly? And fix the `openstack_networking_port_ids_v2` data source as well.

kayrus:
@othman-essabir I updated your PR with changes to keep backward compatibility.

kayrus:
The problem for this fix is that fetching fixed_ips list must be done in sets, because neutron API returns a list of fixed IPs in a random order. Changing `TypeList` to `TypeSet` would break backward compatibility and existing terraform scripts will need to be rewritten.

@ozerovandrei @nikParasyr what do you think?

nikParasyr:
> Changing TypeList to TypeSet will break backward compatibility and existing terraform scripts need to be rewritten.

I think this is the proper way forward but it is a breaking change and we would have to change quite some tests etc as you mention. Since `ports` is a very basic and major resource that is used a lot, many users will be affected. I think we should consider planning a major release and include this change to it. I think there are more breaking changes that we could include. Personally I would like to refactor all lb resources as currently octavia and LBaaS code is quite intertwined and has stopped us from easily adding new features that octavia is releasing quite often(that being said I'm also lacking the time to refactor it).

So to sum up:
- i think the way forward would be to make the change from `TypeList` to `TypeSet`
- since its going to be a breaking change to a rather commonly used resource it should come with a major release
- we should discuss whether and when we can make a major release and what changes we can incorporate to it

@ozerovandrei looking for your input as well

kayrus:
@nikParasyr the only one thing that doesn't really prevent us on switching to TypeSet is that this field was never a computed one. It was always an input field and it's less probable that this input field was used as an input for other resources.

nikParasyr:
> the only one thing that doesn't really prevent us on switching to TypeSet is that this field was never a computed one

ok. It's a go from my side when tests etc are fixed as well

Issue #1366, opened 2022-04-08
Zuul testing failing


flashjmp:
Hi everybody,

this in reference to the failing acceptance tests when a PR is created. As per the devstack project the currently used "devstack wrapper script" should not be used anymore and the usage of zuul native "zuul/zuul-jobs ensure-devstack" role is recommended as per (https://storyboard.openstack.org/#!/story/2009973).

@nikParasyr , as per PR #1263  you seem to have setup the zuul configuration, maybe you can have a go at this topic?

Cheers,
Thomas

PR #1363, opened 2022-04-04
Added monitor_port and monitor_address to lb_members_v2

flashjmp:
Added the missing arguments for monitor_address and monitor_port to the lb_members_v2 resource.

This should fix my issue #1360

PR #1390, opened 2022-07-07
Added monitor-address and monitor-port to openstack_lb_member_v2

luigizhou:
Added missing argument for monitor_address and monitor_port to the lb_member_v2 resource. (different than lb_members_v2)
Replaced the pools library with the one from octavia loadbalancer instead of neutron.

Doesn't seem to break functionality from my test, but happy to review this more in depth to understand the implication of such change

### DIFF
--- a/openstack/resource_openstack_lb_member_v2_test.go
+++ b/openstack/resource_openstack_lb_member_v2_test.go
@@ -330,7 +330,7 @@ resource "openstack_lb_member_v2" "member_2" {
   protocol_port = 8080
   pool_id = "${openstack_lb_pool_v2.pool_1.id}"
   subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
-  monitor_address = "192.168.199.111
+  monitor_address = "192.168.199.111"
   monitor_port = 8080
 
   timeouts {
--- a/openstack/resource_openstack_lb_member_v2_test.go
+++ b/openstack/resource_openstack_lb_member_v2_test.go
@@ -96,6 +96,42 @@ func testAccCheckLBV2MemberExists(n string, member *pools.Member) resource.TestC
 	}
 }
 
+func TestAccLBV2Member_monitor(t *testing.T) {
+	var member1 pools.Member
+	var member2 pools.Member
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckNonAdminOnly(t)
+			testAccPreCheckLB(t)
+			testAccPreCheckUseOctavia(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckLBV2MemberDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: TestAccLbV2MemberConfigBasic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLBV2MemberExists("openstack_lb_member_v2.member_1", &member1),
+					testAccCheckLBV2MemberExists("openstack_lb_member_v2.member_2", &member2),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_address", "192.168.199.110"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "8080"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_2", "monitor_address", "192.168.199.111"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "8080"),
+				),
+			},
+			{
+				Config: TestAccLbV2MemberConfigUpdate,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_2", "monitor_address", "192.168.199.110"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "443"),
+				),
+			},
+		},
+	})
+}
+
 const TestAccLbV2MemberConfigBasic = `
 resource "openstack_networking_network_v2" "network_1" {
   name = "network_1"
@@ -141,6 +177,8 @@ resource "openstack_lb_member_v2" "member_1" {
   pool_id = "${openstack_lb_pool_v2.pool_1.id}"
   subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
   weight = 0
+  monitor_address = "192.168.199.110"
+  monitor_port = 8080
 
   timeouts {
     create = "5m"
@@ -154,6 +192,8 @@ resource "openstack_lb_member_v2" "member_2" {
   protocol_port = 8080
   pool_id = "${openstack_lb_pool_v2.pool_1.id}"
   subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
+  monitor_address = "192.168.199.111"
+  monitor_port = 8080
 
   timeouts {
     create = "5m"
@@ -208,6 +248,8 @@ resource "openstack_lb_member_v2" "member_1" {
   admin_state_up = "true"
   pool_id = "${openstack_lb_pool_v2.pool_1.id}"
   subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
+  monitor_address = "192.168.199.110"
+  monitor_port = 8080
 
   timeouts {
     create = "5m"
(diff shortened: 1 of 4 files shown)


flashjmp:
Hi @luigizhou ,
I added these parameters to the lb_members_v2 (which is based on octavia, see here: #1363 )
From my understanding the lb_member_v2 was intentionally kept as a neutron "legacy" version of this?

flashjmp:
I am pretty much done from my end, but there is obviously a review required and the whole project seems to be a little "stale" at the moment. The key issue is that the testing pipeline (#1366) is currently broken... :-(

luigizhou:
@flashjmp I've taken some time to look at the lb_members_v2 resource which is based on octavia as you mentioned.
One thing I noticed is that, compared to member_v2, there is now only a need for a single resource to manage all members of a pool, which looks less flexible and also more disruptive in cases you want to migrate from the old member_v2 to the new members_v2 (defining a members_v2 removes all members of a pool)

Don't you think it would make sense to upgrade the member_v2 resource instead to support octavia api?
Or alternatively defining a new resource that retain the same behaviour as member_v2, but uses octavia api underneath? Something like a 'lb_octavia_member_v2' ? 

I'm happy to work out another PR for it

luigizhou:
@ozerovandrei what do you think?

luigizhou:
Hey @nikParasyr , I should've been able to add support for octavia LB without breaking LBaaS compatibility

luigizhou:
Fixed linting @nikParasyr !

nikParasyr (changes requested):
Hello.

Thank you for your MR. It looks good but there are a couple things missing:
- please update docs of the resource found [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/website/docs/r/lb_member_v2.html.markdown) with the new attributes of the resource
- please update the acceptance tests of the resource found [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_lb_member_v2_test.go) to test the new attributes as well.

luigizhou:
@nikParasyr I've updated the docs, but I'm not entirely sure what version exactly Octavia started supporting monitor_port and address.

I've laso updated the acceptance tests, let me know if those looks good 👍

openstack/resource_openstack_lb_member_v2_test.go
```diff
@@ -96,6 +96,42 @@ func testAccCheckLBV2MemberExists(n string, member *pools.Member) resource.TestC
...
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "8080"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_2", "monitor_address", "192.168.199.111"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "8080"),
+				),
+			},
+			{
+				Config: TestAccLbV2MemberConfigUpdate,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_2", "monitor_address", "192.168.199.110"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "443"),
```
nikParasyr, line 128: `openstack_lb_member_v2.member_1` => needs to be openstack_lb_member_v2.member_2 if im correct.

Test failing with:
```
2022-07-26T11:23:23.6415792Z 2022-07-26T11:23:23.641Z [ERROR] sdk.helper_resource: Unexpected error: test_name=TestAccLBV2Member_monitor test_step_number=2 test_terraform_path=/usr/local/bin/terraform test_working_directory=/tmp/plugintest4229929319 error="Check failed: Check 2/2 error: openstack_lb_member_v2.member_1: Attribute 'monitor_port' expected \"443\", got \"8080\""
2022-07-26T11:23:23.6417153Z     resource_openstack_lb_member_v2_test.go:103: Step 2/2 error: Check failed: Check 2/2 error: openstack_lb_member_v2.member_1: Attribute 'monitor_port' expected "443", got "8080"
```

openstack/resource_openstack_lb_member_v2_test.go
```diff
@@ -96,6 +96,42 @@ func testAccCheckLBV2MemberExists(n string, member *pools.Member) resource.TestC
...
+		Steps: []resource.TestStep{
+			{
+				Config: TestAccLbV2MemberConfigBasic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLBV2MemberExists("openstack_lb_member_v2.member_1", &member1),
+					testAccCheckLBV2MemberExists("openstack_lb_member_v2.member_2", &member2),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_address", "192.168.199.110"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "8080"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_2", "monitor_address", "192.168.199.111"),
+					resource.TestCheckResourceAttr("openstack_lb_member_v2.member_1", "monitor_port", "8080"),
```
nikParasyr, line 121: openstack_lb_member_v2.member_1 => needs to be openstack_lb_member_v2.member_2 if im correct.

openstack/resource_openstack_lb_member_v2_test.go
```diff
@@ -96,6 +96,42 @@ func testAccCheckLBV2MemberExists(n string, member *pools.Member) resource.TestC
...
+			testAccPreCheck(t)
+			testAccPreCheckNonAdminOnly(t)
+			testAccPreCheckLB(t)
+			testAccPreCheckUseOctavia(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckLBV2MemberDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: TestAccLbV2MemberConfigBasic,
```
nikParasyr, line 114: `nit change`
- since you are adding a new test scenario all together can you add a specific config for this scenario as well instead of using the existing `TestAccLbV2MemberConfigBasic`?

openstack/resource_openstack_lb_member_v2_test.go
```diff
@@ -96,6 +96,42 @@ func testAccCheckLBV2MemberExists(n string, member *pools.Member) resource.TestC
 	}
 }
 
+func TestAccLBV2Member_monitor(t *testing.T) {
```
nikParasyr, line 99: `nit change`:
could you also put this before `func testAccCheckLBV2MemberDestroy`? we typically want to have all test scenarios one after the other. Thanks

nikParasyr (changes requested):
Hey, tests are failing and i have some nit fixes as well.

openstack/resource_openstack_lb_member_v2_test.go
luigizhou: this should be fixed

openstack/resource_openstack_lb_member_v2_test.go
luigizhou: This is also fixed now

openstack/resource_openstack_lb_member_v2_test.go
luigizhou: Added a specific config and undone the changes in the basic one

openstack/resource_openstack_lb_member_v2_test.go
luigizhou: Moved the test scenario before the memberdestroy function 👍

luigizhou:
@nikParasyr when you have time can you review again?

openstack/resource_openstack_lb_member_v2_test.go
```diff
@@ -231,3 +269,147 @@ resource "openstack_lb_member_v2" "member_2" {
...
+    delete = "5m"
+  }
+}
+
+resource "openstack_lb_member_v2" "member_2" {
+  address = "192.168.199.111"
+  protocol_port = 8080
+  pool_id = "${openstack_lb_pool_v2.pool_1.id}"
+  subnet_id = "${openstack_networking_subnet_v2.subnet_1.id}"
+  monitor_address = "192.168.199.111
```
nikParasyr, line 333: `"` missing at the end. Tests are failing due to this.

nikParasyr (changes requested):

openstack/resource_openstack_lb_member_v2_test.go
luigizhou: Thanks, I sorted that out 👍

nikParasyr (approved):

PR #1399, opened 2022-07-11
[GH-A] Add workflow for keymanager/barbican

nikParasyr:
Add new workflow for running functional tests for
keymanager/barbican.

Requires: https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1391

### DIFF
--- a/.github/workflows/functional-keymanager.yml
+++ b/.github/workflows/functional-keymanager.yml
@@ -1,0 +1,60 @@
+# Keymanager/barbican tests run only for versions >= wallaby
+# due to barbican's default policies changing completely
+# on wallaby
+name: functional-keymanager
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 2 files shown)


nikParasyr:
@ozerovandrei This is  ready from my side.
There is an important note to be made:
- Im testing only versions after `wallaby` because there was a huge change on barbicans default policies. Basically a big sub-set of our tests requires the admin user on versions before wallaby, and the demo user on versions after wallaby. I decided to just remove the tests on versions < wallaby. We are still testing 4 versions and that should be enough and a big improvement of our openlab setup. Alternatively, I could create a function to skip based on user and version, but i dont think it's rly required atm. If another service has the same issue then ill probably make something like that.

ozerovandrei (approved):
LGTM.
I think 4 versions is enough.

PR #1402, opened 2022-07-12
[GH-A] Add workflow for vpnaas

nikParasyr:


nikParasyr:
@ozerovandrei This is also rdy from my side. 

I had to disable 2 tests that i'm not sure why they are failing. Couldnt figure it out. I would still like it to be merged since we get a job for vpnaas that wasnt tested before so it's an improvement. We could open seperate issues to look into these specific tests.

ozerovandrei (approved):

PR #1403, opened 2022-07-12
[GH-A] Add workflow for fwaas

nikParasyr:
FWaaS was not available after Openstack Ussuri up to Yoga.
Recently it has been revived.


nikParasyr:
@ozerovandrei closing this. FWaaS_v1 has been deprecated since Newton release making it way too old to be testing in my opinion. FWaaS_v2 has been recently revived and if we create resources for it then we can add this workflow

PR #1409, opened 2022-07-14
[GH-A] Remove zuul and openlab configuration

nikParasyr:
Remove all configuration and references for zuul and openlab as they have been replaced by github actions.

Close: https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1381
Close: https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1366


nikParasyr:
@ozerovandrei This can be merged from my side as well.

Some notes:
- the workflow of containerinfra/magnum is still not passing. All the clusters do not reach the ready state and the tests timeout, so does the workflow. This will need quite some work/checks to see what is going on
- the workflow of database/trove is failing on devstack creation for various reasons on each openstack version. Not sure how to properly fix this one and it will also take quite some work to make it happen
- I will look into both of these the coming weeks but ill have a bit less time to do so than this week.
- I will leave it up to you whether to merge this one now or when all workflows are in place

ozerovandrei (approved):

PR #1420, opened 2022-07-28
Update github workflows with concurrency

nikParasyr:
Add concurrency on github worklows that will cancel
running jobs of the same workflow and github-reference.
We often re-push changes before the jobs finish due to
minor fixes, running jobs should be cancelled and new
ones triggered


nikParasyr:
@ozerovandrei This can be merged as well from my side. The 2 failing checks are not related to my changes. One is devstack failing to created the environment, and the other one is a list of strings being returned in different order. For these 2 services all the other checks on the other openstack versions are passing just fine

ozerovandrei (approved):

PR #1423, opened 2022-08-02
Add created/updated to compute instance

MrFreezeex:
Add created/updated attribute to compute instance. This is useful for us
to detecte and change another resource whenever an instance has been
updated.

Signed-off-by: Arthur Outhenin-Chalandre <arthur.outhenin-chalandre@cern.ch>


nikParasyr (approved):
Thank you for the PR. This looks good :)

Issue #1413, opened 2022-07-26
Add github workflow for containerinfra/magnum


nikParasyr:
We have recently moved to github actions( see #1381 ) and most of our services are being tested with their own workflow.
Containerinfra/magnum is failing, all the tests of it.
Need to figure out what the issue is and create a working workflow for it.

kayrus:
@nikParasyr have you tried to rebase the PR and test again?

PR #1428, opened 2022-08-23
User can set role on containerinfra_nodegroup

### LINKED ISSUE #1427: openstack_containerinfra_nodegroup_v1 does not allow to set role
### Terraform Version
```
Terraform v1.2.4
on linux_amd64
+ provider registry.terraform.io/terraform-provider-openstack/openstack v1.48.0
```

### Affected Resource(s)
- openstack_containerinfra_nodegroup_v1

```hcl
resource "openstack_containerinfra_nodegroup_v1" "worker" {
  for_each = { for group in var.worker_nodegroups : group.name => group }

  name           = each.value.name
  cluster_id     = openstack_containerinfra_cluster_v1.main.id
  image_id       = data.openstack_containerinfra_clustertemplate_v1.main.image
  // [...]

  role           = "test"

  // [...]
}
```

### Debug Output
```
╷
│ Error: Value for unconfigurable attribute
│ 
│   with module.cluster.openstack_containerinfra_nodegroup_v1.worker,
│   on ../path/to/file.tf line 32, in resource "openstack_containerinfra_nodegroup_v1" "worker":
│   32:   role           = "test"
│ 
│ Can't configure a value for "role": its value will be decided automatically
│ based on the result of applying this configuration.
```

### Expected Behavior
Terraform created a nodegroup, each node will have the role that is given via `role` attribute

### Actual Behavior
Terraform does not allow to set a role for the nodegroup and stops applying

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
-

### References
-

nikParasyr:
Close #1427

This should allow users to set `role`. Unfortunately the api for nodegroup is not documented but it seems to be the case that setting `role` is allowed based on code and docs linked [here](https://github.com/gophercloud/gophercloud/issues/1767).

@ozerovandrei we currently cannot test this unfortunately due to #1413 but the change is rather small.

### DIFF
--- a/openstack/resource_openstack_containerinfra_nodegroup_v1.go
+++ b/openstack/resource_openstack_containerinfra_nodegroup_v1.go
@@ -91,6 +91,7 @@ func resourceContainerInfraNodeGroupV1() *schema.Resource {
 
 			"role": {
 				Type:     schema.TypeString,
+				Optional: true,
 				ForceNew: true,
 				Computed: true,
 			},


jan-di-cgm:
If it is possible to do a prerelease or something I would be happy to test it in a real environment

nikParasyr:
> If it is possible to do a prerelease or something I would be happy to test it in a real environment

@jan-di-cgm  This would be really helpful as i dont have an environment that supports magnum. We don't need to make a pre-release as you can build the provider locally and use it. Unfortunately it's been a while since i've done it and we have not documented it.

The article [here](https://faun.pub/lets-do-devops-compile-and-test-local-terraform-provider-6f056b69c587) shows how to test locally a provider, it is for azure but the steps more or less are similar:
```
# pull the branch with the changes

# build the provider
go build
# this should give you a binary called terraform-provider-openstack in current directory

# create the required directory
mkdir -p ./.terraform.d/plugins/terraform.example.com/local/openstack/1.48.0/darwin_amd64 <- or different arch if not on mac

# copy the provider in the directory
cp ./terraform-provider-openstack ./.terraform.d/plugins/terraform.example.com/local/openstack/1.48.0/darwin_amd64 

# create a simple main.tf or example.tf with the contents below + the required resources to test
terraform {
  required_version = ">=0.13.0"
  required_providers {
    openstack = {
      version = "~> 2.40.0"
      source  = "terraform.example.com/local/openstack"
    }
  }
}
... and the resources to test

# Run terraform init and plan etc
# make sure when running terraform init to check logs and see that terraform uses the existing provider and does not try to download a different one
# Moreover i think its good to use terraform 0.13 for this one as there were some changes later on that im not sure how they might affect this
```

There might be some troubleshooting required to make it work but the basic steps and ideas are the above. Let me know how it goes

jan-di-cgm:
I've tested it successfully.

```hcl
resource "openstack_containerinfra_nodegroup_v1" "worker" {
  role           = "test"
}
```

The cluster node will get a label
`magnum.openstack.org/role=test`

ozerovandrei (approved):

PR #1436, opened 2022-09-05
Update openstack_containerinfra_nodegroup_v1 documentation.

pawcykca:
Rename 'cluster' to 'node group' in description and add more details about ID used to import nodegroup.


nikParasyr (approved):
@pawcykca thanks for the doc fix. This looks good to me

PR #1406, opened 2022-07-12
[GH-A] Add workflow for containerinfra

nikParasyr:
Close #1413

PR #1437, opened 2022-09-05
Set missing 'docker_volume_size' parameter while reading 'openstack_containerinfra_nodegroup_v1' resource

pawcykca:
Fix nodegroup recreation after import due missing 'docker_volume_size' parameter while reading 'openstack_containerinfra_nodegroup_v1' resource.


nikParasyr (approved):
@pawcykca thanks for the fix. This looks good to me.

This is a simple change so i can merge it, i've seen your other PRs but im hoping to get the magnum ci to work because they are a bit bigger changes and i actually want to test them (see #1406 ).

Once again, thanks :)

PR #1448, opened 2022-10-13
Set terraform version on acceptance tests

nikParasyr:
Acceptance tests for imports where a data source is used are failing
in terraform >= 1.3.X. Pin tf version to 1.2.9 till a fix is introduced
upstream.

RelatesTo: #1449


nikParasyr:
@ozerovandrei I've run the failing tests locally and they pass with this patch.

ozerovandrei (approved):

Issue #1456, opened 2022-11-16
TF Openstack provider deletes security groups rule by rule


AlbertBraden:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see https://www.terraform.io/community.html.

### Terraform Version
Terraform v1.0.1

### Affected Resource(s)
Please list the resources as a list, for example:
security group rules

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
# Instance
resource "openstack_compute_instance_v2" "instance_test" {
  count           = var.instance_count
  name            = "alberttest-terraform-${count.index}"
  image_name      = var.image_name
  flavor_name     = var.flavor_name
  key_pair        = var.key_pair
  security_groups = var.security_groups
variable security_groups {
  default     = [“group1","group2"]
}
```

### Debug Output
Please provide a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. 
I didn't collect debug output because this is a procedural error. When we destroy a TF Openstack cluster, it deletes the security groups rule by rule.

In addition, running Terraform with the `OS_DEBUG=1` environment variable set will print additional information specific to the OpenStack provider which will be helpful for troubleshooting.

Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
Security groups should be deleted without touching the rules. Openstack takes care of the rules.

### Actual Behavior
Rules are deleted first, one at a time. This causes delay and adds load to the Openstack infrastructure. Deleting a security group with 1000 rules via Openstack CLI takes 3 seconds. Deleting 1000 rules via TF takes about 30 minutes. Increasing parallelism doesn't help; the solution is to not delete the rules.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Build Openstack cluster via TF
2. terraform destroy

### Important Factoids
Is there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? A custom version of OpenStack? Tight ACLs?

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:

taquy:
[...]

PR #1458, opened 2022-11-29
Implement rule for openstack_networking_secgroup_v2

baurmatt:
This PR implements a new `rule` argument for the `openstack_networking_secgroup_v2` resource. Most of this was copied from the deprecated `openstack_compute_secgroup_v2` resource.

# The problem

We're already able to deploy security group rules with the help of the [`openstack_networking_secgroup_rule_v2`](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/networking_secgroup_rule_v2). But this comes with problems when defining a bunch of security group rules.

There are, as far as I known, 2 ways to iterate over resources:

## count

```terraform
terraform {
  required_providers {
    openstack = {
      source = "terraform-provider-openstack/openstack"
    }
  }
}

locals {
  allow_ssh_server = ["10.0.0.1/32", "10.0.0.2/32"]
  rules = [for ip in local.allow_ssh_server : {
    direction        = "ingress"
    ethertype        = "IPv4"
    protocol         = "tcp"
    port_range_min   = 22,
    port_range_max   = 22,
    remote_ip_prefix = ip
    }
  ]
}

resource "openstack_networking_secgroup_v2" "secgroup_1" {
  name                 = "secgroup_1"
  delete_default_rules = true
}

resource "openstack_networking_secgroup_rule_v2" "secgroup_rule_1" {
  count             = length(local.rules)
  direction         = local.rules[count.index]["direction"]
  ethertype         = local.rules[count.index]["ethertype"]
  protocol          = local.rules[count.index]["protocol"]
  port_range_min    = local.rules[count.index]["port_range_min"]
  port_range_max    = local.rules[count.index]["port_range_max"]
  remote_ip_prefix  = local.rules[count.index]["remote_ip_prefix"]
  security_group_id = openstack_networking_secgroup_v2.secgroup_1.id
}
```

## for_each

```terraform
terraform {
  required_providers {
    openstack = {
      source = "terraform-provider-openstack/openstack"
    }
  }
}

locals {
  allow_ssh_server = ["10.0.0.1/32", "10.0.0.2/32"]
  rules = {for ip in local.allow_ssh_server : ip => {
    direction        = "ingress"
    ethertype        = "IPv4"
    protocol         = "tcp"
    port_range_min   = 22,
    port_range_max   = 22,
    remote_ip_prefix = ip
    }
  }
}

resource "openstack_networking_secgroup_v2" "secgroup_1" {
  name                 = "secgroup_1"
  delete_default_rules = true
}

resource "openstack_networking_secgroup_rule_v2" "secgroup_rule_1" {
  for_each          = local.rules
  direction         = each.value.direction
  ethertype         = each.value.ethertype
  protocol          = each.value.protocol
  port_range_min    = each.value.port_range_min
  port_range_max    = each.value.port_range_max
  remote_ip_prefix  = each.value.remote_ip_prefix
  security_group_id = openstack_networking_secgroup_v2.secgroup_1.id
}
```

Both options have disadvantages:

* count
  * Changing the order of the IPs recreates security group rules because the resource key changes
* for_each
  * Requires a key in a map that isn't used anywhere else because a security group rule doesn't have a name/title.

# Solution

This new approach brings the advantage of allowing a list to be the input and doesn't come with the drawback of securiy group recreation.

```terraform
terraform {
  required_providers {
    openstack = {
      source = "terraform-provider-openstack/openstack"
    }
  }
}

locals {
  allow_ssh_server = ["10.0.0.1/32", "10.0.0.2/32"]
  rules = [for ip in local.allow_ssh_server : {
    direction        = "ingress"
    ethertype        = "IPv4"
    protocol         = "tcp"
    port_range_min   = 22,
    port_range_max   = 22,
    remote_ip_prefix = ip
    }
  ]
}

resource "openstack_networking_secgroup_v2" "secgroup_1" {
  name = "secgroup_1"

  dynamic "rule" {
    for_each = local.rules
    content {
      direction         = rule.value["direction"]
      ethertype         = rule.value["ethertype"]
      protocol          = rule.value["protocol"]
      port_range_min    = rule.value["port_range_min"]
      port_range_max    = rule.value["port_range_max"]
      remote_ip_prefix  = rule.value["remote_ip_prefix"]
    }
  }
}
```

For us this is important when using a generic security group (rule) module because it allows us to specify a variable of type `list(object)`:

```terraform
variable "rules" {
  type = list(object({
    direction        = optional(string, "ingress")
    ethertype        = optional(string, "IPv4")
    description      = optional(string)
    protocol         = optional(string)
    port_range_min   = optional(number)
    port_range_max   = optional(number)
    remote_ip_prefix = optional(string)
    tenant_id        = optional(string)
    remote_group_id  = optional(string)
  }))
  default = []

  validation {
    condition = alltrue([
      for r in var.rules : contains(["ingress", "egress"], r.direction)
    ])
    error_message = "All rules must have 'direction' of either 'ingress' or 'egress'."
  }
  # ...
}
```

With this users of that module don't have to bother about an irrelevant key for their security group rules.

### DIFF
--- a/openstack/resource_openstack_networking_secgroup_v2_test.go
+++ b/openstack/resource_openstack_networking_secgroup_v2_test.go
@@ -84,6 +84,28 @@ func TestAccNetworkingV2SecGroup_timeout(t *testing.T) {
 	})
 }
 
+func TestAccNetworkingV2SecGroup_rule(t *testing.T) {
+	var securityGroup groups.SecGroup
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			//testAccPreCheckNonAdminOnly(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckNetworkingV2SecGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNetworkingV2SecGroupRule,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckNetworkingV2SecGroupExists("openstack_networking_secgroup_v2.secgroup_1", &securityGroup),
+					testAccCheckNetworkingV2SecGroupRuleCount(&securityGroup, 2),
+				),
+			},
+		},
+	})
+}
+
 func testAccCheckNetworkingV2SecGroupDestroy(s *terraform.State) error {
 	config := testAccProvider.Meta().(*Config)
 	networkingClient, err := config.NetworkingV2Client(osRegionName)
@@ -177,6 +199,30 @@ resource "openstack_networking_secgroup_v2" "secgroup_1" {
 
   timeouts {
     delete = "5m"
+  }
+}
+`
+const testAccNetworkingV2SecGroupRule = `
+resource "openstack_networking_secgroup_v2" "secgroup_1" {
+  name = "security_group"
+	delete_default_rules = true
+
+  rule {
+    port_range_min   = 22
+    port_range_max   = 22
+    protocol         = "tcp"
+	  ethertype        = "IPv4"
+	  direction        = "ingress"
+    remote_ip_prefix = "0.0.0.0/0"
+  }
+
+  rule {
+    port_range_min   = 80
+    port_range_max   = 80
+    protocol         = "tcp"
+	  ethertype        = "IPv4"
+	  direction        = "ingress"
+    remote_ip_prefix = "0.0.0.0/0"
   }
 }
 `
--- a/openstack/resource_openstack_networking_secgroup_v2_test.go
+++ b/openstack/resource_openstack_networking_secgroup_v2_test.go
@@ -211,8 +211,8 @@ resource "openstack_networking_secgroup_v2" "secgroup_1" {
     port_range_min   = 22
     port_range_max   = 22
     protocol         = "tcp"
-	  ethertype        = "IPv4"
-	  direction        = "ingress"
+    ethertype        = "IPv4"
+    direction        = "ingress"
     remote_ip_prefix = "0.0.0.0/0"
   }
 
@@ -220,8 +220,8 @@ resource "openstack_networking_secgroup_v2" "secgroup_1" {
     port_range_min   = 80
     port_range_max   = 80
     protocol         = "tcp"
-	  ethertype        = "IPv4"
-	  direction        = "ingress"
+    ethertype        = "IPv4"
+    direction        = "ingress"
     remote_ip_prefix = "0.0.0.0/0"
   }
 }
--- a/openstack/networking_secgroup_v2.go
+++ b/openstack/networking_secgroup_v2.go
@@ -1,12 +1,18 @@
 package openstack
 
 import (
+	"bytes"
+	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
+	"github.com/gophercloud/utils/terraform/hashcode"
 )
 
 // networkingSecgroupV2StateRefreshFuncDelete returns a special case resource.StateRefreshFunc to try to delete a secgroup.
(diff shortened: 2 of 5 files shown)


openstack/resource_openstack_networking_secgroup_v2_test.go
```diff
@@ -180,3 +202,27 @@ resource "openstack_networking_secgroup_v2" "secgroup_1" {
...
+const testAccNetworkingV2SecGroupRule = `
+resource "openstack_networking_secgroup_v2" "secgroup_1" {
+  name = "security_group"
+	delete_default_rules = true
+
+  rule {
+    port_range_min   = 22
+    port_range_max   = 22
+    protocol         = "tcp"
+	  ethertype        = "IPv4"
```
nikParasyr, line 214: fix indentation plz

openstack/resource_openstack_networking_secgroup_v2_test.go
```diff
@@ -180,3 +202,27 @@ resource "openstack_networking_secgroup_v2" "secgroup_1" {
...
+	  direction        = "ingress"
+    remote_ip_prefix = "0.0.0.0/0"
+  }
+
+  rule {
+    port_range_min   = 80
+    port_range_max   = 80
+    protocol         = "tcp"
+	  ethertype        = "IPv4"
```
nikParasyr, line 223: indentation here as well

nikParasyr (changes requested):
Couple of indentation changes please.

Other than that, it is ok from my side. Although i have to mention that i'm not a big fan of managing multiple openstack resources from a single terraform resource. Makes the code more complicated than it should be and less maintainable. It also goes against terraform best practises. That being said i will not block it, but i will leave the final call to @ozerovandrei. This PR should also address the issue mentioned in #1456 as the delete will wipe the security group which automatically will delete all the rules.

baurmatt:
@nikParasyr Thanks for the review! :) I've fix the indention. About https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1456, that's correctly implemented for `openstack_networking_secgroup_v2` - it will just deletes the full security group, leaving the rules to OpenStack.

baurmatt:
Errr, I can't explain the CI failures and are unable to locate the actual error in the CI logs. I don't think it's related, as I've only changed the indention of some Terraform code -> https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1458/commits/c44799e5cabd62cf1e5d850470bb51a72f224bd1

nikParasyr (approved):

ozerovandrei (changes requested):
> Although i have to mention that i'm not a big fan of managing multiple openstack resources from a single terraform resource. Makes the code more complicated than it should be and less maintainable. It also goes against terraform best practises.

@baurmatt Could you please provide more context about why we need to implement security group rules inside `openstack_networking_secgroup_v2` resource? I have the same concerns as @nikParasyr here.
 
And how it differs from the existing [resource_openstack_networking_secgroup_rule_v2.go](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_networking_secgroup_rule_v2.go)?

baurmatt:
@ozerovandrei I should have probably done this when creating this PR, sorry! I've updated the description with an explanation of the motivation behind this MR. Hope that helps! :)

nikParasyr:
@baurmatt thanks for the explanation. I’ve had the same issue and the way I resolved it is to have a list as the input of the module and then with the use of locals I build a map. The key of each item is the merge of ‘direction-protocol-min_port-max_port-…’ which should be unique. This allows me to use for_each and the benefits of it instead of count, while users don’t have to bother about passing/giving keys. Although it has some shortcomings as the generation of the map is a bit hard to understand.

baurmatt:
@nikParasyr You're right! This worked out pretty easy!

```terraform
locals {
  _rules_keys = [for rule in var.rules : "${coalesce(rule.direction, "x")}-${coalesce(rule.ethertype, "x")}-${coalesce(rule.description, "x")}-${coalesce(rule.protocol, "x")}-${coalesce(rule.port_range_min, "x")}-${coalesce(rule.port_range_max, "x")}-${coalesce(rule.remote_ip_prefix, "x")}-${coalesce(rule.remote_group_id, "x")}"]
  _rules      = { for rule in var.rules : local._rules_keys[index(var.rules, rule)] => rule }
}

resource "openstack_networking_secgroup_v2" "secgroup" {
  name                 = var.name
  description          = var.description
  delete_default_rules = true
}

resource "openstack_networking_secgroup_rule_v2" "secgroup_rule" {
  for_each          = local._rules
  security_group_id = openstack_networking_secgroup_v2.secgroup.id
  direction         = each.value.direction
  ethertype         = each.value.ethertype
  description       = each.value.description
  protocol          = each.value.protocol
  port_range_min    = each.value.port_range_min
  port_range_max    = try(coalesce(each.value.port_range_max, each.value.port_range_min), null)
  remote_ip_prefix  = each.value.remote_ip_prefix
  tenant_id         = each.value.tenant_id
  remote_group_id   = each.value.remote_group_id == "this" ? openstack_networking_secgroup_v2.secgroup.id : each.value.remote_group_id
}
```

I remember that I've tried this before, but ended up with so much validation code that it was insane to maintain. This seems to have gotten way better with the latest updates of our module to list(object), optional attributes and validation!

Would there be any interest to build community driven modules where we can share those best practices? Can't really find good ones on the registry.

nikParasyr:
Good to hear that it works for you. Recent tf features have made this much easier than it used to be.

> Would there be any interest to build community driven modules where we can share those best practices? 

Not sure if there is interest for modules and maintaining them. Maybe we could update `examples/` with a set of more realistic modules/deployments. We could also consider to enable `Discussions` which could be more for questions/best-practices etc. Not sure what @ozerovandrei thinks about it.

In regards with this MR, do you still need the functionality it includes? I'm still against it for the reasons mentioned above, and since you can achieve what you want easily with modules I don't think it's really needed.

baurmatt:
I'm fine closing this PR. Thanks for your help! :)

baurmatt:
I have to reopen this as I've found another usecase for the implementation of this PR.

The suggested solution (https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1458#issuecomment-1346311175) above doesn't work if keys derived from resource attributes e.g. reference to an floating IP resources.

```terraform
terraform {
  required_providers {
    openstack = {
      source = "terraform-provider-openstack/openstack"
    }
  }
}

resource "openstack_networking_floatingip_v2" "fip" {
  pool = "ext-net"
}

module "security-group" {
  source = "/Users/mbaur/Sources/terraform-modules/openstack-security_group"
  name   = "security-group"
  rules = [
    {
      "protocol"         = "tcp"
      "port_range_min"   = "9200"
      "remote_ip_prefix" = "${openstack_networking_floatingip_v2.fip.address}/32"
    },
  ]
}
```

(The openstack-security_group module contains, more or less, the code I've posted in https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1458#issuecomment-1346311175)

The leads to the following error:

```
module.security-group.openstack_networking_secgroup_v2.secgroup: Refreshing state... [id=9ff0720a-1388-4494-80a1-e08013a25831]
╷
│ Error: Invalid for_each argument
│
│   on .terraform/modules/security-group/main.tf line 13, in resource "openstack_networking_secgroup_rule_v2" "secgroup_rule":
│   13:   for_each          = local._rules
│     ├────────────────
│     │ local._rules will be known only after apply
│
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will identify
│ the instances of this resource.
│
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place apply-time results only in the map values.
│
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends on, and then apply a second time to fully converge.
```

This works just fine for the implementation I suggest in this PR.

rossengeorgiev:
One thing I found frustrating about `openstack_networking_secgroup_v2` is that you can't deal with configuration drift. If someone adds rules outside of terraform, then terraform remains oblivious to it. This creates the following problems:

* configuration dift is undetected
* rules added outside of terraform would prevent `openstack_networking_secgroup_v2` from being destroyed
* rules added outside of terraform would collide when the exact same rule is added via terraform

This PR could potentially solve that. Alternatively, we `openstack_networking_secgroup_v2` could be implemented to manage both the security group and rules, while `openstack_networking_secgroup_rule_v2` would be like `null_resource` or `terraform_data`. Then a list of rules would be pasted to a `rules` attribute `openstack_networking_secgroup_v2`, which would then perform all the necessary API calls to make it so.

baurmatt:
Hey @nikParasyr and @ozerovandrei, what do you think about merging/releasing this PR? :)

nikParasyr:
Hello @baurmatt. Tbh, I’m still against it. I understand the reasoning and the use-cases behind it but this is an anti-pattern and the provider is “thinly”maintained atm.

PR #1462, opened 2022-12-15
Facilitate providing a list of tags in data_source_openstack_images_image_v2

RickyGrassmuck:
The current implementation of `data_source_openstack_images_image_v20` only allows for setting a single `tag` attribute while the API is able to accept a list of tags to filter on.

This change allows users to directly set the `tags` attribute to a list of tags in their data source block while maintaining backwards compatibility with the existing `tag` attribute. If both are provided, the singular `tag` is simply added to the `tags` attribute before being added to the `ListOpts`.

Examples

**Original Syntax**
```hcl
data "openstack_images_image_v2" "image_1" {
  tag   = "cirros-tf_1"
}
```

**New Syntax with `tags` attribute**
```hcl
data "openstack_images_image_v2" "ubuntu" {
  tags = ["cirros-tf_1"]
}
```
**New Syntax using both `tag` and `tags` attribute**
```hcl
data "openstack_images_image_v2" "ubuntu" {
  tag   = "cirros-tf_1"
  tags = ["cirros-tf_1"]
}
```

I've added an additional acceptance test as well to account for being able to specify the `tags` attribute in the data source.

### DIFF
--- a/openstack/data_source_openstack_images_image_v2.go
+++ b/openstack/data_source_openstack_images_image_v2.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"sort"
 	"time"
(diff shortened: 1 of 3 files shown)


nikParasyr (changes requested):
@RickyGrassmuck Thanks for the PR. It looks good to me, just the minor linting error to fix

RickyGrassmuck (commented):
Suggested changes added

PR #1468, opened 2022-12-22
Allow providing a list of tags in data_source_openstack_images_image_ids_v2

### LINKED ISSUE #1467: [FR] data.openstack_images_image allow a list of tags to be provided
### Terraform Version
`Terraform v1.0.11`

### Affected Resource(s)
- openstack_images_image_ids_v2
- openstack_images_image_v2

Below is example config for specifying multiple tags. Not currently valid.
The current way is with parameter named `tag`, which takes only a string, and not list. Thus multiple tags cannot be specified.

```hcl
data "openstack_images_image_v2" "image" {
  visibility  = "public"
  tags         = ["tag1", "tag2"]
  most_recent = true
}
```

### Expected Behavior
Able to specific a list of tags, and an image matching all those tags would be resolved.

### Actual Behavior
Can only specify a single tag. If I want to find an image that matches 2 or more tags, there is no way of doing that.

rossengeorgiev:
Ports change from #1462 for `openstack_images_image_ids_v2`  
Fixes #1467

Allows filtering using a list a tags by setting the `tags` attribute, while maintaining backwards compatibility with `tag` attribute.

Original Syntax

```hcl
data "openstack_images_image_ids_v2" "image_1" {
  tag   = "cirros-tf_1"
}
```

New Syntax with tags attribute

```hcl
data "openstack_images_image_ids_v2" "image_1" {
  tags = ["cirros-tf_1"]
}
````

New Syntax using both tag and tags attribute

```hcl
data "openstack_images_image_ids_v2" "image_1" {
  tag   = "cirros-tf_1"
  tags = ["cirros-tf_1"]
}
```

### DIFF
--- a/openstack/data_source_openstack_images_image_ids_v2.go
+++ b/openstack/data_source_openstack_images_image_ids_v2.go
@@ -122,6 +122,13 @@ func dataSourceImagesImageIDsV2() *schema.Resource {
 				ConflictsWith: []string{"name"},
 			},
 
+			"tags": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+
 			// Computed values
 			"ids": {
 				Type:     schema.TypeList,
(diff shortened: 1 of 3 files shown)


nikParasyr (approved):
LGTM. thank you

rossengeorgiev:
@nikParasyr thanks. Stupid question, how do build and use the provider locally?

nikParasyr:
> @nikParasyr thanks. Stupid question, how do build and use the provider locally?

this comment should be helpful => https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1428#issuecomment-1224004468

PR #1469, opened 2022-12-23
Add name_regex attribute to data_source_openstack_images_image_v2

rossengeorgiev:
Noticed that `data_source_openstack_images_image_ids_v2` has `name_regex` to filter images, while `data_source_openstack_images_image_v2` doesn't. This PR adds it, and works in the same way.

**Before:**

Exact name matching only

```hcl
data "openstack_images_image_ids_v2" "images" {
  name = "Ubuntu_16.04_cloudimage_amd64"
 }
```

**After:**

Partial name matching possible

```hcl
data "openstack_images_image_ids_v2" "images" {
  name_regex = "^Ubuntu_16\\.04.*_amd64$"
 }
```

### DIFF
--- a/openstack/data_source_openstack_images_image_v2.go
+++ b/openstack/data_source_openstack_images_image_v2.go
@@ -27,9 +27,10 @@ func dataSourceImagesImageV2() *schema.Resource {
 			},
 
 			"name": {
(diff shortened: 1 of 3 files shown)


kayrus (approved):
LGTM, thanks

PR #1482, opened 2023-02-16
Add decompression flag to openstack_images_image_v2 resource

### LINKED ISSUE #1481: openstack_images_image_v2 resource not uncompressing dowloaded image
### Affected Resource(s)
- openstack_images_image_v2

### Expected Behavior
- Defining an image resource with an url pointing to a compressed image (for example flatcar ones which have well defined "content-type" headers)
- Uncompress automatically
- Upload uncompressed image

See:
```
$ curl -I "https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_openstack_image.img.bz2"
HTTP/2 200 
server: nginx
date: Wed, 15 Feb 2023 21:56:28 GMT
content-type: application/x-bzip2
content-length: 371091603
etag: "rosr8b64xs43"
last-modified: Fri, 20 Jan 2023 18:26:35 GMT
accept-ranges: bytes

$ curl -I "https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_openstack_image.img.gz"
HTTP/2 200 server: nginx
date: Wed, 15 Feb 2023 21:58:46 GMT
content-type: application/gzip
content-length: 388364110
etag: "rosr8b6f7zny"
last-modified: Fri, 20 Jan 2023 18:26:35 GMT
accept-ranges: bytes
```

### Actual Behavior
Images are uploaded without errors but in their compressed format (tried it with gzip and bz2) which is unusable.

### Steps to Reproduce
Trying to use a compressed image in a  openstack_images_image_v2 resource.

spnngl:
Some image are compressed and only filling Content-Type header resulting if uploading a compressed image to openstack.

With this we automatically detects and uncompress images in this case.

Example:
```
$ curl -I "https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_openstack_image.img.bz2"
HTTP/2 200
server: nginx
date: Thu, 16 Feb 2023 15:54:03 GMT
content-type: application/x-bzip2
content-length: 371535602
etag: "rq4xkz657ape"
last-modified: Wed, 15 Feb 2023 18:48:35 GMT
accept-ranges: bytes
```

Resolves #1481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ FEATURES
 * Updated Golangci-lint to `v1.51.2` ([#1488](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1488))
 * Updated Terraform SDK to `v2.25.0` ([#1490](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1490))
 
+FEATURES
+
 IMPROVEMENTS
 
+* Added `decompress` argument to `images_image_v2` resource ([#1482](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1482))
 * Added `name_regex` argument to `openstack_images_image_v2` data source ([#1469](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1469))
 * Added ability to provide a list of tags into `openstack_images_image_v2`, `openstack_images_image_ids_v2` data sources ([#1462](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1462)), ([#1468](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1468))
 * Added `name_regex` attribute to `openstack_images_image_v2` data source ([#1469](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1469))
@@ -16,6 +19,8 @@ IMPROVEMENTS
 BUG FIXES
 
 * Fixed `containerinfra_clustertemplate_v1` labels update ([#1455](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1455))
+
+BUG FIXES
 
 ## 1.49.0 (26 October, 2022)
 
@@ -473,8 +478,8 @@ IMPROVEMENTS
 
 BUG FIXES
 
-* Fixed the bug with unchecked errors in initialization of Identity V3 client in `identity_auth_scope_v3` data source [[#878](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/878)] 
-* Fixed the bug with unchecked errors in initialization of Compute V2 client in `compute_floatingip_associate_v2` resource [[#878](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/878)] 
+* Fixed the bug with unchecked errors in initialization of Identity V3 client in `identity_auth_scope_v3` data source [[#878](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/878)]
+* Fixed the bug with unchecked errors in initialization of Compute V2 client in `compute_floatingip_associate_v2` resource [[#878](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/878)]
 * Fixed the bug with 404 errors handling while getting statuses tree in `openstack_lb_loadbalancer_v2` resource ([#883](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/883))
 * Fixed the bug where is was unable to remove TLS references in `openstack_lb_listener_v2` resource ([#891](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/891))
 * Fixed the bug where empty `scheduler_hints` list caused a panic in `openstack_compute_instance_v2` resource ([#885](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/885))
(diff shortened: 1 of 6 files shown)


openstack/images_image_v2.go
```diff
@@ -127,7 +129,30 @@ func resourceImagesImageV2File(client *gophercloud.ServiceClient, d *schema.Reso
 
 			defer resp.Body.Close()
 
-			if _, err = io.Copy(file, resp.Body); err != nil {
+			var reader io.ReadCloser
+			decompress := d.Get("decompress").(bool)
+
+			if decompress && !resp.Uncompressed {
```
kayrus, line 135: checking for `resp.Uncompressed` doesn't make sense, since it is used only for `Content-Encoding`

openstack/images_image_v2.go
```diff
@@ -127,7 +129,30 @@ func resourceImagesImageV2File(client *gophercloud.ServiceClient, d *schema.Reso
...
+				// we'll read "Content-Type" to select format
+				switch resp.Header.Get("Content-Type") {
+				case "gzip", "application/gzip":
+					reader, err = gzip.NewReader(resp.Body)
+					if err != nil {
+						return "", fmt.Errorf("Error decompressing gzip image: %s", err)
+					}
+				case "bzip2", "application/bzip2", "application/x-bzip2":
+					bz2_reader := bzip2.NewReader(resp.Body)
+					reader = io.NopCloser(bz2_reader)
```
kayrus, line 146: Let's follow the golang var name standards.
```suggestion
					bz2Reader := bzip2.NewReader(resp.Body)
					reader = io.NopCloser(bz2Reader)
```

openstack/images_image_v2.go
```diff
@@ -127,7 +129,30 @@ func resourceImagesImageV2File(client *gophercloud.ServiceClient, d *schema.Reso
...
+				default:
+					return "", fmt.Errorf("Error decompressing image, format %s is not supported", resp.Header.Get("Content-Type"))
+				}
+				defer reader.Close()
+			} else {
+				reader = resp.Body
+			}
```
kayrus, line 153: you can avoid `else` by initially setting the reader to `resp.Body` on the 132 line:

```
reader := resp.Body
```

openstack/resource_openstack_images_image_v2.go
```diff
@@ -176,6 +176,13 @@ func resourceImagesImageV2() *schema.Resource {
 				ConflictsWith: []string{"local_file_path", "verify_checksum"},
 			},
 
+			"decompress": {
```
kayrus, line 179: can you mention this flag in documentation?

kayrus (changes requested):
thanks for your PR. I left a couple of comments.

kayrus:
In addition please update the `CHANGELOG.md`:

```
## 1.50.0 (Not released yet)

FEATURES

IMPROVEMENTS

* ...

```

openstack/images_image_v2.go
spnngl: I wanted to check if we're not trying to decompress something already decompressed by golang stack or which was never compressed to begin with

openstack/resource_openstack_images_image_v2.go
spnngl: done

spnngl:
Did the modifications in the last version

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.50.0 (17 February, 2023)
```
kayrus, line 1:
```suggestion
## 1.50.0 (Not released yet)
```

website/docs/r/images_image_v2.html.markdown
```diff
@@ -97,6 +97,11 @@ The following arguments are supported:
     be used to let Openstack download the image directly from the remote source.
     Conflicts with `local_file_path`. Defaults to false.
 
+* `decompress` - (Optional) If true, this provider will decompress downloaded
+    image before uploading it to Openstack. Decompression algorithm is chosen by
```
kayrus, line 101:
```suggestion
    image before uploading it to OpenStack. Decompression algorithm is chosen by
```

kayrus (commented):
Thanks. LGTM. Please fix the comments. And I think it's ready to be merged.

website/docs/r/images_image_v2.html.markdown
spnngl: done

CHANGELOG.md
spnngl: done

spnngl:
Updated

kayrus (approved):

kayrus:
@spnngl can you fix golang linter issues?

kayrus:
It also looks like OpenStack test env doesn't support web-download anymore. Can you add an exception and skip these tests for a while?

spnngl:
Pushed a new version with lint fix + web_download tests deactivated

maxime1907:
Any update? We really need this to be merged!

kayrus:
I re-triggered glance tests, once they pass, the PR can be merged.

kayrus:
can you add a `"decompress"` exception into the import tests? Tests fail because of
```
ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

(map[string]string) {
}


(map[string]string) (len=1) {
 (string) (len=10) "decompress": (string) (len=5) "false"
}
```

https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/bd2071b1a27dfbe07f08797c3acd7410ff0a61c4/openstack/import_openstack_images_image_v2_test.go#L29-L33

spnngl:
I added the decompress ignore + rebase on main

openstack/resource_openstack_images_image_v2.go
```diff
@@ -176,6 +176,13 @@ func resourceImagesImageV2() *schema.Resource {
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: false,
```
kayrus, line 182: I think it makes sense to set `ForceNew` to `true`. Objections?

spnngl: You're right, I will make the modifications

spnngl: Pushed it

nikParasyr:
Since `changelog` is updated this will trigger all the ci jobs. 

I think recently opendev,org added some limits to pulling repos etc and as we also have our daily check failing with a timeout. I need to have a look to make sure that the timeout is due to limits and think of a way to bypass it.

kayrus:
@nikParasyr @ozerovandrei  I was not aware about the behavior when changelog is changed. I think we can skip failed `functional-images` test, it has some issues with the data source. other tests are fine.
I think we can merge this PR.

nikParasyr:
@kayrus the idea behind this behavior with changelog is that we update it only before a release (or after multiple MRs are changed) and the whole ci runs to make sure all services are ok. But it can be a bit annoying if changelog is updated on each MR (also often we have conflicts due to that).

From my side this can be merged as well. The failing ci jobs are not related to this change.

nikParasyr (approved):

PR #1497, opened 2023-02-28
Use github mirrors instead of opendev

### LINKED ISSUE #1496: CI jobs failing due to opendev introducing limits
Many ci jobs are failing due to opendev.orf introducing limits to pulling repos etc.
Scheduled runs are failing as well as MR's that trigger multiple jobs or if we have multiple MRs at the same time.

Need to change to pull from github mirrors instead of opendev

nikParasyr:
Update devstack-action to 0.10 that uses github mirrors instead of opendev that has introduced limits on pulling repos. Moreover update links to any plugin so that it's also pulled from github mirror and not opendev

Close #1496

### DIFF
--- a/.github/workflows/functional-blockstorage_v2.yml
+++ b/.github/workflows/functional-blockstorage_v2.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout TPO
         uses: actions/checkout@v3
       - name: Deploy devstack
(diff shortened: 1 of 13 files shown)


nikParasyr:
@ozerovandrei could have have a look?

Sorry for not tackling it earlier but i was on vacation for a month

ozerovandrei (approved):

PR #1505, opened 2023-03-02
keymanager v1: fix payload retrieve for binary data

### LINKED ISSUE #1492: Always changes in tf plan using resource "openstack_keymanager_secret_v1" when payload_content_type = "application/octet-stream"
Hello all.

We are trying to use Terraform to manage Openstack Barbican secrets for Openstack LBAAS and would like to only upload new secrets when the cert/payload is changed. However, when running tf plan with an unchanged certificate file and payload_content_type is set to "application/octet-stream", a change is forced. Setting an ignore on payload attribute is not an acceptable workaround since we must detect when certificate files are changed. We started seeing this issue when trying to pull certs using  Hashicorp Vault data source, but this issue also occurs even just using a local file.

This issue does not occur when  payload_content_type is set to "text/plain". It looks like maybe the secret is not read back from Openstack properly, so a new payload is generated.

### Terraform Version
Terraform v1.3.9
on linux_amd64
+ provider registry.terraform.io/terraform-provider-openstack/openstack v1.49.0
(Openstack version is Zed)

### Affected Resource(s)
- openstack_keymanager_secret_v1

```hcl
terraform {
  required_version = ">= 1.3.0"
  required_providers {
    openstack = {
      source = "terraform-provider-openstack/openstack"
    }
  }
}
resource "openstack_keymanager_secret_v1" "certificate_1" {
  name                 = "certificate"
  payload_content_type = "application/octet-stream"
  payload_content_encoding = "base64"
  payload              = filebase64("my_cert.p12")
  secret_type          = "certificate"
}
```

### Debug Output
https://gist.github.com/lcurtis-datto/737f2a029529da1f3d07810f7dee9577

### Expected Behavior
Running `terraform plan` should have resulted in no changes

### Actual Behavior
Output of "openstack_keymanager_secret_v1.certificate_1 must be replaced"
      + payload                  = (sensitive value) # forces replacement

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`

### Important Factoids

Looking at the debug output, it seems that the payload_content_type is not being set when trying to read the secret. I can produce similar error using openstack-cli and not specifying payload content-type.

EX: 
`openstack secret get https://<REDACTED>:9311/v1/secrets/11f1812b-8b67-4157-83de-d8a4cc8652ba -p
4xx Client error: Not Found: Secret not found.
Not Found: Secret not found.`

https://gist.github.com/lcurtis-datto/737f2a029529da1f3d07810f7dee9577#file-plan-out-L219

kayrus:
Fixes #1492 

Now payload state is stored in base64 format in case when it's not the text. In these cases it's responsibility of end user to detect the encoding/content-type and call https://developer.hashicorp.com/terraform/language/functions/base64decode

Payloads with whitespaces and newlines are also supported now, e.g. passwords that have whitespace at the end.

### DIFF
--- a/openstack/data_source_openstack_keymanager_secret_v1.go
+++ b/openstack/data_source_openstack_keymanager_secret_v1.go
@@ -233,7 +233,7 @@ func dataSourceKeyManagerSecretV1Read(ctx context.Context, d *schema.ResourceDat
 	payloadContentType := secret.ContentTypes["default"]
 	d.Set("payload_content_type", payloadContentType)
 
-	d.Set("payload", keyManagerSecretV1GetPayload(kmClient, d.Id()))
+	d.Set("payload", keyManagerSecretV1GetPayload(kmClient, d.Id(), payloadContentType))
 	metadataMap, err := secrets.GetMetadata(kmClient, d.Id()).Extract()
 	if err != nil {
 		log.Printf("[DEBUG] Unable to get %s secret metadata: %s", uuid, err)
--- a/openstack/keymanager_secret_v1.go
+++ b/openstack/keymanager_secret_v1.go
@@ -97,32 +97,18 @@ func keyManagerSecretMetadataV1WaitForSecretMetadataCreation(kmClient *gopherclo
 	}
 }
 
-func keyManagerSecretV1GetPayload(kmClient *gophercloud.ServiceClient, id string) string {
-	payload, err := secrets.GetPayload(kmClient, id, nil).Extract()
+func keyManagerSecretV1GetPayload(kmClient *gophercloud.ServiceClient, id, contentType string) string {
+	opts := secrets.GetPayloadOpts{
+		PayloadContentType: contentType,
+	}
+	payload, err := secrets.GetPayload(kmClient, id, opts).Extract()
 	if err != nil {
 		log.Printf("[DEBUG] Could not retrieve payload for secret with id %s: %s", id, err)
 	}
-	return string(payload)
-}
 
-func resourceSecretV1PayloadBase64CustomizeDiff(diff *schema.ResourceDiff) error {
-	encoding := diff.Get("payload_content_encoding").(string)
-	if diff.Id() != "" && diff.HasChange("payload") && encoding == "base64" {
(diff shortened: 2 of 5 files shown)


kayrus:
That's wired. My local tests:

```
checking payload: "p3\xb2q\x10E\x9e\x1e" vs "cDOycRBFnh4=" decoded("p3\xb2q\x10E\x9e\x1e")
```

differ from CI tests:

```
checking payload: "p3�q\x10E�\x1e" vs "cDOycRBFnh4=" decoded("p3\xb2q\x10E\x9e\x1e")
```

with extra debug added:

```go
func resourceSecretV1PayloadBase64CustomizeDiff(diff *schema.ResourceDiff) error {
        encoding := diff.Get("payload_content_encoding").(string)
        if diff.Id() != "" && diff.HasChange("payload") && encoding == "base64" {
                o, n := diff.GetChange("payload")
                oldPayload := o.(string)
                newPayload := n.(string)

                v, err := base64.StdEncoding.DecodeString(newPayload)
                if err != nil {
                        return fmt.Errorf("The Payload is not in the defined base64 format: %s", err)
                }
                newPayloadDecoded := string(v)

                log.Printf("[DEBUG] checking payload: %q vs %q decoded(%q)", oldPayload, newPayload, newPayloadDecoded)
                if oldPayload == newPayloadDecoded {
                        log.Printf("[DEBUG] payload has not changed. clearing diff")
                        return diff.Clear("payload")
                }
        }

        return nil
}
```

kayrus:
I decided to store binary data in base64 to avoid these issues.

kayrus:
@ozerovandrei @nikParasyr please review
@lcurtis-datto can you test this PR in your environment?

lcurtis-datto:
This works! Thanks a million for fixing this so quickly.

LGTM

kayrus:
TODO: I need to ensure that provider upgrade will not mess up with base64 payload.

nikParasyr (approved):
LGTM. is it ready from your side @kayrus? Cause i saw some conversation going on on the related issue

kayrus:
yes, it's ready.

PR #1514, opened 2023-03-13
[compute v2]: prefer cinder V3 API over V2 on import

### LINKED ISSUE #1425: resource_openstack_compute_instance_v2 prefers blockstorage v2 on imports
### Terraform Version
1.2.6

### Affected Resource(s)
- resource_openstack_compute_instance_v2

### Expected Behavior
Since the Blockstorage V2 API is deprecated or even removed on newer OpenStack releases

> Version 2 of the Block Storage API was [deprecated in the Pike release](https://docs.openstack.org/releasenotes/cinder/pike.html#deprecation-notes) and was removed during the Xena development cycle.
( see https://docs.openstack.org/api-ref/block-storage/v2/) 

it should only be used if there is no V3 available.

### Actual Behavior
When importing a `openstack_compute_instance_v2` an blockstorage_v2 client is preferred and v3 is only used if that fails. See
https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/e1906e4bece4a6e081387034c269145ee808df6e/openstack/resource_openstack_compute_instance_v2.go#L1231

while this works, it should actually be the other way around.

kayrus:
Fixes #1425

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -1228,7 +1228,7 @@ func resourceOpenStackComputeInstanceV2ImportState(ctx context.Context, d *schem
 
 	bds := []map[string]interface{}{}
 	if len(serverWithAttachments.VolumesAttached) > 0 {
-		blockStorageClient, err := config.BlockStorageV2Client(GetRegion(d, config))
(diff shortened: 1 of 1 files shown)


nikParasyr (approved):
LGTM

PR #1518, opened 2023-03-15
Update CHANGELOG.md

kayrus:
Preparing for a new release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+## 1.51.0 (15 March, 2023)
+
+NOTES
+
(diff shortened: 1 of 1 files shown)


nikParasyr:
> then we can include the https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1386 as well

Check comment in that PR. If you want this change to be included then tests need to be updated etc before merging it.

> As for https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1517, I asked the author. IMHO the gophercloud/utils is not ready for the system scope, since https://github.com/gophercloud/utils/blob/769528992f29ab711cc6fea99146e433f76eb4a5/terraform/auth/config.go#L19 was not yet updated accordingly. The feature will work only with ENV variable or clouds.yaml, but not via the provider configuration arguments.

ACK. is that being worked on gophercloud or should we make an issue for it?
I guess it would take some time to be implemented so we can go forward with the release without waiting for it.

---

From my side we can make a release without the PRs mentioned above.

nikParasyr (approved):

PR #1532, opened 2023-03-24
Add ci job for unit tests

nikParasyr:
Rename all unit tests to contain `Unit` and add a ci job that runs all unit tests.


nikParasyr:
@kayrus can you have a look at this? The blockstorage failing job is a cinder change that im working to fix on #1531. I think this can be merged

Issue #1533, opened 2023-03-27
[Doc] Facilitate documentation browsing through subcategory usage


yomovh:
The goal of this feature request (that I can implement) is to use of the [subcategory](https://developer.hashicorp.com/terraform/registry/providers/docs#yaml-frontmatter) feature in the YAML frontmatter. 
This should smooth the documentation browsing. 

I propose we use the pattern `Category  / Project code name`. This would be :
* Block Storage / Cinder
* Compute VM / Nova
* Container Infra / Magnum
* Database / Trove
* DNS / Designate
* Networking / Neutron
* Identity / Keystone
* Images / Glance
* Secrets / Barbican
* Load Balancing as a Service / Octavia
* Object Storage / Swift
* Orchestration / Heat
* Shared Filesystem / Manila

Let me know if seems ok for you and I will implement the PR.

nikParasyr:
This is something i was thinking myself as well as the number of resources and data sources have increased.
From my side I would be happy to see a PR making this change. @kayrus ?

PR #1540, opened 2023-03-31
Add subcategory to better group documentation 

yomovh:
As discussed in #1533


website/docs/r/compute_volume_attach_v2.html.markdown
```diff
@@ -1,4 +1,5 @@
 ---
+subcategory: "Compute VM / Nova"
```
kayrus, line 2: Nova can spawn baremetal entities as well, so I'd rename this to `Compute / Nova"`

kayrus (changes requested):
Thanks for the PR. See my comments.

yomovh:
Feedbacks taken into account

yomovh:
@kayrus : can you tell me if this is ok for you? Thanks !

nikParasyr (approved):

Issue #1057, opened 2020-09-09
Wrong client microversion (X-Openstack-Nova-Api-Version) for volume type with multiattach option enabled


pawcykca:
### Terraform Version
```
Terraform v0.13.2
+ provider registry.terraform.io/terraform-providers/openstack v1.31.0
```

### Affected Resource(s)
- openstack_compute_instance_v2

### Terraform Configuration Files
```hcl
resource "openstack_blockstorage_volume_v3" "bastion01_vol01" {
  name = "bastion01_vol1"
  size = "20"
  image_id = var.open_stack_image_id
  enable_online_resize = true
  volume_type = "gp-ceph-nolimit"
  availability_zone = "az1-sse6"
}

resource "openstack_compute_instance_v2" "bastion01" {
  name            = "bastion01"
  flavor_id       = var.open_stack_flavor_id
  key_pair        = "czk-devops"
  security_groups = ["allow_all_out","allow_ssh_in_from_ca","allow_icmp"]
  availability_zone = "az1-sse6"

  block_device {
    uuid                  = openstack_blockstorage_volume_v3.bastion01_vol01.id
    source_type           = "volume"
    boot_index            = 0
    destination_type      = "volume"
    delete_on_termination = false
  }

  network {
    name = "cloud_testnet"
  }
}
```

### Debug Output
https://gist.github.com/pawcykca/65bfcf964a6e2b19125ce7e0f91c6e7f

### Expected Behavior
What should have happened?

The Terraform Openstack Provider should have realized that the volume type of volume defined in `block_device` section has  multiattach option enabled and choose correct microversion for compute endpoint (set `X-Openstack-Nova-Api-Version` header).

or

There should be a ability do define microversion for selected service endpoints.

### Actual Behavior
What actually happened?

The Terraform Openstack Provider sends the default microsversion for the compute endpoint (does not set the `X-Openstack-Nova-Api-Version` header) and the operation fails.
```
openstack_compute_instance_v2.bastion01: Creating...

Error: Error creating OpenStack server: Bad request with: [POST https://my-cloud:13774/v2.1/servers], error message: {"badRequest": {"code": 400, "message": "Multiattach volumes are only supported starting with compute API version 2.60."}}

  on instances.tf line 13, in resource "openstack_compute_instance_v2" "bastion01":
  13: resource "openstack_compute_instance_v2" "bastion01" {
```

### Steps to Reproduce
[...]

PR #1542, opened 2023-04-09
[compute v2] Add multiattach flag to instance block_device (#1057)

ralgar:
### Issue

Please see #1057 for detailed description.

It is not currently possible to use multiattach volumes with the `block_device` block of compute instances, as it requires compute microversion `>=2.60`. This is a major show stopper for my use case, where I need to attach such a volume to an instance *before* it boots.

### Proposal

~~The provider *already* uses microversion `2.67` when the `volume_type` parameter is set in the `block_device` block. Rather than adding more arbitrary checks for these edge cases, I propose we stick to a KISS philosophy by removing this check altogether, and simply pin to this "BlockDevice" microversion any time the `block_device` block is used.~~

As per a maintainer's suggestion below, I have added a `multiattach` flag to the `block_device` block, which sets the compute microversion to the required `2.60`. This is similar to how other resources already handle multiattach volumes, allowing for an easy way to utilize them, while maintaining backwards compatibility with older versions of OpenStack.

I've also made note of the added flag in the relevant documentation page.

### DIFF
--- a/openstack/compute_instance_v2.go
+++ b/openstack/compute_instance_v2.go
@@ -28,6 +28,7 @@ const (
 	computeV2InstanceCreateServerWithTagsMicroversion        = "2.52"
 	computeV2TagsExtensionMicroversion                       = "2.26"
 	computeV2InstanceBlockDeviceVolumeTypeMicroversion       = "2.67"
+	computeV2InstanceBlockDeviceMultiattachMicroversion      = "2.60"
 )
 
 // InstanceNIC is a structured representation of a Gophercloud servers.Server
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -284,6 +284,12 @@ func resourceComputeInstanceV2() *schema.Resource {
 							Optional: true,
 							ForceNew: true,
 						},
+						"multiattach": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+							ForceNew: true,
+						},
 					},
 				},
 			},
@@ -552,6 +558,14 @@ func resourceComputeInstanceV2Create(ctx context.Context, d *schema.ResourceData
 		blockDevices, err := resourceInstanceBlockDevicesV2(d, vL.([]interface{}))
 		if err != nil {
 			return diag.FromErr(err)
+		}
+
+		// Check if Multiattach was set in any of the Block Devices.
+		// If so, set the client's microversion appropriately.
+		for _, bd := range d.Get("block_device").([]interface{}) {
+			if bd.(map[string]interface{})["multiattach"].(bool) {
+				computeClient.Microversion = computeV2InstanceBlockDeviceMultiattachMicroversion
+			}
 		}
 
 		// Check if VolumeType was set in any of the Block Devices.
--- a/website/docs/r/compute_instance_v2.html.markdown
+++ b/website/docs/r/compute_instance_v2.html.markdown
@@ -479,6 +479,9 @@ The `block_device` block supports:
 * `disk_bus` - (Optional) The low-level disk bus that will be used. Most common
     thing is to leave this empty. Changing this creates a new server.
 
+* `multiattach` - (Optional) Enable the attachment of multiattach-capable
+    volumes.
+
 The `scheduler_hints` block supports:
 
 * `group` - (Optional) A UUID of a Server Group. The instance will be placed


ralgar:
Thanks for the feedback @nikParasyr, that makes sense.

To clarify, what you'd like to see would be a boolean flag `multiattach`? And you would prefer it to pin specifically to microversion 2.60, rather than using the same microversion 2.67 already used by the `volume_type` flag?

nikParasyr:
@ralgar 
> To clarify, what you'd like to see would be a boolean flag multiattach?

Yes. inside the `block_device`. This should be defaulted to false, but when enabled the compute microversion should be set correctly.

> Also, am I correct that this option would need to be implemented in Gophercloud first?

No. This will not be propagated to the request.Opts. It's just a flag to allow us to understand that we need to bumb the microversion correctly.


@kayrus what is your opinion? I dont like my suggestion in all honesty, but i cant find a better way to achieve this given the microversion labyrinth...

ralgar:
@nikParasyr 10-4.

For what it's worth, your suggestion is in line with other volume related resources. For example, both `openstack_blockstorage_volume_v3`, and `openstack_compute_volume_attach_v2` have a `multiattach` flag.

I think it's a decent suggestion but, yes, let's wait for more input.

ralgar:
Hey, @nikParasyr. This didn't get any movement over the last month, so I've gone and implemented your original suggestion. As I mentioned previously, this appears to be how other resources are already handling multiattach volumes, so it seems like a sensible change.

All works well on my end, and I've updated the docs page to reflect the change. Let me know what you think. Thanks.

ralgar:
Hey @kayrus, let me try to clarify things as I probably didn't provide enough info here.

This PR is for the `openstack_compute_instance_v2` resource, which currently does *not* appear to have any support for multiattach volumes. This is *not* a boot volume, it is a data volume.

As this compute instance is running Fedora CoreOS, the volume needs to be attached to the instance before it is started, so that it can be customized and mounted during the boot process. FCOS will fail to boot if the volume is not available at this time. The `openstack_compute_volume_attach_v2` resource is unable to fulfill this requirement as the instance is started before the volume attachment even begins.

2.60 does appear to be a requirement for multiattach, as it is explicitly mentioned in the error message:

```
╷
│ Error: Error creating OpenStack server: Bad request with: [POST http://xx.x.x.xx:8774/v2.1/servers], error message: {"badRequest": {"code": 400, "message": "Multiattach volumes are only supported starting with compute API version 2.60."}}
│
│   with module.fcos_server.openstack_compute_instance_v2.fcos,
│   on ../../modules/openstack/fcos_server/compute.tf line 1, in resource "openstack_compute_instance_v2" "fcos":
│    1: resource "openstack_compute_instance_v2" "fcos" {
│
╵
```

The problem is solved by setting the microversion, as seen in this PR.

Hopefully this all makes sense now :)

kayrus (approved):
> This PR is for the `openstack_compute_instance_v2` resource

Thanks a lot for the clarification. Please don't get me wrong, I'm not against the change, I'm here to avoid overcomplicating the fragile `openstack_compute_instance_v2` resource, which is already overloaded with exceptions, and unfortunately some of them are not clearly defined, so it's not obvious to see them while reading the code. As you can see there is already a `2.60` microversion assignment in another resource, which, due to legacy reasons, isn't aligned with constants definition.

What I'm afraid of is that at some point someone will add an extra microversion, which isn't compatible with API that works with a lower microversion, or another microversion setting will downgrade one defined earlier. @nikParasyr I think we need to open a new issue and develop a helper func that ensures that microversions don't override each other.

> As this compute instance is running Fedora CoreOS, the volume needs to be attached to the instance before it is started, so that it can be customized and mounted during the boot process. FCOS will fail to boot if the volume is not available at this time. The openstack_compute_volume_attach_v2 resource is unable to fulfill this requirement as the instance is started before the volume attachment even begins.

I assume you're using ignition that prepares a VM with attached block devices, and it runs only during the bootstrap.
According to my experience, most of the problems can be solved by the correct use of existing tools (e.g. udev rules may be used in your case). But in your case (and according to [this change](https://github.com/openstack/nova/commit/7e6ae9afd9f6f6bb1bf4289c598fbc95dfd52612), nova API reference note wasn't clear enough) **it looks reasonable to set a microversion while creating a compute instance**.

@ralgar have you tested this change in your env? I have some doubts about whether DevStack supports multiattach.

ralgar:
@kayrus No problem. I completely understand your concerns, as I had the same thoughts when I was implementing this. A helper function would certainly be a good idea going forward. I would have liked to implement such a thing here but, as I am not too familiar with the nuances of the code base, I thought it best to keep things simple.

>I assume you're using ignition that prepares a VM with attached block devices, and it runs only during the bootstrap.
According to my experience, most of the problems can be solved by the correct use of existing tools (e.g. udev rules may be used in your case).

That's a correct assumption. I was aware that udev has some potentially useful functionality here, however I'm not sure how that would fit into the FCOS bootstrap sequence, and I can't seem to find any documentation or examples for doing such a thing. If you happen to have more information, or a functional example, I'd certainly be interested.

>@ralgar have you tested this change in your env? I have some doubts about whether DevStack supports multiattach.

I believe you are correct about DevStack. AFAIK, multiattach is always opt-in for any OpenStack / Cinder environment. I *have* tested the change in my env, and everything appears to work as expected. With the flag set to `false`, or unset (so defaulting to `false`), it behaves as before - producing the error message I posted above. With the flag set to `true` it successfully provisions the instance with the multiattach volume.

nikParasyr (approved):

nikParasyr:
@ralgar thanks for your work and sorry for the delays :)

ralgar:
No problem at all. Thank you both for your hard work as well! It really shows in the excellent quality of the provider :)

@nikParasyr @kayrus

Issue #1485, opened 2023-02-21
Changing any non-name argument values unnecesarily forces cluster rebuild


colans:
### Terraform Version

1.49.0 (Latest version)

### Affected Resource(s)

* `openstack_containerinfra_cluster_v1`

### Problem description

When looking at the documentation, all of the arguments (except for `name`) state:
> Changing this creates a new cluster.

This shouldn't be necessary for minor changes.  

### Expected Behavior

Minor changes shouldn't force a rebuild of the cluster.

### Actual Behavior

Minor changes force a rebuild of the cluster.

### Steps to Reproduce

For example, it should be possible to change `node_count` from `1` to `3` without rebuilding the entire cluster.  But this is what will happen if you run `terraform apply`.

It should add the two new nodes and enable them without having to rebuild everything.

mnaser:
https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_containerinfra_cluster_v1.go#L165

it shouldn't force a new if node count changes, is Terraform perhaps giving you some other reason that's forcing the change?   If you can share the diff, that would be helpful

colans:
Maybe that was a bad example then.  I didn't actually try that particular one; I was simply going by the [documentation](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/containerinfra_cluster_v1).  Here are the ones we tried, IIRC (note all of the TODOs):

```terraform
resource "openstack_containerinfra_cluster_v1" "c2_development" {
  name                = "c2_development"
  # ref. https://dashboard.vexxhost.net/project/cluster_templates
  cluster_template_id = "cdf4cfc8-2dce-4671-8639-a66f1a325e5c"
  # TODO: This change forces cluster replacement. Update before next cluster migration.
  master_count        = 1
  #master_count        = 3
  master_flavor       = "v3-starter-2"
  node_count          = 2
  # TODO: This change forces cluster replacement. Update before next cluster migration.
  flavor              = "v3-starter-2"
  #flavor              = "v3-starter-4"
  keypair             = "christopher-keypair"
  # TODO: This change forces cluster replacement. Update before next cluster migration.
  # merge_labels doesn't work, so we need to reproduce the labels from the template here.
  #merge_labels        = true
  labels              = {
    # Disable PodSecurityPolicy
[...]

PR #1550, opened 2023-04-20
docs(containerinfra): update node_count

mnaser:
The documentation states that changing node_count will result in the recreation of the cluster which is not valid, this patch updates the docs to match the behaviour that it will not create a new cluster.

Related: #1485

### DIFF
--- a/website/docs/r/containerinfra_cluster_v1.html.markdown
+++ b/website/docs/r/containerinfra_cluster_v1.html.markdown
@@ -37,8 +37,8 @@ The following arguments are supported:
     the `region` argument of the provider is used. Changing this creates a new
     cluster.
 
-* `name` - (Required) The name of the cluster. Changing this updates the name
-    of the existing cluster template.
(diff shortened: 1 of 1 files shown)


website/docs/r/containerinfra_cluster_v1.html.markdown
```diff
@@ -37,8 +37,8 @@ The following arguments are supported:
     the `region` argument of the provider is used. Changing this creates a new
     cluster.
 
-* `name` - (Required) The name of the cluster. Changing this updates the name
-    of the existing cluster template.
+* `name` - (Required) The name of the cluster. Changing this creates a new
+    cluster.
```
colans, line 41: So the name change also initiates a new cluster?  Have you tested this?  That's disappointing.

mnaser: The API doesn't allow for changing of name clusters, so we really gotta bring this up to the Magnum level.

nikParasyr (approved):

PR #1551, opened 2023-04-20
fix(containerinfra): replace nodegroup if cluster_id changes

mnaser:
If the cluster attached to a node group changed, that node group
cannot be "re-attached", it should just be fully rebuilt.


nikParasyr:
@mnaser has this been tested or just based on docs? cause unfortunately we dont have a ci for magnum

mnaser:
@nikParasyr I did not test it, no, sorry, I can try to find sometime to validate this.

nikParasyr (approved):

PR #1567, opened 2023-05-21
Create new identity_inherit_role_assignment_v3 resource

### LINKED ISSUE #1372: Inheritable role assignment using OS-INHERIT Identity API extension
Hi,

[OpenStack Identity API v3 OS-INHERIT Extension](https://opendev.org/openstack-attic/identity-api/src/branch/master/v3/src/markdown/identity-api-v3-os-inherit-ext.md) makes possible to assign inheritable permissions on project to an user/group. It would be nice to be able to use this feature with this provider, either with a `inherit` boolean in `openstack_identity_role_assignment_v3` or a new resource (e.g. `openstack_osinherit_identity_role_assignment_v3`).

### Terraform Version

v1.0.11

### Affected Resource(s)

- openstack_identity_role_assignment_v3

```hcl
resource "openstack_identity_project_v3" "parent_project" {
  name = "parent_project"
}

resource "openstack_identity_project_v3" "child_project" {
  name = "child_project"
  parent_id = resource.openstack_identity_project_v3.parent_project.id
}

resource "openstack_identity_user_v3" "user_1" {
  name = "user_1"
  default_project_id = "${openstack_identity_project_v3.parent_project.id}"
}

resource "openstack_identity_role_v3" "role_1" {
  name = "role_1"
}
```

### Behavior proposal

The following tf file should create a role assignment using the OS-INHERIT API extension, thus making it inheritable to child projects. user_1 should then have access to both parent_project and child_project.

```hcl
resource "openstack_identity_role_assignment_v3" "role_assignment_1" {
  user_id = "${openstack_identity_user_v3.user_1.id}"
  project_id = "${openstack_identity_project_v3.parent_project.id}"
  role_id = "${openstack_identity_role_v3.role_1.id}"
  inherit = true
}
```

nikParasyr:
This creates new identity resource for inherited roles using the OS-INHERIT keystone API. Chose to create a new resource instead of adding this functionality to existing role assignment to:
- keep the resources clean and easy
- this is a seperate api, thus i prefer a new resource to manage it
- adding it to existing assignment would require updating the resource id and could impact users

Close #1372

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/terraform-provider-openstack/terraform-provider-openstack
 go 1.20
 
 require (
-	github.com/gophercloud/gophercloud v1.2.1-0.20230309142102-36ac4a411ba7
+	github.com/gophercloud/gophercloud v1.3.1-0.20230517135124-33412becb1d0
 	github.com/gophercloud/utils v0.0.0-20230324070755-05e9e7f5ea4d
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.25.0
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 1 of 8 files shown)


kayrus (approved):
Thanks for the PR. The only drawbacks I see are duplicate code and documentation. The risk of human error is quite high, especially when someone makes a fix to a similar piece of code, as the duplicate part tends to be overlooked. In terms of code, this doesn't seem to be a major issue, as we can create additional functions that share the logic. However, when it comes to documentation, the only solution seems to be automatic documentation generation based on the source code. I believe we are not yet prepared to implement the autodocs, so I approve the PR.

PR #1571, opened 2023-05-23
Add openstack_identity_registered_limit_v3 resource

### LINKED ISSUE #1569: New resource for identity v3 registered limits
New resource should be created to support identity v3 registered limits under unified limits.

Upstream work should be finished with: https://github.com/gophercloud/gophercloud/pull/2616

nikParasyr:
Add openstack_identity_registered_limit_v3 resource, import test and documentation. Registered Limit resource creates only 1 limit, instead of multiple as the api allows. The reasoning is that a single limit leads to much easier code to understand and maintain, while multiple limits would create issues in handling ids, updates etc.

Close #1569


openstack/resource_openstack_identity_registered_limit_v3.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if d.HasChange("service_id") {
+		hasChange = true
+		updateOpts.ServiceID = d.Get("service_id").(string)
+	}
+
+	if d.HasChange("resource_name") {
+		hasChange = true
+		updateOpts.ResourceName = d.Get("service_id").(string)
```
kayrus, line 128:
```suggestion
		updateOpts.ResourceName = d.Get("resource_name").(string)
```

kayrus (changes requested):
> I decided to add one limit per resource instead of multiple for simplicity. You think this is ok?

yes, I think this is reasonable.

> For Update the api docs mention....

Unfortunately I have nothing to comment, I don't have an env to test this API behavior right now.

I'd start with the following local test cases:
* limit already exists and you're trying to create a new one
* you're creating two identical limit resources in terraform and see how would it behave
* remove the existing limit and see what would happen on next apply
* what will happen if you set an empty `resource_name` on update?
* can you change just a `description` for the limit?

nikParasyr:
hi @sxmatch. I dont have time to work on this currently.

Issue I had back then was to enable system scope on the ci tests.

If you want to pick this up, that would be very nice

sxmatch:
hi @nikParasyr  Thanks for the information, will continue this work recently.

kayrus (approved):
LGTM

I prefer this PR over #1607.

PR #1572, opened 2023-05-23
Wait for Orchestration Stack to update

nmiletic:
Wait for UPDATE_COMPLETE in resourceOrchestrationStackV1Update. This change fixes the issue where provider does not honor  terraform parallelism option and just pushes all the stack updates at the same time causing overload.

Following test have passed: TestAccOrchestrationV1Stack_update, TestAccOrchestrationV1Stack_basic


nikParasyr (approved):
LGTM. thanks

PR #1576, opened 2023-06-01
Use `multierror` in `Read` operation in `r/compute_instance_v2`

rgyrbu:
Minor changes in `r/compute_instance_v2`

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha1"
 	"encoding/hex"
 	"fmt"
+	"github.com/hashicorp/go-multierror"
 	"log"
 	"os"
 	"strings"
@@ -59,16 +60,16 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Computed: true,
 				ForceNew: true,
 			},
-
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: false,
 			},
 			"image_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validation.IsUUID,
 			},
 			"image_name": {
 				Type:     schema.TypeString,
@@ -148,10 +149,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"uuid": {
-							Type:     schema.TypeString,
-							Optional: true,
-							ForceNew: true,
-							Computed: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							ForceNew:     true,
+							Computed:     true,
+							ValidateFunc: validation.IsUUID,
 						},
 						"name": {
 							Type:     schema.TypeString,
@@ -166,16 +168,18 @@ func resourceComputeInstanceV2() *schema.Resource {
 							Computed: true,
 						},
 						"fixed_ip_v4": {
-							Type:     schema.TypeString,
-							Optional: true,
-							ForceNew: true,
-							Computed: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							ForceNew:     true,
+							Computed:     true,
+							ValidateFunc: validation.IsIPv4Address,
 						},
 						"fixed_ip_v6": {
-							Type:     schema.TypeString,
-							Optional: true,
-							ForceNew: true,
-							Computed: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							ForceNew:     true,
+							Computed:     true,
+							ValidateFunc: validation.IsIPv6Address,
 						},
 						"floating_ip": {
 							Type:       schema.TypeString,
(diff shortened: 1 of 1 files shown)


rgyrbu:
@nikParasyr hope you can review this :)

nikParasyr (commented):
This sort of looks good to me but i have 2 reservations:
- The compute resource is quite bloated and brittle. Any change to it can potentially break things in ways that the ci doesnt cover (unfortunately)
- `multierror` seems quite useful but i think it's something you either standardize across your code or dont use. I dont think that just adding it in one resource is really beneficial.

@kayrus what is your opinion?

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -59,16 +60,16 @@ func resourceComputeInstanceV2() *schema.Resource {
...
 				ForceNew: false,
 			},
 			"image_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validation.IsUUID,
```
kayrus, line 72: Last time I implemented extra validate checks, I unintentionally caused a bug :)
Is there a chance that an image_id can have a non-uuid format like flavors?

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -231,17 +237,22 @@ func resourceComputeInstanceV2() *schema.Resource {
 			"block_device": {
 				Type:     schema.TypeList,
 				Optional: true,
+				ForceNew: true,
```
kayrus, line 240: This is also a suspicious change. I have a gut feeling that this may cause issues for existing resources.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1402,28 +1413,28 @@ func resourceInstanceBlockDevicesV2(_ *schema.ResourceData, bds []interface{}) (
 }
 
 func resourceInstanceSchedulerHintsV2(_ *schema.ResourceData, schedulerHintsRaw map[string]interface{}) schedulerhints.SchedulerHints {
-	differentHost := []string{}
+	var differentHost []string
```
kayrus, line 1416: once you started this initiative, does it make sense to allocate the slice with the `len(v)` and assign a value instead of appending it to a slice? :)

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1484,7 +1495,7 @@ func getImageIDFromConfig(imageClient *gophercloud.ServiceClient, d *schema.Reso
 		return imageID, nil
 	}
 
-	return "", fmt.Errorf("Neither a boot device, image ID, or image name were able to be determined")
+	return "", fmt.Errorf("neither a boot device, image ID, or image name were able to be determined")
```
kayrus, line 1498: See https://preply.com/en/blog/using-either-neither-nor-and-or-in-english/

```suggestion
	return "", fmt.Errorf("neither a boot device, image ID, nor image name could be determined")
```

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1551,7 +1562,7 @@ func getFlavorID(computeClient *gophercloud.ServiceClient, d *schema.ResourceDat
 		return flavorID, nil
 	}
 
-	return "", fmt.Errorf("Neither a flavor_id or flavor_name could be determined")
+	return "", fmt.Errorf("neither a flavor_id or flavor_name could be determined")
```
kayrus, line 1565:
```suggestion
	return "", fmt.Errorf("neither a flavor ID nor a flavor name could be determined")
```

kayrus (commented):
Overall looks good, but I'm against extra changes in the state struct definition. I have a gut feeling that we may end up in a yet another minor release fix :\

openstack/resource_openstack_compute_instance_v2.go
rgyrbu: changes in error messages were just fixing IDE warnings :D

openstack/resource_openstack_compute_instance_v2.go
rgyrbu: right

openstack/resource_openstack_compute_instance_v2.go
rgyrbu: hm, here I disagree. It should be there from my pov, because we don't have `update` operation for `block_device`

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -795,6 +812,10 @@ func resourceComputeInstanceV2Read(_ context.Context, d *schema.ResourceData, me
 		computeV2InstanceReadTags(d, instanceTags)
 	}
 
+	if err := mErr.ErrorOrNil(); err != nil {
+		return diag.FromErr(err)
+	}
+
 	return nil
```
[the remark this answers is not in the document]
rgyrbu: done

openstack/resource_openstack_compute_instance_v2.go
rgyrbu: heh, you are right :)

openstack/resource_openstack_compute_instance_v2.go
rgyrbu: done

rgyrbu:
> This sort of looks good to me but i have 2 reservations:
> 
> * The compute resource is quite bloated and brittle. Any change to it can potentially break things in ways that the ci doesnt cover (unfortunately)
> * `multierror` seems quite useful but i think it's something you either standardize across your code or dont use. I dont think that just adding it in one resource is really beneficial.
> 
> @kayrus what is your opinion?

you are right, I hope that I will have time and I will continue here

openstack/resource_openstack_compute_instance_v2.go
kayrus: It's not about the update, it's about how we store this data from API response into the state. If the API format is changed, then we can have troubles.

rgyrbu: right, removed

rgyrbu:
Guys, thanks for the review, but I can't wait that long. My main account was unban and I will return there, so I will not answer here anymore, but I am deleting the account.

PR #1583, opened 2023-06-19
[manila v2]: introduce a new share access rules read method

### LINKED ISSUE #1582: Manila: POST to list share access rules is deprecated
Hi,

with microversion [2.45 in manila listing share access rules](https://docs.openstack.org/api-ref/shared-file-system/?expanded=list-share-access-rules-detail#list-share-access-rules) can be done with the new `GET /v2/share-access-rules` endpoint, the old [pre 2.45 ](https://docs.openstack.org/api-ref/shared-file-system/?expanded=list-access-rules-deprecated-detail#list-access-rules-deprecated) `POST ../action` endpoint is deprecated.

The new endpoint is already available in [gophercloud](https://github.com/gophercloud/gophercloud/tree/master/openstack/sharedfilesystems/v2/shareaccessrules).

This makes categorising/auditing/rate-limiting easier if a call that is listing/reading and not creating/updating is done with a GET instead of a POST.

Thanks in advance,
Maurice

kayrus:
Resolves #1582

### DIFF
--- a/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
+++ b/openstack/resource_openstack_sharedfilesystem_share_access_v2.go
@@ -13,8 +13,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/apiversions"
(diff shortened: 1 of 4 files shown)


nikParasyr (approved):
@kayrus LGTM. you need to rebase due to conflict. But i cna merge after you do

kayrus:
@nikParasyr I noticed that POST method was used in the access rules refresh func. I reworked the PR with a cleaner code.

PR #1585, opened 2023-06-21
Remove unnecessary string substitutions in examples

strugee:


strugee:
Immediately after filing this I realized that probably a bunch of other files had this problem too, so I'm uploading a new patch that fixes this (almost) everywhere. To generate this, I:

1. Ran `grep -R '"${[^{]*}"$'` and manually looked through the entire results list to find instances that may be problematic for a basic `sed` script. That resulted in the `grep` call below, which targets one particular instance that just seems a little tricky and happens to have `base64encode` in it as well as [`.*` splat operators](https://developer.hashicorp.com/terraform/language/v1.1.x/upgrade-guides/0-12#working-with-count-on-resources) which are better done in a separate patch
2. Ran `sed -Ei 's/"\$\{([^{]*)\}"$/\1/' $(grep -R '"${[^{]*}"$' | cut -d: -f1 | sort | uniq)` to do the syntax fixup
3. Ran `git checkout -p $(grep -lRFe '.*' -e 'base64encode')` and discarded hunks with the problematic syntax mentioned in step 1/manually edited these files to fix them up
4. Did a final manual check with `nano $(grep -lRFe '"${')`

Note that I did not really look at the diff here since I think in order for that to be even vaguely useful I need some time not staring at this thing. I _did_ vaguely sanity-check some parts of it after the `sed` command but it was just a glance - nothing thorough.

kayrus (approved):
LGTM.

>  .* splat operators

are you going to submit a new PR or do you want to add an extra commit here?

nikParasyr (approved):

strugee:
I can submit a new one

Issue #115, opened 2017-09-14
feature request: support for FWaaS v2.0


Crapworks:
Hi there,

my OpenStack provider only supports FWaaS v2.0, the v1.0 endpoint is disabled. Are there any plans already to add support for FWaaS v2.0 since v1.0 is flagged as deprecated?

Cheers,
Christian

fatmcgav:
@Crapworks I've had a quick look, and currently Gophercloud doesn't support FWaaS v2. 

Therefore support will need to be added to Gophercloud before we can add support into TF.

Crapworks:
@fatmcgav Got it, I've created an issue there: https://github.com/gophercloud/gophercloud/issues/514

jtopjian:
@Crapworks Just a heads up that this might take a number of weeks/months to put in place since the entirety of FWaaS v2 will have to be added to Gophercloud first. That's not to say it won't ever happen, but I just wanted to give you a realistic expectation.

Crapworks:
@jtopjian Thanks for the estimation. I justed wanted it to be on the list, since I haven't seen anyone requesting it yet. Now I have something to track it :)

khdegraaf:
FYI, I have mostly working code in my private branch for gophercloud at https://github.com/khdegraaf/gophercloud, and the matching code for Terraform in https://github.com/khdegraaf/terraform-provider-huaweicloud.  Note this is currently only targeting older production Huawei cloud variants like OTC, but I also plan to do PRs for the official OpenStack variants once I am comfortable with its status and have some time to do so.  See my comments on the GopherCloud feature request on one issue I am currently stuck on.  Although I think the code may work if you are running current versions of OpenStack.  Any help or pointers on how to properly create an "ACTIVE" router port I can attach to a firewall group on OTC would be appreciated.

khdegraaf:
[...]

PR #1588, opened 2023-06-27
[FWaaS_V2]: Add FW Group V2 resource

Koodt:
Add resourceFWGroupV2 CRUD
Add resourceFWGroupV2 tests
Add resourceFWGroupV2 docs
Add import_openstack_fw_group_v2_test
Enable openstack_fw_group_v2 in provider.go

Close part of #115

### DIFF
--- a/openstack/fw_group_v2.go
+++ b/openstack/fw_group_v2.go
@@ -15,9 +15,7 @@ import (
 
 func fwGroupV2RefreshFunc(networkingClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
 	return func() (interface{}, string, error) {
-		var group groups.Group
-
-		err := groups.Get(networkingClient, id).ExtractIntoStructPtr(&group, "firewall_group")
+		group, err := groups.Get(networkingClient, id).Extract()
 		if err != nil {
 			return nil, "", err
 		}
--- a/openstack/fw_group_v2.go
+++ b/openstack/fw_group_v2.go
@@ -1,0 +1,88 @@
+package openstack
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas_v2/groups"
+)
+
+func fwGroupV2RefreshFunc(networkingClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		var group groups.Group
+
+		err := groups.Get(networkingClient, id).ExtractIntoStructPtr(&group, "firewall_group")
+		if err != nil {
+			return nil, "", err
+		}
+
+		return group, group.Status, nil
+	}
+}
+
+func fwGroupV2DeleteFunc(networkingClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		group, err := groups.Get(networkingClient, id).Extract()
+
+		if err != nil {
+			if _, ok := err.(gophercloud.ErrDefault404); ok {
+				return "", "DELETED", nil
+			}
+			return nil, "", fmt.Errorf("Unexpected error: %s", err)
+		}
+
+		return group, "DELETING", nil
+	}
+}
+
+func fwGroupV2IngressPolicyDeleteFunc(networkingClient *gophercloud.ServiceClient, d *schema.ResourceData, ctx context.Context, groupID string) diag.Diagnostics {
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"PENDING_CREATE", "PENDING_UPDATE"},
+		Target:     []string{"ACTIVE", "INACTIVE", "DOWN"},
+		Refresh:    fwGroupV2RefreshFunc(networkingClient, d.Id()),
+		Timeout:    d.Timeout(schema.TimeoutUpdate),
+		Delay:      0,
+		MinTimeout: 2 * time.Second,
+	}
+
+	_, err := stateConf.WaitForStateContext(ctx)
+	if err != nil {
+		return diag.Errorf("Error waiting for openstack_fw_group_v2 %s to become active: %s", d.Id(), err)
+	}
+
+	_, err = groups.RemoveIngressPolicy(networkingClient, groupID).Extract()
+	if err != nil {
+		return diag.Errorf("Error removing ingress firewall policy from openstack_fw_group_v2 %s: %s", d.Id(), err)
+	}
+
+	return nil
+}
+
+func fwGroupV2EgressPolicyDeleteFunc(networkingClient *gophercloud.ServiceClient, d *schema.ResourceData, ctx context.Context, groupID string) diag.Diagnostics {
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"PENDING_CREATE", "PENDING_UPDATE"},
+		Target:     []string{"ACTIVE", "INACTIVE", "DOWN"},
+		Refresh:    fwGroupV2RefreshFunc(networkingClient, d.Id()),
+		Timeout:    d.Timeout(schema.TimeoutUpdate),
+		Delay:      0,
+		MinTimeout: 2 * time.Second,
+	}
+
+	_, err := stateConf.WaitForStateContext(ctx)
+	if err != nil {
+		return diag.Errorf("Error waiting for openstack_fw_group_v2 %s to become active: %s", d.Id(), err)
+	}
+
+	_, err = groups.RemoveEgressPolicy(networkingClient, groupID).Extract()
+	if err != nil {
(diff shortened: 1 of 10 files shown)


openstack/fw_group_v2.go
```diff
@@ -0,0 +1,88 @@
...
+		Target:     []string{"ACTIVE", "INACTIVE", "DOWN"},
+		Refresh:    fwGroupV2RefreshFunc(networkingClient, d.Id()),
+		Timeout:    d.Timeout(schema.TimeoutUpdate),
+		Delay:      0,
+		MinTimeout: 2 * time.Second,
+	}
+
+	_, err := stateConf.WaitForStateContext(ctx)
+	if err != nil {
+		return diag.Errorf("Error waiting for openstack_fw_group_v2 %s to become active: %s", d.Id(), err)
```
kayrus, line 56:
```suggestion
		return diag.Errorf("Error waiting for openstack_fw_group_v2 %s to become deleted: %s", d.Id(), err)
```

openstack/fw_group_v2.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas_v2/groups"
+)
+
+func fwGroupV2RefreshFunc(networkingClient *gophercloud.ServiceClient, id string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		var group groups.Group
+
+		err := groups.Get(networkingClient, id).ExtractIntoStructPtr(&group, "firewall_group")
```
kayrus, line 20: Is there a reason to call the `ExtractIntoStructPtr`? In the `fwGroupV2DeleteFunc` you're just calling `Extract`.
Maybe it makes sense to combine `fwGroupV2RefreshFunc` and `fwGroupV2DeleteFunc` into the one `fwGroupV2RefreshFunc`?

openstack/resource_openstack_fw_group_v2.go
```diff
@@ -0,0 +1,363 @@
...
+func resourceFWGroupV2Create(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	config := meta.(*Config)
+	networkingClient, err := config.NetworkingV2Client(GetRegion(d, config))
+	if err != nil {
+		return diag.Errorf("Error creating OpenStack networking client: %s", err)
+	}
+
+	groupcreateOpts := groups.CreateOpts{
+		Name:                    d.Get("name").(string),
+		TenantID:                d.Get("tenant_id").(string),
```
kayrus, line 111: Does it make sense to add `ProjectID` here as well?

openstack/resource_openstack_fw_group_v2.go
```diff
@@ -0,0 +1,363 @@
...
+		Target:     []string{"DELETED"},
+		Refresh:    fwGroupV2DeleteFunc(networkingClient, d.Id()),
+		Timeout:    d.Timeout(schema.TimeoutDelete),
+		Delay:      0,
+		MinTimeout: 2 * time.Second,
+	}
+
+	_, err = stateConf.WaitForStateContext(ctx)
+	if err != nil {
+		return diag.Errorf("Error waiting for openstack_fw_firewall_v2 %s to Delete:  %s", d.Id(), err)
```
kayrus, line 359:
```suggestion
		return diag.Errorf("Error waiting for openstack_fw_firewall_v2 %s to delete:  %s", d.Id(), err)
```

openstack/resource_openstack_fw_group_v2.go
```diff
@@ -0,0 +1,363 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
+			"status": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
```
kayrus, line 83: This is a computed attribute. Also please move it to the bottom.
```suggestion
			"status": {
				Type:     schema.TypeString,
				Computed: true,
			},
```

website/docs/r/fw_group_v2.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+    id for the firewall group. Changing this updates the
+    `egress_firewall_policy_id` of an existing firewall group.
+
+* `admin_state_up` - (Optional) Administrative up/down status for the firewall
+    group (must be "true" or "false" if provided - defaults to "true").
+    Changing this updates the `admin_state_up` of an existing firewall group.
+
+* `status` - (Optional) Administrative up/down status for the firewall
+    group (must be "true" or "false" if provided - defaults to "true").
+    Changing this updates the `admin_state_up` of an existing firewall group.
```
kayrus, line 97: this should be removed, since it's a computed attribute. Also please add a corresponding description in the attribute reference.

website/docs/r/fw_group_v2.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+    updates the `name` of an existing firewall.
+
+* `description` - (Optional) A description for the firewall group. Changing this
+    updates the `description` of an existing firewall group.
+
+* `tenant_id` - (Optional) The owner of the firewall group. Required if admin
+    wants to create a firewall group for another tenant. Changing this creates
+    a new firewall group.
+
+* `project_id` - (Optional) The owner of the firewall group. Required if admin
```
kayrus, line 79: Please give a description about the difference of `project_id` and `tenant_id`. Do you think it makes sense to add `ConflictsWith` in the schema definition?

kayrus (commented):
Thanks for the PR. See my notes.

website/docs/r/fw_group_v2.html.markdown
```diff
@@ -0,0 +1,131 @@
...
+  protocol         = "udp"
+  destination_port = "123"
+  enabled          = "false"
+}
+
+resource "openstack_fw_policy_v2" "policy_1" {
+  name = "firewall_ingress_policy"
+
+  rules = [
+    "${openstack_fw_rule_v2.rule_1.id}",
```
nikParasyr, line 41: Can you remove `${`, `}` from the examples as it is quite old syntax? with https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1585 all the examples were updated

website/docs/r/fw_group_v2.html.markdown
nikParasyr: I also think `ConflictsWith` might be good.

nikParasyr (changes requested):

openstack/fw_group_v2.go
Koodt: We don't delete the group, but wait for its status after deleting a policy from it.

openstack/fw_group_v2.go
Koodt: Call the `ExtractIntoStructPtr` is needed to return the status dynamically.
I think returning nill error after `gophercloud.ErrDefault404` when we create or update firewall group would be a bad idea.
So, I decided to separate them

website/docs/r/fw_group_v2.html.markdown
Koodt: Description - ok
Why tenant conflict with project? It's the same.
They will only conflict if there are different values. 
In this case, we need the `ConflictsWith`

nikParasyr: I think in its current form it can be a bit confusing for the user:
1. they might either define both with the same value, duplication for no reason
2. they might define them with a different value which will lead to an error

So `ConflictsWith` solves the above and then it can be documented as well to make it clearer for the user that only one is required (and accepted).

Koodt: I completely agree
Done

nikParasyr (approved):

website/docs/r/fw_group_v2.html.markdown
kayrus: The status is still not removed.

openstack/fw_group_v2.go
kayrus: > return the status dynamically

not clear what exactly does it mean :\

website/docs/r/fw_group_v2.html.markdown
Koodt: Shared was removed, not status. Lol.
Fix

openstack/fw_group_v2.go
Koodt: Understood your question.
It's V1 legacy. Align style fwGroupV2RefreshFunc with fwGroupV2DeleteFunc

openstack/resource_openstack_fw_group_v2.go
```diff
@@ -0,0 +1,365 @@
...
+				portIds = append(portIds, v.(string))
+			}
+			updateOpts.Ports = &portIds
+		}
+	}
+
+	if hasChange {
+		log.Printf("[DEBUG] openstack_fw_group_v2 %s update options: %#v", d.Id(), updateOpts)
+
+		err = groups.Update(networkingClient, d.Id(), updateOpts).Err
```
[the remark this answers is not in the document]
Koodt: Done
Also for rules and policies

Koodt:
@kayrus @nikParasyr all checks have passed, all conversations was answered.
Merge?

kayrus (approved):
LGTM, thanks again fro your patience.

kayrus:
@Koodt are you planning to add a data source for FW groups?

Koodt:
@kayrus yes
Right now

PR #1590, opened 2023-06-29
DOCS: Add subcategory for FWaaS and VPNaaS

nikParasyr:


website/openstack.erb
```diff
@@ -1,572 +0,0 @@
-<% wrap_layout :inner do %>
```
nikParasyr, line 1: @kayrus Based on the migration docs i couldn't find where to move this. After checking a bit the contents of the file i think it's no longer required at all ( not 100% though )

PR #1591, opened 2023-06-29
Add deprecation warning to various resources

nikParasyr:
Various resources have been documented as deprecated. Additionally add a deprecation warning for the user.

The related resources have been deprecated on openstack side for many releases now. We do not have any deprecation policy in place and there is no big harm keeping the code (if it doesn't block newer features etc which typically it doesn't). Nonetheless we do not have any ci coverage for these resources for years and it is good to warn the user that these are deprecated.

### DIFF
--- a/openstack/data_source_openstack_blockstorage_snapshot_v2.go
+++ b/openstack/data_source_openstack_blockstorage_snapshot_v2.go
@@ -14,6 +14,7 @@ func dataSourceBlockStorageSnapshotV2() *schema.Resource {
 	return &schema.Resource{
 		ReadContext: dataSourceBlockStorageSnapshotV2Read,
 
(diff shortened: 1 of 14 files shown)


nikParasyr:
@ozerovandrei @mnaser Do you object in "deprecating" these resources?

We don't have a deprecation policy and I do not intend to remove them, but there is no test coverage in the ci for them for years and they are also removed on openstack upstream for many many releases now. This adds a warning message to the user when they use them which I consider to be helpful so that users do not use very old versions.

on https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1590 we moved them on the deprecated subcategory of the docs.

ozerovandrei (approved):

PR #1592, opened 2023-06-29
Add err checks on aggregate on fw_v2 resources

nikParasyr:
Add error checking on aggregate types for fw_v2 resources based on [best practises](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/detecting-drift#error-checking-aggregate-types)

### DIFF
--- a/openstack/data_source_openstack_fw_group_v2.go
+++ b/openstack/data_source_openstack_fw_group_v2.go
@@ -170,9 +170,12 @@ func dataSourceFWGroupV2Read(ctx context.Context, d *schema.ResourceData, meta i
 	d.Set("admin_state_up", group.AdminStateUp)
 	d.Set("ingress_firewall_policy_id", group.IngressFirewallPolicyID)
 	d.Set("egress_firewall_policy_id", group.EgressFirewallPolicyID)
-	d.Set("ports", group.Ports)
 	d.Set("status", group.Status)
 	d.Set("region", GetRegion(d, config))
+
+	if err := d.Set("ports", group.Ports); err != nil {
+		diag.Errorf("Unable to set ports for openstack_fw_group_v2 %s: %s", group.ID, err)
+	}
 
 	return nil
 }
--- a/openstack/data_source_openstack_fw_policy_v2.go
+++ b/openstack/data_source_openstack_fw_policy_v2.go
@@ -129,8 +129,11 @@ func dataSourceFWPolicyV2Read(ctx context.Context, d *schema.ResourceData, meta 
 	d.Set("description", policy.Description)
 	d.Set("shared", policy.Shared)
 	d.Set("audited", policy.Audited)
-	d.Set("rules", policy.Rules)
 	d.Set("region", GetRegion(d, config))
+
+	if err := d.Set("rules", policy.Rules); err != nil {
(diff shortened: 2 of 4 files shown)


nikParasyr:
@kayrus I noticed that err checks were added on https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/456 but they are just logged. Do you recall the reasoning?

kayrus:
>  Do you recall the reasoning?

Not really...

BTW, There is a special `TF_SCHEMA_PANIC_ON_ERROR=1` env variable that allows to panic on `Set` to catch such issues. See #601

openstack/data_source_openstack_fw_group_v2.go
```diff
@@ -170,9 +170,12 @@ func dataSourceFWGroupV2Read(ctx context.Context, d *schema.ResourceData, meta i
 	d.Set("admin_state_up", group.AdminStateUp)
 	d.Set("ingress_firewall_policy_id", group.IngressFirewallPolicyID)
 	d.Set("egress_firewall_policy_id", group.EgressFirewallPolicyID)
-	d.Set("ports", group.Ports)
 	d.Set("status", group.Status)
 	d.Set("region", GetRegion(d, config))
 
+	if err := d.Set("ports", group.Ports); err != nil {
+		diag.Errorf("Unable to set ports for openstack_fw_group_v2 %s: %s", group.ID, err)
```
kayrus, line 177: shouldn't we return the `diag.Errorf`? And does it make sense to check all `Set` err codes?

nikParasyr: > shouldn't we return the diag.Errorf?

yes

> And does it make sense to check all Set err codes?

Only the aggregate types (schema.TypeList, schema.TypeSet, and schema.TypeMap)
but we probably want to standardize this. We are logging it in various resourcess, but not in all

nikParasyr:
> BTW, There is a special TF_SCHEMA_PANIC_ON_ERROR=1 env variable that allows to panic on Set to catch such issues. See https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/601

I'm reading that this is enabled by default on sdkv2. But i'll trigger a ci pipeline to see what we get

openstack/data_source_openstack_fw_group_v2.go
kayrus: There is also a change to set Bool to String and String to Bool ;)

nikParasyr: I set `TF_SCHEMA_PANIC_ON_ERROR=1` on all ci in #1593 and everything passed. From various docs i found that it was defaulted to 1 when TF_ACC is set on sdkv2. So probably we were able to catch all the possible issues on the ci.

I'm not sure whether we want to implement this tbh. We have it in some resources but not all, and it is just logged. I came across it while going through the docs and I recalled the checks werent added of fw resources.
I think i'll close the PR as it is a more "generic" issue, to standardize the check.

I wonder what could be a good way of standardizing this. I checked a bit the aws provider and they only check for attributes they flatten etc. Checking all aggregate types + Bool + String seems quite an overkill to me.

Not sure how you feel but the aws approach seems good to me. Checks the attributes with the highest possibility of being wrong, but doesn't add 100s of checks on all kind of attributes.

PR #1594, opened 2023-06-30
Add project_id to fw_v2 resources and datasources

nikParasyr:
Add project_id to the remaining fw_v2 resources and data sources. Bump gophercloud for required updates


kayrus (approved):
LGTM, thanks

PR #1595, opened 2023-06-30
Fix firewall_policy_id in fw_rule_v2 data-source

nikParasyr:
firewall_policy_id is TypeString. Moreover add it to the lisstOpts


nikParasyr:
Closing this... Seems like it's a list after all and the api docss are wrong.
More info: https://github.com/gophercloud/gophercloud/pull/2057#pullrequestreview-543179868

PR #1615, opened 2023-10-17
Upgrade to use the Glance API instead of Nova for images

Wouter0100:
The images API available in the Nova API has been deprecated and is currently only a proxy call. This migrates the provider over towards using the Glance API directly.


nikParasyr (approved):
@Wouter0100 thanks for the PR
LGTM

PR #1624, opened 2023-11-03
data_source_blockstorage_volume_v3: Add Attachment Information

nudelfabrik:
Hello,
I am currently reworking the terraform setup for the openstack cloud at my work. While figuring out how to best handle long-term volumes, like databases that should ideally not be deleted recreated if something changes, I noticed a discrepancy between the _**resource** openstack_blockstorage_volume_v3_ and the _**data source** openstack_blockstorage_volume_v3_:

### Issue

The result of creating a volume via terraform has the attribute "attachment", which exposes the instance and the location on which the Volume is mounted (e.g. /dev/vdb on Linux) See [documentation here.](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/blockstorage_volume_v3#attributes-reference)
However, when creating the volume manually outside of terraform, and using the data source to get the same information, the result is missing this attribute.

Example:
```
# Get info on an existing volume with this name
data "openstack_blockstorage_volume_v3" "existing_volume" {
  name = "existing-volume-for-instance-x"
}
# Create new volume, managed by terraform
resource "openstack_blockstorage_volume_v3" "new_volume" {
  name = "new-volume-for-instance-x"
  size = 42
  # etc.
}

# Works, gives the right output
output "new_attachment" {
  value = openstack_blockstorage_volume_v3.new_volume.attachment
}

# Error, This object does not have an attribute named "attachment"
output "existing_attachment" {
  value = data.openstack_blockstorage_volume_v3.existing_volume.attachment
}
```

### Proposal 
Taking a curious look at the code of the resource [(this part](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_blockstorage_volume_v3.go#L269-L274) especially), I noticed that there the code itself is very similar between the resource and the data source, so I tried to just copy the relevant part over to the corresponding function for the data source.

Testing it in our environment, this does indeed work, I now get the correct mountpoint for the volume, (if I make sure the volume is attached before I retrieve the data).
I have to admit I did no further in-depth testing than this, I am not familiar enough with either the code or openstack/terraform to identify potential edge cases that may make it more complicated than this. For me it looks like it just needs to extract and expose the information the same way it is done for the resource, since it only reads data, not needing to manipulate anything.

I hope this PR provides some useful functionality,
Jess Scholtes

### DIFF
--- a/openstack/data_source_openstack_blockstorage_volume_v3_test.go
+++ b/openstack/data_source_openstack_blockstorage_volume_v3_test.go
@@ -122,3 +122,50 @@ func testAccBlockStorageV3VolumeDataSourceBasic(snapshotName string) string {
     }
   `, snapshotName)
 }
+
+func TestAccBlockStorageV3VolumeDataSource_attachment(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckNonAdminOnly(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckBlockStorageV3VolumeDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccBlockStorageV3VolumeDataSourceAttachment(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("data.openstack_blockstorage_volume_v3.volume_1", "attachment", "openstack_blockstorage_volume_v3.data_1", "attachment"),
+				),
+			},
+		},
+	})
+}
+
+func testAccBlockStorageV3VolumeDataSourceAttachment() string {
+	return fmt.Sprintf(`
+resource "openstack_blockstorage_volume_v3" "volume_1" {
+  name = "volume_1"
+  size = 1
+}
+
+resource "openstack_compute_instance_v2" "instance_1" {
+  name = "instance_1"
+  security_groups = ["default"]
+  network {
+    uuid = "%s"
+  }
+}
+
+resource "openstack_compute_volume_attach_v2" "va_1" {
+  instance_id = "${openstack_compute_instance_v2.instance_1.id}"
+  volume_id = "${openstack_blockstorage_volume_v3.volume_1.id}"
+  device = "/dev/vdc"
+}
+
+data "openstack_blockstorage_volume_v3" "volume_1" {
+  name = "volume_1"
+  depends_on = [ openstack_compute_volume_attach_v2.va_1 ]
+}
+`, osNetworkID)
+}
--- a/openstack/data_source_openstack_blockstorage_volume_v3_test.go
+++ b/openstack/data_source_openstack_blockstorage_volume_v3_test.go
@@ -135,7 +135,7 @@ func TestAccBlockStorageV3VolumeDataSource_attachment(t *testing.T) {
 			{
 				Config: testAccBlockStorageV3VolumeDataSourceAttachment(),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttrPair("data.openstack_blockstorage_volume_v3.volume_1", "attachment", "openstack_blockstorage_volume_v3.data_1", "attachment"),
+					resource.TestCheckResourceAttrPair("data.openstack_blockstorage_volume_v3.volume_1", "attachment", "openstack_blockstorage_volume_v3.volume_1", "attachment"),
 				),
 			},
 		},
--- a/openstack/data_source_openstack_blockstorage_volume_v3_test.go
+++ b/openstack/data_source_openstack_blockstorage_volume_v3_test.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


nikParasyr:
@nudelfabrik The failing ci job is unrelated to this PR.
Could you please add an acceptance test for this case?

nudelfabrik:
@nikParasyr Sure, I guess you mean the tests in the _test.go files? https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/data_source_openstack_blockstorage_volume_v3_test.go

I will take a look at how that works, and add a test for it.

nikParasyr:
@nudelfabrik yes indeed. In this test case you should also create a instance, a volume and an attachment, so that in the data source you can verify that the attachment field is set.

To give some guidelines I'll use this test ( for a different resource): https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/data_source_openstack_compute_instance_v2_test.go

So in your case, you will have an initial config step that creates all 3 resources. Similar to [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/data_source_openstack_compute_instance_v2_test.go#L20). The next [step](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/data_source_openstack_compute_instance_v2_test.go#L23) shouldbe a simple volume data source that will read the volume. Then in the check of that [step](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/data_source_openstack_compute_instance_v2_test.go#L26) you should add checks that the attachment has been set as expected.

You test should be added [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/data_source_openstack_blockstorage_volume_v3_test.go) as a new test.

hope this helps a bit

nudelfabrik:
Thanks, that really helped!
Except I accidentally looked into the tests for the resource, not the data-source, and added there. I will add it to the right one as well, but I noticed that the unittests fail with a 404 error, that the Volume is not created. Not sure, if I made a mistake with the test somewhere?

openstack/resource_openstack_blockstorage_volume_v3_test.go
```diff
@@ -145,6 +145,28 @@ func TestAccBlockStorageV3Volume_timeout(t *testing.T) {
...
+			testAccPreCheckNonAdminOnly(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckBlockStorageV3VolumeDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccBlockStorageV3VolumeAttachment(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckBlockStorageV3VolumeExists("openstack_blockstorage_volume_v3.volume_1", &volume),
+					testAccCheckBlockStorageV3VolumeAttachment(&volume, "dev/vdc"),
```
nikParasyr, line 163: This should be `/dev/vdc`. Ci tests are failing due to the missing `/`

openstack/data_source_openstack_blockstorage_volume_v3_test.go
```diff
@@ -122,3 +122,50 @@ func testAccBlockStorageV3VolumeDataSourceBasic(snapshotName string) string {
...
+			testAccPreCheckNonAdminOnly(t)
+		},
+		ProviderFactories: testAccProviders,
+		CheckDestroy:      testAccCheckBlockStorageV3VolumeDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccBlockStorageV3VolumeDataSourceAttachment(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("data.openstack_blockstorage_volume_v3.volume_1", "attachment", "openstack_blockstorage_volume_v3.volume_1", "attachment"),
+				),
```
nikParasyr, line 139: Can you also add a `testAccCheckBlockStorageV3VolumeAttachment` check here as well?

nikParasyr (changes requested):
@nudelfabrik Sorry for the delayed review but was away.

CI tests fail due to the comments. Other than that, looks good

openstack/resource_openstack_blockstorage_volume_v3_test.go
kayrus: actually this case won't work in other OS env. I'd suggest to make it more multipurpose/dynamic.

> Note that as of the 12.0.0 Liberty release, the Nova libvirt driver no longer honors a user-supplied device name. This is the same behavior as if the device name parameter is not supplied on the request.

nikParasyr: Hmmm, Idd. I do recall seeing this. In the ci though it is still mounted under `/dev/vdc`.
But a `VolumeAttachment` exists should avoid these issues and be enough right?

kayrus: I'd just add a regexp check... Since block device can be sdc/vdc/hdc, etc...

nudelfabrik:
Sorry, I was not able to respond sooner, something else got priority, and I had no time to look into what was still required for the Unittests.

Thank you for finishing and merging it :heart:

PR #1629, opened 2023-12-05
Add deprecation notice on multiattach volumes

nikParasyr:
Multiattach volumes has been deprecated since Queens and completely removed on Bobcat. Add deprecation notice on the argument and set the tests to skip


kayrus (approved):
LGTM, thanks

PR #1631, opened 2023-12-05
Update Terraform SDK and gophercloud

nikParasyr:


.github/workflows/functional-blockstorage_v2.yml
```diff
@@ -8,7 +8,6 @@ on:
   pull_request:
     paths:
       - '.github/workflows/functional-blockstorage_v2.yml'
-      - 'go.mod'
```
kayrus, line 11: does it mean that we'll skip functional tests on go.mod update?

kayrus (commented):
Don't you think it's time to bump ubuntu VMs from 20.04 to 22.04?

.github/workflows/functional-blockstorage_v2.yml
nikParasyr: > Don't you think it's time to bump ubuntu VMs from 20.04 to 22.04?

blockstorage_v2 was completely removed on Xena and devstack xena does not support 22.04. Or at least it didnt when i tried to set it up. 22.04 is used on the other ci jobs that are not deprecated/removed.

> does it mean that we'll skip functional tests on go.mod update?

Indeed it does, and now im realizing that even with it being deprecated, we probably still want to test it when changing `go.mod`. So ill add it back there

kayrus (approved):

PR #1637, opened 2024-01-02
[openstack_blockstorage_volume_v3] Create cinder volume from backup

ZdenekPesek:
Close: https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1636

### DIFF
--- a/openstack/resource_openstack_blockstorage_volume_v3.go
+++ b/openstack/resource_openstack_blockstorage_volume_v3.go
@@ -93,6 +93,11 @@ func resourceBlockStorageVolumeV3() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"backup_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
 			"volume_type": {
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


openstack/resource_openstack_blockstorage_volume_v3.go
```diff
@@ -93,6 +93,11 @@ func resourceBlockStorageVolumeV3() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"backup_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
nikParasyr, line 99: Also `ForceNew: true,` should be set

openstack/resource_openstack_blockstorage_volume_v3.go
```diff
@@ -193,6 +198,7 @@ func resourceBlockStorageVolumeV3Create(ctx context.Context, d *schema.ResourceD
 		ConsistencyGroupID: d.Get("consistency_group_id").(string),
 		Description:        d.Get("description").(string),
 		ImageID:            d.Get("image_id").(string),
+		BackupID:           d.Get("backup_id").(string),
```
nikParasyr, line 201: When the user has defined `backup_id` you need to set the blockstorage client microversion to 3.47

nikParasyr (changes requested):
@ZdenekPesek Hello. thanks for opening this.

Have left a couple of comments.

Moreover there should be a test case for this and also docs should be updated.

openstack/resource_openstack_blockstorage_volume_v3.go
kayrus: I think there should be a conflict set for all sources:
* `snapshot_id`
* `source_vol_id`
* `image_id`
* `backup_id`

PR #1654, opened 2024-01-29
Remove support for "multiattach" on blockstorage_volume_v3

nikParasyr:
Remove the deprecated "multiattach" option on blockstorage_volume_v3. This includes both the resource and the data-source. To create a multiattach volume, a multiattach-enabled volume-type has to be used. Update the docs to provide such an example.

Part of #1660


kayrus (approved):
LGTM, thanks!

PR #1655, opened 2024-01-29
Remove support of neutron-lbaas from lb_l7policy

nikParasyr:
Part of: https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1638

### DIFF
--- a/openstack/lb_v2_octavia.go
+++ b/openstack/lb_v2_octavia.go
@@ -1,0 +1,305 @@
+package openstack
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/l7policies"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+)
+
+func resourceLBV2ListenerRefreshFuncOctavia(lbClient *gophercloud.ServiceClient, lbID string, listener *listeners.Listener) resource.StateRefreshFunc {
+	if listener.ProvisioningStatus == "" {
+		return resourceLBV2LoadBalancerStatusRefreshFuncOctavia(lbClient, lbID, "listener", listener.ID, "")
+	}
+
+	return func() (interface{}, string, error) {
+		lb, status, err := resourceLBV2LoadBalancerRefreshFunc(lbClient, lbID)()
+		if err != nil {
+			return lb, status, err
+		}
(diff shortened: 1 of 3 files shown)


openstack/lb_v2_octavia.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/l7policies"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/listeners"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+)
+
+func resourceLBV2ListenerRefreshFuncOctavia(lbClient *gophercloud.ServiceClient, lbID string, listener *listeners.Listener) resource.StateRefreshFunc {
+	if listener.ProvisioningStatus != "" {
```
kayrus, line 20: This is my code I wrote 5 years ago :) Not sure if this is the right place to ask this, but with my current experience I'd check for empty string and return `resourceLBV2LoadBalancerStatusRefreshFuncOctavia` right away.

openstack/lb_v2_octavia.go
```diff
@@ -0,0 +1,305 @@
...
+		if len(listener.Loadbalancers) > 0 {
+			return listener.Loadbalancers[0].ID, nil
+		}
+	}
+
+	return "", fmt.Errorf("Unable to determine loadbalancer ID from pool %s", pool.ID)
+}
+
+func resourceLBV2PoolRefreshFuncOctavia(lbClient *gophercloud.ServiceClient, lbID string, pool *pools.Pool) resource.StateRefreshFunc {
+	if pool.ProvisioningStatus != "" {
```
kayrus, line 95: here as well

openstack/lb_v2_octavia.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		return fmt.Errorf("Error waiting for l7policy %s to become %s: %s", l7policy.ID, target, err)
+	}
+
+	return nil
+}
+
+func resourceLBV2L7PolicyRefreshFuncOctavia(lbClient *gophercloud.ServiceClient, lbID string, l7policy *l7policies.L7Policy) resource.StateRefreshFunc {
+	if l7policy.ProvisioningStatus != "" {
```
kayrus, line 285: and here too

kayrus (commented):
Thanks for the PR. See my comments

openstack/lb_v2_octavia.go
nikParasyr: @kayrus can you elaborate a bit more?

> I'd check for empty string and return resourceLBV2LoadBalancerStatusRefreshFuncOctavia right away.

Return resourceLBV2LoadBalancerStatusRefreshFuncOctavia when == "" or != "" ? And what do we return in the other case? 
Cause after looking a bit the code i think we can always return resourceLBV2LoadBalancerStatusRefreshFuncOctavia no matter the check, but maybe im misunderstanding something.

kayrus: I mean this: https://danp.net/posts/reducing-go-nesting/

kayrus (approved):
LGTM, thanks!

Issue #1660, opened 2024-01-29
Deprecate various attributes for v2.0.0


nikParasyr:
There are mutliple attributes marked as deprecated for a while:
```
openstack/data_source_openstack_images_image_ids_v2.go:                         Deprecated:    "Use option 'sort' instead.",
openstack/data_source_openstack_images_image_ids_v2.go:                         Deprecated:   "Use option 'sort' instead.",
openstack/data_source_openstack_networking_subnet_v2.go:                                Deprecated:    "use dhcp_enabled instead",
openstack/provider.go:                          Deprecated:  "This option will be removed in the next major release. Support for neutron-lbaas will be removed in next major release. Octavia will be the only option supported.",
openstack/resource_openstack_blockstorage_volume_attach_v2.go:                          Deprecated: "instance_id is no longer used in this resource",
openstack/resource_openstack_blockstorage_volume_v3.go:                         Deprecated: "multiattach parameter has been deprecated and removed on Openstack Bobcat. The default behavior is to use multiattach enabled volume types",
openstack/resource_openstack_compute_instance_v2.go:                            Deprecated: "Use the openstack_compute_floatingip_associate_v2 resource instead",
openstack/resource_openstack_compute_instance_v2.go:                                                    Deprecated: "Use the openstack_compute_floatingip_associate_v2 resource instead",
openstack/resource_openstack_compute_instance_v2.go:                            Deprecated: "Use block_device or openstack_compute_volume_attach_v2 instead",
openstack/resource_openstack_images_image_v2.go:                                Deprecated: "Use updated_at instead",
openstack/resource_openstack_lb_pool_v1.go:                             Deprecated: "Use openstack_lb_member_v1 instead",
openstack/resource_openstack_networking_router_v2.go:                           Deprecated:    "use external_network_id instead",
openstack/resource_openstack_networking_subnet_v2.go:                           Deprecated:    "use allocation_pool instead",
openstack/resource_openstack_networking_subnet_v2.go:                           Deprecated: "Use openstack_networking_subnet_route_v2 instead",
openstack/resource_openstack_objectstorage_container_v1.go:                             Deprecated:    "Use newer \"versioning\" implementation",
```

the 2.0.0 release is a good opportunity to remove some/all of them.

PR #1661, opened 2024-01-29
Remove deprecated sort_key and sort_dir from glance data sources

nikParasyr:
Glance data sources allowed sorting in 2 different ways. Sort or sort_key + sort_dir. Sort can fully replace the sort_key + sort_dir combination and is capable of providing multiple sorting criteria. sort_key and sort_dir are marked as deprecated on the provider. Update glance data sources to use solely `sort` and remove support for the deprecated (in the provider) sort_key + sort_dir

Part of #1660

### DIFF
--- a/openstack/images_image_v2.go
+++ b/openstack/images_image_v2.go
@@ -372,3 +372,42 @@ func imagesFilterByProperties(v []images.Image, p map[string]string) []images.Im
 
 	return result
 }
+
+// dataSourceValidateImageSortFilter checks that the sorting filter passed
+// by a user follows the glance API definition of "sort_key:sort_dir,sort_key:sort_dir:.."
+// where sort_dir is optional. For more details, check the glance api docs
+// https://docs.openstack.org/api-ref/image/v2/index.html#list-images
+// at the `sorting` section.
+func dataSourceValidateImageSortFilter(v interface{}, k string) (ws []string, errors []error) {
+	validSortKeys := []string{
+		"name",
+		"owner",
+		"protected",
+		"status",
+		"tag",
+		"visibility",
+		"container_format",
+		"disk_format",
+		"os_hidden",
+		"member_status",
+		"created_at",
+		"updated_at",
+	}
+	validSortDirections := []string{
+		"asc",
+		"desc",
+	}
+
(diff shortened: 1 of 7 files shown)


kayrus (commented):
Removing validation may cause issues, since these sorting options will be most probably ignored by openstack API.

nikParasyr:
> Removing validation may cause issues, since these sorting options will be most probably ignored by openstack API.

Added new validation func that should cover this

openstack/images_image_v2.go
```diff
@@ -372,3 +372,25 @@ func imagesFilterByProperties(v []images.Image, p map[string]string) []images.Im
 
 	return result
 }
+
+// dataSourceValidateImageSortFilter checks that the sorting filter passed
+// by a user follows the glance API definition of "sort_key:sort_dir,sort_key:sort_dir:.."
+// where sort_dir is optional.
+func dataSourceValidateImageSortFilter(v interface{}, k string) (ws []string, errors []error) {
+	validSortKeys := []string{"name", "owner", "protected", "status", "tag", "visibility",
+		"os_hidden", "member_status", "created_at", "updated_at"}
```
kayrus, line 381: It's better to enum these value on each line. This will reduce confusion on diff once new fields are added.

```suggestion
	validSortKeys := []string{
		"name",
		"owner",
		"protected",
		"status",
		"tag",
		"visibility",
		"os_hidden",
		"member_status",
		"created_at",
		"updated_at",
	}
```

In addition I think it makes sense to add a link to upstream OpenStack docs, e.g. https://docs.openstack.org/glance/latest/user/glanceapi.html#filtering-images-lists

kayrus (commented):
thanks for the update, see my comments

openstack/images_image_v2.go
nikParasyr: Done. 

> https://docs.openstack.org/glance/latest/user/glanceapi.html#filtering-images-lists

This link covers filtering and not sorting. I looked for the equivalent for sorting but couldn't find it. I have, instead, added the link to the glance api docs the cover sorting briefly.

kayrus (approved):
LGTM, thanks!

PR #1663, opened 2024-01-29
Remove dhcp_disabled from data_source_openstack_networking_subnet_v2

nikParasyr:
dhcp_disabled has been deprecated in favor of dhcp_enabled. Remove it from the provider

Part of #1660


kayrus (approved):
Sounds reasonable without the deprecation note. Besides there is no a reference in docs.

Issue #1626, opened 2023-11-30
Security group rule operations are very slow


echaiv:
Security group rule operations are serialized by #994. This seems to have been done to workaround a bug in "Contrail." Neutron does not have this defect and can preform security group rule operations in parallel. Serializing rule operations causes a significant performance regression.

It would be nice to make this mutex optional and add a flag to either enable or disable the workaround for contrail. I have a patch for this and can make a PR.

### Terraform Version
Affects multiple terraform versions. Tested with several terraform versions and several versions of the openstack provider. The issue seems to have been introduced with version 1.28.0 of the openstack provider.

```
1.0.1   + 1.27.0 -> 3m53s
1.2.5   + 1.27.0 -> 3m39s
1.6.5   + 1.27.0 -> 3m38s

1.0.1   + 1.28.0 -> 17m40s
1.2.5   + 1.28.0 -> 17m27s
1.6.5   + 1.28.0 -> 25m31s

1.0.1   + 1.53.0 -> 31m30s
1.2.5   + 1.53.0 -> 31m26s
1.6.5   + 1.53.0 -> 31m24s
```

### Affected Resource(s)
- openstack_networking_secgroup_rule_v2

### Terraform Configuration Files
```hcl
# Minimal test case to create a bunch of rules
locals {
  rules = {
    for i in range(350):
    i => {
      group = "test_sg"
      direction = "ingress"
      ethertype = "IPv4"
      protocol = "tcp"
      port_range_min = 1234
      port_range_max = 1234
      remote_ip_prefix = format("10.%d.%d.%d/32", floor(i / (256 * 256)), floor(i / 256) % 256, i % 256)
    }
  }
  groups = toset([for k, rule in local.rules: rule.group])
}

resource "openstack_networking_secgroup_v2" "group" {
  for_each             = local.groups
  name                 = each.key
  description          = "${each.key}"
  delete_default_rules = true
}

resource "openstack_networking_secgroup_rule_v2" "rule" {
  for_each          = local.rules
  security_group_id = resource.openstack_networking_secgroup_v2.group[each.value.group].id
  direction         = each.value.direction
  ethertype         = each.value.ethertype
  protocol          = each.value.protocol
  port_range_min    = each.value.port_range_min
  port_range_max    = each.value.port_range_max
  remote_ip_prefix  = each.value.remote_ip_prefix
}
```

### Panic Output
No panic

### Expected Behavior
Terraform creates security group rules in parallel.

[...]

PR #1672, opened 2024-02-13
Add a flag to allow parallel network API calls

echaiv:
This PR adds a configuration option `use_mutex` which defaults to true. When set to true, a mutex is used to serialize network API calls. If the user overrides this by setting `use_mutex` to false, then parallel network operations are allowed. This greatly improves performance when using terraform-provider-openstack to manage a large number of network resources.

Provides a workaround for #1626.


kayrus:
@echaiv Thanks for the PR. Does it make sense to implement a new flag per resource?

PR #1689, opened 2024-03-06
Remove versioning_legacy from objectstorage_container_v1

nikParasyr:
Partially implements https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1660


nikParasyr:
@kayrus would really like review and your opinion on this as you wrote the deprecation and the state removal.

I went through swift [docs](https://docs.openstack.org/swift/latest/middleware.html#module-swift.common.middleware.versioned_writes.object_versioning), and i'm not sure whether the 2 legacy modes are deprecated and/or planned for removal upstream.

PR #1691, opened 2024-03-08
[compute v2] Support "build" status

pdesgarets:
## Current behavior

if an instance takes too much time to spawn, terraform throws an error "Error waiting for instance (...) to become ready: context deadline exceeded"

Then, if `terraform apply` is run while the instance is still building, the whole command fails with "Invalid power_state for instance [...]: BUILD"

## Future behavior

This PR aims at solving this by supporting the "build" power_state when reading the instance.
1. The `terraform apply` fails with `context deadline exceeded`
2. The instance is tainted due to this error and must be manually untainted
3. When `terraform apply` is run again while the instance is building, the following diff is shown :
```
  # openstack_compute_instance_v2.worker will be updated in-place
  ~ resource "openstack_compute_instance_v2" "worker" {
        id                  = "..."
        name                = "mydocker-terraform-test-pdg"
      ~ power_state         = "build" -> "active"
        tags                = []
        # (14 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
Plan: 0 to add, 1 to change, 0 to destroy.
```
4. When apply is confirmed, terraform waits for the instance to be "active" (and possibly fails again with "context deadline exceeded" if the instance creation takes again too much time in regard to the configured timeout)

Note : as security groups and networks are not available during building phase, we must ignore them while instance is building (otherwise it would always trigger a Renew). This means that, if we change network or security groups in the config while the instance is still building, the diff will be ignored. This seems an edge case and a reasonable behavior to me, but might worth a line in the doc.

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -422,6 +422,19 @@ func resourceComputeInstanceV2() *schema.Resource {
 			customdiff.ForceNewIfChange("flavor_name", func(ctx context.Context, old, new, meta interface{}) bool {
 				return old.(string) == ""
 			}),
+			func(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
+				currentState, _ := d.GetChange("power_state")
+				if currentState == "build" {
+					// In "build" state, network and security groups are not yet available
+					if err := d.Clear("network"); err != nil {
+						return err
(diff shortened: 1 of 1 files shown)


pdesgarets:
Code review for the poor

kayrus (approved):
LGTM, thanks!

PR #1712, opened 2024-05-13
Add references between compute_servergroup and compute_instance docs

jlalford:
docs/resources/compute_instance_v2.md: Add reference to servergroups resource docs for instance resource attribute
docs/resources/compute_servergroup_v2.md: Add reference to instance resource attribute in overview

### DIFF
--- a/docs/resources/compute_instance_v2.md
+++ b/docs/resources/compute_instance_v2.md
@@ -397,6 +397,8 @@ The following arguments are supported:
 
 * `scheduler_hints` - (Optional) Provide the Nova scheduler with hints on how
     the instance should be launched. The available hints are described below.
+    See [reference](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_servergroup_v2)
+    for details on managing servergroup resources
 
 * `personality` - (Optional) Customize the personality of an instance by
     defining one or more files and their contents. The personality structure
--- a/docs/resources/compute_instance_v2.md
+++ b/docs/resources/compute_instance_v2.md
@@ -397,8 +397,6 @@ The following arguments are supported:
 
 * `scheduler_hints` - (Optional) Provide the Nova scheduler with hints on how
     the instance should be launched. The available hints are described below.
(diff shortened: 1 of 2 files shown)


docs/resources/compute_instance_v2.md
```diff
@@ -397,6 +397,8 @@ The following arguments are supported:
 
 * `scheduler_hints` - (Optional) Provide the Nova scheduler with hints on how
     the instance should be launched. The available hints are described below.
+    See [reference](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_servergroup_v2)
```
nikParasyr, line 400: `scheduler_hints` can be used for more than just server groups. I think the reference is better put [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/docs/resources/compute_instance_v2.md?plain=1#L489) where it covers just server groups

docs/resources/compute_servergroup_v2.md
```diff
@@ -10,6 +10,8 @@ description: |-
 # openstack\_compute\_servergroup\_v2
 
 Manages a V2 Server Group resource within OpenStack.
+See [reference](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_instance_v2)
```
nikParasyr, line 13: I would prefer instead of the reference to extend the example below with a instance resource that uses schedulers hint to use the server group. This way it will make the connection clear but also offer a code example for users

nikParasyr (changes requested):

docs/resources/compute_instance_v2.md
jalford-ns: Thanks! Updated

docs/resources/compute_servergroup_v2.md
jalford-ns: Thanks! Updated

nikParasyr (approved):
LGTM. thanks for the fixes :)

PR #1713, opened 2024-05-14
Add volume attach tag with nova microversion 2.49

jichenjc:
The API is 
https://docs.openstack.org/api-ref/compute/#attach-a-volume-to-an-instance

and the opts in gophercloud is 
https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/volumeattach/requests.go#L34

### DIFF
--- a/docs/resources/compute_volume_attach_v2.md
+++ b/docs/resources/compute_volume_attach_v2.md
@@ -156,6 +156,10 @@ The following arguments are supported:
 
 * `multiattach` - (Optional) Enable attachment of multiattach-capable volumes.
 
+* `tag` - (Optional) Add a device role tag that is applied to the volume when
+  attaching it to the VM. Changing this creates a new volume attachment with
+  the new tag. Requires microversion >= 2.49.
+
 * `vendor_options` - (Optional) Map of additional vendor-specific options.
   Supported options are described below.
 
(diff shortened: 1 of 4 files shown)


jichenjc:
@nikParasyr could you please help review this ? THanks

docs/resources/compute_volume_attach_v2.md
```diff
@@ -156,6 +156,9 @@ The following arguments are supported:
 
 * `multiattach` - (Optional) Enable attachment of multiattach-capable volumes.
 
+* `tag` - (Optional) The tag of the volume attachment. Changing this
```
nikParasyr, line 159:
```suggestion
* `tag` - (Optional) Add a device role tag that is applied to the volume when attaching it to the VM. Changing this
```

docs/resources/compute_volume_attach_v2.md
```diff
@@ -156,6 +156,9 @@ The following arguments are supported:
...
+    updates the tag of volume attachment. Requires microversion >= 2.49.
```
nikParasyr, line 160:
```suggestion
    creates a new volume attachment with the new tag. Requires microversion >= 2.49.
```

Since `ForceNew` is set below updating the tag will force a recreation of the resource.

nikParasyr (changes requested):
Thank you for this.
Some small nits on docs

jichenjc:
@nikParasyr thanks for the info, updated the doc accordingly

nikParasyr (approved):
LGTM

PR #1716, opened 2024-05-16
Add openstack_lb_flavorprofile_v2 resource

nikParasyr:
Add openstack_lb_flavorprofile_v2 resource for Octavia. Add acceptance tests. Add import function as well with tests.


openstack/resource_openstack_lb_flavorprofile_v2.go
```diff
@@ -0,0 +1,171 @@
...
+
+			"provider_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			// flavor_data depends on which provider is being used.
+			// Therefore we stay close to the API and make it type
+			// String. The user can use jsonencode to pass it properly
+			"flavor_data": {
```
kayrus, line 52: maybe it's better to use `github.com/hashicorp/terraform-plugin-sdk/v2/helper/structure.NormalizeJsonString`?

PR #1719, opened 2024-05-21
Allow Magnum node_count=0 for cluster update

artpej:
Since  1.10 Magnum allow cluster creation with worker node count set to 0. This one  set the correct microversion to allow node_count=0 on cluster update.


nikParasyr (approved):
LGTM. thanks

PR #1733, opened 2024-06-25
Updated: Record Load-balancer in error

### LINKED ISSUE #1734: Add an Load-Balancer in error in the state
Hi there,

When a Load-Balancer is created, if the creation fails on OpenStack side, the Load-Balancer is in error state.

Even if the resource is not available, networks ports are still created and therefore not adding the LB in error in the tfstate leads to errors while deleting network.

If TF tries to delete a network while there are remaining ports that where linked to a failed LB (which exists when you do a openstack loadblancer list but was never present on the TF state) it will fail.

### Terraform Version
1.8.5.

### Affected Resource(s)
 - resource_openstack_lb_loadbalancer_v2

### Expected Behavior
A load-balancer present in the openstack project, created by terraform should be present in the tf state. 

### Actual Behavior
If the creation fails for some openstack reason, the loadbalancer is in error state in openstack but absent in the tf state. 
During the creation, openstack may have created ports and other attachments leading tf unable to delete the network because tf does not record a lb in error and therefore will never delete linked resources. 

### Steps to Reproduce
1. `terraform apply`
2. Set the lb in error state
3. `tf apply` will error, timing out while waiting for the lb to be active
4. LB is present on openstack side, ports are created. But tf will try to recreate another LB. 

### References
- #1733

al26p:
Closes #1734 

When a Load-Balancer is created, if the creation fails on OpenStack side, the Load-Balancer is in error state. 
Even if the resource is not available, networks ports are still created and therefore not adding the LB in error in the tfstate leads to error while deleting network. 
TF tries to delete a network while there are remaining ports that where linked to the failed LB (which exists when you do a `openstack loadblancer list`). But was never present on the TF state.

### DIFF
--- a/openstack/lb_v2_shared.go
+++ b/openstack/lb_v2_shared.go
@@ -598,11 +598,11 @@ func flattenLBPoolPersistenceV2(p pools.SessionPersistence) []map[string]interfa
 	}
 }
 
-func waitForLBV2LoadBalancer(ctx context.Context, lbClient *gophercloud.ServiceClient, lbID string, targets []string, pending []string, timeout time.Duration) error {
-	log.Printf("[DEBUG] Waiting for loadbalancer %s to become %+q.", lbID, targets)
+func waitForLBV2LoadBalancer(ctx context.Context, lbClient *gophercloud.ServiceClient, lbID string, target string, pending []string, timeout time.Duration) error {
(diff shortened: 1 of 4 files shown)


nikParasyr (approved):
LGTM

PR #1747, opened 2024-07-11
Upgrade to gophercloud v2

### LINKED ISSUE #1699: Upgrade to Gophercloud v2
Gophercloud is planning a v2 release.
Upgrade to it when possible.
As there will be breaking changes with v2, it is probably better to sync it with a major release

kayrus:
Update gophercloud to v2 

Resolves #1699


nikParasyr:
Ill check this one after we have agreed on how to proceed with the resource removal and whether we go for a major release or not

PR #1756, opened 2024-07-18
[core]: introduce "parsePairedIDs" helper

kayrus:
I noticed duplicate code, let's get rid of it.


nikParasyr (approved):
LGTM.

I think we can wait out a bit till devstack master is fixed ( although i dont expect any real issue on the workflows )

PR #1760, opened 2024-07-19
[compute]: allow metadata update along with the power_state

### LINKED ISSUE #1749: Cannot update metadata and shelf/offload at the same time.
### Terraform Version
Terraform 1.8.4

### Affected Resource(s)

- `openstack_compute_instance_v2`

```hcl
resource "openstack_compute_instance_v2" "vm" {
    name = "my_vm"
    image_id = "..."
    flavor_id = 1
    security_groups = ["default"]
    network {
        name = "my_network"
    }

    # change values below and re-apply
    power_state = "active" # change to "shelved_offloaded"
    metadata = {
        value = "abc" # change to "cdf"
    }
}
```
### Expected Behavior

If the resource's `power_state` changes from from "active" to "shelved_offloaded" and the metadata value changes (in the same apply), the metadata value should be updated before the instance in shelved.

### Actual Behavior
the provider tries to update the instance metadata value after the instance has entered the `shelved_offloaded` state, resulting in an error: 
```
{"conflictingRequest": {"code": 409, "message": "Cannot 'update metadata' instance [...] while it is in vm_state shelved_offloaded"}}
```

### Steps to Reproduce

1. apply the configuration above.
2. change the `power_state` shelved_offloaded and the `metadata` value to something else.
3. apply again.

kayrus:
Resolves #1749

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -798,82 +798,6 @@ func resourceComputeInstanceV2Update(ctx context.Context, d *schema.ResourceData
 		}
 	}
 
-	if d.HasChange("power_state") {
-		powerStateOldRaw, powerStateNewRaw := d.GetChange("power_state")
-		powerStateOld := powerStateOldRaw.(string)
-		powerStateNew := powerStateNewRaw.(string)
-		if strings.ToLower(powerStateNew) == "shelved_offloaded" {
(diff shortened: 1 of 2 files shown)


nikParasyr (approved):
LGTM

kayrus:
Let's hold with this for a while. Since once the VM is initially in the `shelved_offloaded` mode, its metadata also won't be updated.

PR #1761, opened 2024-07-19
[core]: add provider version to user-agent

### LINKED ISSUE #604: Prepend/append openstack provider version into the user agent
Not sure whether it is possible, because there is no release constant/variable available in openstack provider.

At the moment it is only:

```
User-Agent: Terraform/0.11.7 gophercloud/2.0.0
```

kayrus:
Resolves #604
Depends on #1759 

This change will build a user-agent header like:

`
User-Agent: HashiCorp Terraform/1.4.4 (+https://www.terraform.io) Terraform Plugin SDK/2.30.0 Terraform Provider OpenStack/2.0.0 gophercloud/v1.13.0
`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,7 +15,7 @@ builds:
     flags:
       - -trimpath
     ldflags:
-      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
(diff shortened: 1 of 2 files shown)


nikParasyr (approved):
LGTM

PR #1783, opened 2024-09-21
feat: Add loadbalancer cascade delete

neilfarmer:
Adding cascade deletion of loadbalancer. When a loadbalancer has resources attached to it such as, pools, listeners etc, the loadbalancer will fail to delete unless the cascade flag is set. The default for cascade deletion is false.


kayrus:
@neilfarmer thanks for the PR. Can you add functional tests to cover this feature?

kayrus:
@neilfarmer we're planning a new major release soon today. Are you willing to fix the tests?

neilfarmer:
@kayrus Id say go ahead with your release. I dont want to hold that up. I need to dig in more on golang/terraform tests. This is my first time with them.

PR #1790, opened 2024-09-25
Readme update

kayrus:
Add OpenTofu registry UI and remove mentioning the OpenLab and VEXXHOST, since we entirely moved to GitHub actions a while ago.


nikParasyr (approved):
LGTM

PR #1792, opened 2024-09-26
Add lb_flavorprofile_v2 data source

### LINKED ISSUE #1776: Feature request:  Support for Octavia flavorprofiles and flavors
I'd like to request a couple of new resources to provide for management of Octavia flavorprofiles and flavors.  This would allow us to manage these resources using TF.

Documentation is here: https://docs.openstack.org/octavia/latest/admin/flavors.html.

Something like:

```
resource "openstack_lb_flavorprofile_v2 "example" {
  name = "example-flavorprofile"
  provider = "amphora"
  flavor_data {
    loadbalancer_topology = "SINGLE"
    compute_flavor = "0022c049-3652-43c4-9e18-e9da6e92c1f"
  }

resource "openstack_lb_flavor_v2 "example" {
  name = "example-flavor"
  description = "An example Octavia flavor"
  flavorprofile = openstack_lb_flavorprofile_v2.example.id
  enable = true
}
```

nikParasyr:
Close #1776

### DIFF
--- a/openstack/data_source_openstack_lb_flavorprofile_v2.go
+++ b/openstack/data_source_openstack_lb_flavorprofile_v2.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/openstack/loadbalancer/v2/flavorprofiles"
-	"github.com/gophercloud/gophercloud/v2/pagination"
 )
 
 func dataSourceLBFlavorProfileV2() *schema.Resource {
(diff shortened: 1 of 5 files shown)


openstack/provider.go
```diff
@@ -312,6 +312,7 @@ func Provider() *schema.Provider {
 			"openstack_keymanager_secret_v1":                     dataSourceKeyManagerSecretV1(),
 			"openstack_keymanager_container_v1":                  dataSourceKeyManagerContainerV1(),
 			"openstack_loadbalancer_flavor_v2":                   dataSourceLBFlavorV2(),
+			"openstack_lb_flavorprofile_v2":                      dataSourceLBFlavorProfileV2(),
```
nikParasyr, line 315: This is a bit... annoying.
For all the resources we use `openstack_lb_xyz` but for the flavor data source the `openstack_loadbalancer_xyz` was used and i didn't catch it on time... (https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1651 https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1679 )
So there are 2 ways forward i'd say:
- we continue with `openstack_loadbalancer` for data sources and `openstack_lb` for resources which can be an annoying inconsistency
- we rename the `openstack_loadbalancer_flavor_v2` data source to `openstack_lb_flavor_v2` which will be a breaking change for users...

@kayrus this should be handled on a different PR but came across it while writing this one. Would like your opinion on how to go forward

kayrus: We can make an alias with the deprecation note and docs on how to move the state.

openstack/data_source_openstack_lb_flavorprofile_v2.go
```diff
@@ -0,0 +1,127 @@
...
+	return &schema.Resource{
+		ReadContext: dataSourceLBFlavorProfileV2Read,
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"id": {
```
kayrus, line 26: is it a valid argument? for other data sources we use `"%reource%_id"` key

openstack/data_source_openstack_lb_flavorprofile_v2.go
```diff
@@ -0,0 +1,127 @@
...
+
+	allfps, err = flavorprofiles.ExtractFlavorProfiles(allPages)
+	if err != nil {
+		return diag.Errorf("Unable to retrieve Openstack loadbalancer flavors: %s", err)
+	}
+
+	// Loop through all flavorprofiles to find a more specific one
+	if len(allfps) > 0 {
+		var filteredfps []flavorprofiles.FlavorProfile
+		name := d.Get("name").(string)
```
kayrus, line 92: to be honest I don't like an idea of local filtering. Let's implement `name` and `provider_name` in gophercloud and get back to this data source? https://github.com/gophercloud/gophercloud/pull/3193

kayrus (approved):
I applied fixes.

PR #1793, opened 2024-09-26
correctly handle already deleted resources on deletion

kayrus:
Ensure that resource was already deleted on deletion action


nikParasyr (approved):
LGTM

Issue #1796, opened 2024-09-27
"Rename" openstack_loadbalancer_flavor_v2 data source to openstack_lb_flavor_v2


nikParasyr:
the `openstack_loadbalancer_flavor_v2` data source is following the wrong naming convention. We should rename it to
`openstack_lb_flavor_v2`. In order to do it:
- [ ] create a new `openstack_lb_flavor_v2` data source using existing code
- [ ] deprecate `openstack_loadbalancer_flavor_v2` and add documentation on how to move to the new one
- [ ] on 4.0.0 remove deprecated  `openstack_loadbalancer_flavor_v2`

### References
- https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1651
- https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1679
- https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1792

PR #1797, opened 2024-09-27
Deprecate loadbalancer_flavor_v2 data source

nikParasyr:
loadbalancer_flavor_v2 data source should be deprecated in favor of lb_flavor_v2. Add deprecation message to loadbalancer_flavor_v2. Create lb_flavor_v2 data source re-using existing code.

Partially implements: #1796

@kayrus didnt add a guide as this is a data source so there is no need to import really


kayrus (approved):
LGTM, thanks!

PR #1798, opened 2024-09-27
[networking]: use rules.ProtocolAny for "any" secgroup protocol

kayrus:
cosmetic change. use `rules.ProtocolAny` instead of empty string.


kayrus:
linter suggests to keep this as is

PR #1802, opened 2024-10-07
fix: correct resource and floating_ip names in floating IP association

AmirAllahveran:
**Description:**
This pull request fixes an issue with the `resource` and `floating_ip` name in the `openstack_networking_floatingip_associate_v2` resource in the Terraform OpenStack provider `compute_instance_v2` documentation. The original configuration had the `resource name` and `floating_ip name` incorrectly swapped, which caused issues when associating the floating IP to the correct port.


kayrus (approved):
LGTM, thanks!

PR #1804, opened 2024-10-09
[core] override an endpoint with a new service name

### LINKED ISSUE #1803: [core] override an endpoint with a new service name
Current endpoint override logic allows to specify only static URLs. There should be a way to specify a new alias by a new service name. E.g.

```
provider "openstack" {

  endpoint_overrides = {
    "network"  = "network-beta"
    "volumev2" = "volumev3"
  }

}
```

This logic was already implemented in https://github.com/gophercloud/utils/pull/229 we need just to bump deps and adjust the docs.

kayrus:
Resolves #1803


kayrus:
The `volumev3` endpoint was removed: https://github.com/openstack/devstack/commit/d7c3c0accc89e4e99915c24fa7c3bff2e90a715e

PR #1806, opened 2024-10-10
fix(objectstorage-container-v1): set ForceNew to true for container name

### LINKED ISSUE #1769: Changing name of `openstack_objectstorage_container_v1` resource doesn't work
### Terraform Version
Terraform v1.9.3

### Affected Resource(s)
- openstack_objectstorage_container_v1

```hcl
resource "openstack_objectstorage_container_v1" "test_container" {
  name = "test-container"
}
```

###  Output
```
openstack_objectstorage_container_v1.test_container: Refreshing state... [id=test-container]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # openstack_objectstorage_container_v1.test_container will be updated in-place
  ~ resource "openstack_objectstorage_container_v1" "test_container" {
        id             = "test-container"
      ~ name           = "test-container" -> "test-container-renamed"
        # (4 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
openstack_objectstorage_container_v1.test_container: Modifying... [id=test-container]
openstack_objectstorage_container_v1.test_container: Modifications complete after 0s [id=test-container]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

### Expected Behavior
The container is renamed.

### Actual Behavior
The container is not renamed. Each run of `terraform apply` show a change that is not applied.

### Steps to Reproduce
1. `terraform apply`
2. change the value of `name` for resource `openstack_objectstorage_container_v1.test_container`
3. `terraform apply`

AmirAllahveran:
According to the openstack_objectstorage_container_v1 documentation for the name argument, changing the name should result in creating a new container. However, the existing behavior does not reflect this, as changes to the name attribute are not applied. This issue arises because the ForceNew property for the name field is set to false, preventing the expected recreation of the container when the name changes.

This update sets ForceNew to true for the name attribute, ensuring that any name change triggers the creation of a new container. It’s important to note that if there are objects present in the container, the deletion will not be allowed, which aligns with the standard behavior of OpenStack Object Storage to prevent data loss.

Resolves #1769

### DIFF
--- a/openstack/resource_openstack_objectstorage_container_v1.go
+++ b/openstack/resource_openstack_objectstorage_container_v1.go
@@ -46,7 +46,7 @@ func resourceObjectStorageContainerV1() *schema.Resource {
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: false,
+				ForceNew: true,
 			},
 			"container_read": {
 				Type:     schema.TypeString,


kayrus (approved):
Fair enough. Thank you!

kayrus:
Nevertheless I have some doubts. Renaming the container will cause a container removal. If it has a `force_destroy = true` flag, it will remove every object inside the container. I think that the `ForceNew: false` was set by purpose. @nikParasyr do you have some thoughts about this?

AmirAllahveran:
If `forceNew` is set to `true` for the container name and `force_destroy` is `false`, attempting to rename a container that contains objects results in the following error message:

```
<html><h1>Conflict</h1><p>There was a conflict when trying to complete your request.</p></html>
```

However, if the container is empty, applying the rename operation shows a message indicating success:

```
Apply complete! Resources: 0 added, 1 changed, 0 destroyed
```

Despite this message, the container name does not actually change. 

The Terraform logs provide the following warning:

```log
[WARN]  Provider "provider[\"registry.terraform.io/terraform-provider-openstack/openstack\"]" produced an unexpected new value for openstack_objectstorage_container_v1.test_container, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .name: was cty.StringVal("newtest"), but now cty.StringVal("oldtest")
```

The provider version being used is:

```
plugin=.terraform/providers/registry.terraform.io/terraform-provider-openstack/openstack/3.0.0/darwin_arm64/terraform-provider-openstack_v3.0.0 id=14574
```

Additionally, it appears that `force_destroy` and `versioning` settings no longer function correctly. Regardless of whether these options are set to `true` or `false`, `d.Get("force_destroy").(bool)` and `d.Get("versioning").(bool)` always return `false`. The Terraform logs provide a related warning:

```log
[WARN]  Provider "registry.terraform.io/terraform-provider-openstack/openstack" produced an invalid plan for openstack_objectstorage_container_v1.test_container, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .force_destroy: planned value cty.False for a non-computed attribute
      - .versioning: planned value cty.False for a non-computed attribute
```

PR #1821, opened 2024-11-11
Use Go version from the go.mod file

anthonyroussel:
I noticed that 2 GitHub Actions workflows were still using Go 1.20 for builds.

To address this, I propose updating the workflows to leverage the `go-version-file` property of the `actions/setup-go@v5` action. This property allows the action to automatically determine the Go version to use based on the version specified in the `go.mod` file.

This approach is recommended in the [Terraform Provider Scaffolding guide](https://github.com/hashicorp/terraform-provider-scaffolding/blob/main/.github/workflows/release.yml#L33).

Additionally, we can apply a similar improvement for golangci-lint by removing the  `.run.go` property. This will ensure that golangci-lint uses the Go version specified in the `go.mod` file, further aligning the versions across the project.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -4,22 +4,19 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        go-version:
-          - "1.22"
 
     env:
       GO111MODULE: "on"
 
     steps:
-      - name: Setup Go ${{ matrix.go-version }}
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Go
         uses: actions/setup-go@v5
         with:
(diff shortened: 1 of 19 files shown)


kayrus:
@anthonyroussel nice catch, thanks!

.github/workflows/unit-tests.yml
```diff
@@ -4,22 +4,19 @@ on: [push, pull_request]
...
-      matrix:
-        go-version:
-          - "1.22"
 
     env:
       GO111MODULE: "on"
 
     steps:
-      - name: Setup Go ${{ matrix.go-version }}
+      - name: Checkout
```
kayrus, line 12: what is the reason in order change?

anthonyroussel: The order needs to be changed because the `go.mod` file must be checked out from the Git repository so that `actions/setup-go` can determine the Go version to use.

nikParasyr (approved):
LGTM

PR #1835, opened 2025-02-21
Recognize 'CREATED' state for openstack_fw_group_v2 resource

Makaveli96bg:
To reslove issue - https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1834


Makaveli96bg:
this is to fic this issue - https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1834

kayrus:
as discussed in person, the issue relates to a neutron driver not following the API reference. The API ref has only:

> The status of the firewall group. Valid values are ACTIVE, INACTIVE, ERROR, PENDING_UPDATE, or PENDING_DELETE.

see https://docs.openstack.org/api-ref/network/v2/index.html#create-firewall-group

Issue #1745, opened 2024-07-09
Feature request: support for host or hypervisor_hostname arguments in openstack_compute_instance_v2


gespinozat:
Hi, 

Do you have any plans to support the `host` or `hypervisor_hostname` as arguments for instance creation?
Since the OpenStack Nova API microversion 2.74, you can provide a host or hypervisor_hostname when [creating an instance](https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-detail%2Ccreate-server-detail#create-server).

To support this feature some changes are required in gophercloud.
There is already an [open issue.](https://github.com/gophercloud/gophercloud/issues/2499)

### Affected Resource(s)
openstack_compute_instance_v2

ngin8-beta:
@kayrus 
When do you expect the release that includes this merge? I’m interested because I need to use this feature in my project.

kayrus:
@ngin8-beta probably next week.

ngin8-beta:
@kayrus 
I’ve been keeping an eye on the progress, but there doesn’t seem to be any recent activity. I’d like to start using this feature soon—would it be possible for you to release it? I’ve been using a self-built binary for now, but it’s a hassle to require my team members to do the same.

kayrus:
@ngin8-beta  apologies, I'm extremely busy this week

ngin8-beta:
@kayrus 
Are you going to be busy for a while longer?

ngin8-beta:
@kayrus 
I think it's okay to close this issue.

PR #1837, opened 2025-03-05
 Add `hypervisor_hostname` Option to Compute Instance

ngin8-beta:
Issue: #1745

This pull request adds the ability to specify `hypervisor_hostname` during Compute Instance creation in order to support OpenStack Nova API microversion >2.74. 
  
change :
        - Added `hypervisor_hostname` to the resource schema.
        - Introduced an acceptance test to verify that the `hypervisor_hostname` parameter is correctly processed.

Ad-hoc Test Results:
In our local OpenStack (DevStack) environment, we confirmed that:

   - The `hypervisor_hostname` is correctly included in the API request during Compute Instance creation.
   - There are no adverse effects on existing functionalities (e.g., instance creation and updates).

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -179,10 +179,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 				},
 			},
 			"hypervisor_hostname": {
-				Type:     schema.TypeString,
-				Computed: true,
-				Optional: true,
-				ForceNew: true,
+				Type:          schema.TypeString,
+				Computed:      true,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"personality"},
 			},
 			"metadata": {
 				Type:     schema.TypeMap,
@@ -350,7 +351,8 @@ func resourceComputeInstanceV2() *schema.Resource {
 						},
 					},
 				},
-				Set: resourceComputeInstancePersonalityHash,
+				Set:           resourceComputeInstancePersonalityHash,
+				ConflictsWith: []string{"hypervisor_hostname"},
 			},
 			"stop_before_destroy": {
 				Type:     schema.TypeBool,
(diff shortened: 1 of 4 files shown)


ngin8-beta:
@nikParasyr @kayrus
I submitted a pull request, but it failed the acceptance tests. Should I investigate and resolve these errors? Please let me know what actions I need to take for it to be merged.

kayrus:
@ngin8-beta thanks for the PR. Unfortunately it is not complete, you just declared a new `hypervisor_hostname` argument, but it is not used anywhere else. You need to pass it to the corresponding argument to create opts.
Also you need to bump gophercloud to include the https://github.com/gophercloud/gophercloud/pull/3309 PR.
And please make corresponding documentation adjustments.

ngin8-beta:
@kayrus 
I have made revisions to the points you mentioned, as best as I could. How does it look?

kayrus:
Thanks for the update. I noticed that API reference mentions that the microservice API version must be 2.74. https://docs.openstack.org/api-ref/compute/#create-server
e.g. https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/5d9a8714a39f7910988580e5a9a82ab11c24bbf9/openstack/resource_openstack_compute_instance_v2.go#L477

ngin8-beta:
OK. It should include processing to return an error if the version is below v2.74, right? Is there a function provided for checking the API version?

ngin8-beta:
I tried specifying the microservice API version.

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -178,6 +178,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 					},
 				},
 			},
+			"hypervisor_hostname": {
+				Type:     schema.TypeString,
+				Optional: true,
```
kayrus, line 183: you need to add `Computed: true`, otherwise servers, which were created in previous provider versions will be forced to be recreated.

ngin8-beta:
I have made code modifications based on your review.

ngin8-beta:
I investigated an error that occurred during test execution.
- TestAccComputeInstanceV2_hypervisorHostname
```
Compute host my-hypervisor could not be found.
```
This error likely occurs because the hostname of the DevStack itself needs to be specified. Do you know what the default hostname of DevStack is? Or is it necessary to explicitly specify the hostname?
I am not very familiar with automated testing.

ngin8-beta:
After deploying DevStack, I may need to retrieve the hypervisor_hostname information. However, I have not yet found an effective method to achieve this.

kayrus (approved):

PR #1838, opened 2025-03-06
Fix all-in quota creation for resource_openstack_blockstorage_quotaset_v3

StopMotionCuber:
This is a copy of #1316 for the blockstorage quotaset, as the issue is also present there (right now quotas are set to 0 if not set. This is documented, but not intuitive behavior). The fix seems straightforward.

There are 4 quota resources for which this behavior is different:
 - `resource_openstack_blockstorage_quotaset_v3`: Fixed with this PR
 - `resource_openstack_compute_quotaset_v2`: Fixed with #1304, however the [docs](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/docs/resources/compute_quotaset_v2.md?plain=1#L19) have not been updated for that
 - `openstack_networking_quota_v2`: Fixed with #1316, however the [docs](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/docs/resources/networking_quota_v2.md?plain=1#L19) have not been updated for that
 - `resource_openstack_lb_quota_v2`: Has not been fixed (still sets values on the API to 0), however I'm not planning on using that personally so I don't care too much about this [reference](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_lb_quota_v2.go#L106). If wanted, I could extend the scope of this PR to also include a fix on that quota

So a few questions:
1. Is this seen as a bug, such that an update is wanted?
2. Should I extend the scope of the PR to include removing the invalid notes on `resource_openstack_compute_quotaset_v2` and `openstack_networking_quota_v2`?
3. Should I extend the scope of the PR to include a fix for `resource_openstack_lb_quota_v2`?

### DIFF
--- a/openstack/resource_openstack_blockstorage_quotaset_v3.go
+++ b/openstack/resource_openstack_blockstorage_quotaset_v3.go
@@ -103,7 +103,7 @@ func resourceBlockStorageQuotasetV3Create(ctx context.Context, d *schema.Resourc
 
 	updateOpts := quotasets.UpdateOpts{}
 	projectID := d.Get("project_id").(string)
-	if v, ok := getOkExists(d, "fixed_ips"); ok {
+	if v, ok := getOkExists(d, "volumes"); ok {
 		value := v.(int)
 		updateOpts.Volumes = &value
 	}
--- a/openstack/resource_openstack_blockstorage_quotaset_v3.go
+++ b/openstack/resource_openstack_blockstorage_quotaset_v3.go
@@ -101,30 +101,43 @@ func resourceBlockStorageQuotasetV3Create(ctx context.Context, d *schema.Resourc
 		return diag.Errorf("Error creating OpenStack block storage client: %s", err)
 	}
 
+	updateOpts := quotasets.UpdateOpts{}
 	projectID := d.Get("project_id").(string)
-	volumes := d.Get("volumes").(int)
-	snapshots := d.Get("snapshots").(int)
-	gigabytes := d.Get("gigabytes").(int)
-	perVolumeGigabytes := d.Get("per_volume_gigabytes").(int)
-	backups := d.Get("backups").(int)
-	backupGigabytes := d.Get("backup_gigabytes").(int)
-	groups := d.Get("groups").(int)
+	if v, ok := getOkExists(d, "fixed_ips"); ok {
+		value := v.(int)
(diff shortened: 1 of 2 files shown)


kayrus:
@StopMotionCuber please rebase on main

StopMotionCuber:
So, seems like acceptance tests are failing, but my knowledge about the testing is low and I'm not too eager to scan through 30k lines of output to find those lines which are relevant.

@kayrus would be great if you could determine whether the failures are related to my PR or if it's just flaky tests

kayrus:
@StopMotionCuber 

>  Error creating openstack_identity_project_v3: Expected HTTP response code [201 202] when accessing [POST http://10.1.0.9/identity/v3/projects], but got 409 instead: {"error":{"code":409,"message":"Conflict occurred attempting to store project - it is not permitted to have two projects with either the same name or same id in the same domain: name is project_1, project id fa917ab54ebe499494801c279d316df8.","title":"Conflict"}}

openstack/resource_openstack_blockstorage_quotaset_v3.go
```diff
@@ -101,30 +101,43 @@ func resourceBlockStorageQuotasetV3Create(ctx context.Context, d *schema.Resourc
...
+	updateOpts := quotasets.UpdateOpts{}
 	projectID := d.Get("project_id").(string)
-	volumes := d.Get("volumes").(int)
-	snapshots := d.Get("snapshots").(int)
-	gigabytes := d.Get("gigabytes").(int)
-	perVolumeGigabytes := d.Get("per_volume_gigabytes").(int)
-	backups := d.Get("backups").(int)
-	backupGigabytes := d.Get("backup_gigabytes").(int)
-	groups := d.Get("groups").(int)
+	if v, ok := getOkExists(d, "fixed_ips"); ok {
```
kayrus, line 106: You have a typo here
```suggestion
	if v, ok := getOkExists(d, "volumes"); ok {
```

kayrus (approved):

openstack/resource_openstack_blockstorage_quotaset_v3.go
StopMotionCuber: Right... Thanks for spotting and fixing it

PR #1842, opened 2025-03-11
[compute] mark hostname argument optional in hypervisor data source

kayrus:
this should a allow to test `openstack_compute_hypervisor_v2` data source with devstack, when there is only one hypervizor available.


kayrus:
The test against the master devstack branch is failing because of the #1841 
This PR can be merged to avoid unnecessary tests

PR #1860, opened 2025-05-07
fix(openstack_blockstorage_volume_v3) : Allow image_id to be imported when other ids are not present

### LINKED ISSUE #1265: import openstack_blockstorage_volume_v3 does not import image property
### Terraform Version
Terraform v1.0.0
on linux_amd64
+ provider registry.terraform.io/hashicorp/template v2.2.0
+ provider registry.terraform.io/terraform-provider-openstack/openstack v1.42.0

### Affected Resource(s)
- openstack_blockstorage_volume_v3

```hcl
terraform import openstack_blockstorage_volume_v3.bn-p-monitor302-boot 90d5bff0-7fa5-42ab-8f78-b8dbb493a75e

terraform state show openstack_blockstorage_volume_v3.bn-p-monitor302-boot                                                                                                                                                                  
# openstack_blockstorage_volume_v3.bn-p-monitor302-boot:
resource "openstack_blockstorage_volume_v3" "bn-p-monitor302-boot" {
    attachment        = [
        {
            device      = "/dev/vda"
            id          = "90d5bff0-7fa5-42ab-8f78-b8dbb493a75e"
            instance_id = "beb35491-92a2-489b-92ab-55de25645c5e"
        },
    ]
    availability_zone = "nova"
    id                = "90d5bff0-7fa5-42ab-8f78-b8dbb493a75e"
    metadata          = {}
    name              = "bn-p-monitor302-boot"
    region            = "Sto2"
    size              = 100
    volume_type       = "ceph_hdd"

    timeouts {}
}

(openstack) volume show -f yaml bn-p-monitor302-boot
attachments:
- attached_at: '2021-03-07T21:31:56.000000'
  attachment_id: 728a4e92-8793-4fbf-a9ec-482ac36c13ed
  device: /dev/vda
  host_name: null
  id: 90d5bff0-7fa5-42ab-8f78-b8dbb493a75e
  server_id: beb35491-92a2-489b-92ab-55de25645c5e
  volume_id: 90d5bff0-7fa5-42ab-8f78-b8dbb493a75e
availability_zone: nova
bootable: 'true'
consistencygroup_id: null
created_at: '2020-06-15T13:42:07.000000'
description: null
encrypted: false
id: 90d5bff0-7fa5-42ab-8f78-b8dbb493a75e
multiattach: false
name: bn-p-monitor302-boot
os-vol-tenant-attr:tenant_id: 94d681cda9614469969df100d1317e33
properties: {}
replication_status: null
size: 100
snapshot_id: null
source_volid: null
status: in-use
type: ceph_hdd
updated_at: '2021-03-07T21:33:13.000000'
user_id: 3dc521e219804df59a2d4f4e08cdbe32
volume_image_metadata:
  checksum: d89eb49f2c264d29225cecf2b6c83322
  container_format: bare
  disk_format: qcow2
  image_id: 0ef89c1e-1a83-4e8e-98c5-34ed8a841881
  image_name: CentOS-8-GenericCloud-8.1.1911-20200113.3.x86_64
  min_disk: '0'
  min_ram: '0'
  signature_verified: 'False'
  size: '716176896'
```

### Expected Behavior
Expected image to be imported else terraform will want to recreate the resource when adding an image to the volume.

### Actual Behavior
image_id is empty after the import and terraform wants to recreate the volyme as it thinks a new image_id have been added (from blank to 0ef89c1e-1a83-4e8e-98c5-34ed8a841881)

### Steps to Reproduce
1. `create a volume
resource "openstack_blockstorage_volume_v3" "bn-p-monitor302-boot" {
  name                 = "bn-p-monitor302-boot"
  size                    = 1
  image_id            = "0ef89c1e-1a83-4e8e-98c5-34ed8a841881"
}`
2. `terraform apply`
3. `terraform state rm openstack_blockstorage_volume_v3.bn-p-monitor302-boot`
4. `terraform import openstack_blockstorage_volume_v3.bn-p-monitor302-boot VOLUMEID`
5. `terraform apply`

jsittaro:
This should fix the problem with image_id that was not imported into the state by adding it in the importing process. But this comes with two catches:

- The first one happens when image_id and at other id is present at the same time because of OpenStack keeping ``ImageMetadata`` when using a snapshot or another volume as base, therefore making a conflict who the two ids. To make sure that they are not conflicting each other, image_id is only set if the other ids are not present.
- As of today, OpenStack are accepting an image name on image_id field as there are resolving it internally[1]. But in this scenario, as the import function imports the ID and not the name in ``image_id`` field, Terraform will try to force-recreate the volume when applying. For a workaround, user should be using directly a image_id or get it from ``openstack_images_image_v2`` datasource.

As this is my first contribution, I am not sure if this is the correct solution, so I am open to comment if this PR needs improvement.

Fixes #1265

[1]: https://github.com/openstack/cinder/blob/635bdab2c0d44e95e499a1d13296446171d5e4ef/cinder/api/v3/volumes.py#L364 and https://github.com/openstack/cinder/blob/635bdab2c0d44e95e499a1d13296446171d5e4ef/cinder/api/v2/volumes.py#L136

### DIFF
--- a/openstack/resource_openstack_blockstorage_volume_v3.go
+++ b/openstack/resource_openstack_blockstorage_volume_v3.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"net/http"
 	"time"
@@ -23,7 +24,7 @@ func resourceBlockStorageVolumeV3() *schema.Resource {
 		UpdateContext: resourceBlockStorageVolumeV3Update,
 		DeleteContext: resourceBlockStorageVolumeV3Delete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: resourceBlockStorageVolumeV3Import,
 		},
 
 		Timeouts: &schema.ResourceTimeout{
@@ -283,16 +284,6 @@ func resourceBlockStorageVolumeV3Read(ctx context.Context, d *schema.ResourceDat
 		d.Set("volume_retype_policy", "never")
 	}
 
-	// As we can have a snapshot_id with an image_id if the VolumeImageMetadata is stored in the snapshot for exemple,
-	// we want to exclude image_id when one of the others ids are used
-	_, snapshotExists := d.GetOk("snapshot_id")
-	_, backupExists := d.GetOk("backup_id")
-	_, sourcevolExists := d.GetOk("source_vol_id")
-
-	if !snapshotExists && !backupExists && !sourcevolExists {
-		d.Set("image_id", v.VolumeImageMetadata["image_id"])
-	}
-
 	attachments := flattenBlockStorageVolumeV3Attachments(v.Attachments)
 	log.Printf("[DEBUG] openstack_blockstorage_volume_v3 %s attachments: %#v", d.Id(), attachments)
 	if err := d.Set("attachment", attachments); err != nil {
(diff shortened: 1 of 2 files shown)


openstack/resource_openstack_blockstorage_volume_v3.go
```diff
@@ -283,6 +283,16 @@ func resourceBlockStorageVolumeV3Read(ctx context.Context, d *schema.ResourceDat
...
 	}
 
+	// As we can have a snapshot_id with an image_id if the VolumeImageMetadata is stored in the snapshot for exemple,
+	// we want to exclude image_id when one of the others ids are used
+	_, snapshotExists := d.GetOk("snapshot_id")
+	_, backupExists := d.GetOk("backup_id")
+	_, sourcevolExists := d.GetOk("source_vol_id")
+
+	if !snapshotExists && !backupExists && !sourcevolExists {
+		d.Set("image_id", v.VolumeImageMetadata["image_id"])
```
kayrus, line 293: does it makes sense always set this attribute without any condition?

jsittaro: Hello @kayrus and thanks for the review. Also, sorry for the delay of the response, I was a bit occupied at my work.

If I understand your question, you ask why not set ``image_id`` all time without any condition. If this is the question, this is not possible for one reason: ``image_id``, ``snapshot_id``, ``backup_id`` and ``source_vol_id`` is excluded each other (``ConflictWith`` statement on the schema) and two ids may be present both in some circumstance.
For example, when you make a snapshot of a volume that was originally created with an image and someone decides to create a new volume with that snapshot, ``ImageMetadataVolume`` is carried over to the new volume and we got ``image_id`` and ``snapshot_id`` as the same time.
In this case:
    - Terraform will try to replace my volume because image_id is not set in my configuration and the plan wants to remove it therefore forcing recreation of the volume.
    - If I add ``image_id`` to match the volume on OpenStack and the state on Terraform, Terraform will disallow me to do that because now I have two values that are conflicting each other.
So, due to that, ``image_id`` need to be behind a condition to make sure that image ID is out of way when Terraform importing or refreshing it into the state when the volume is based on a snapshot, backup or a source volume.

kayrus: I guess terraform will recreate the resource even with the condition.
I suppose you need this for an import only, right? Then let's adjust the import function only and leave the read function as is.

jsittaro:
Hello,
I have moved this into the import function only and changed the description to match the current state of this PR. 
On the way, I added a import test to ensure that image_id is present when it should be here and slightly changed the condition to pass the basic import test as there were falling when image_id was present with empty string.

kayrus (approved):
LGTM, thanks!

PR #1912, opened 2025-06-25
core: bump gophercloud

kayrus:


kayrus:
octavia master branch is broken, thus we can ignore it.

Issue #922, opened 2019-11-04
Networking V2: Several issues with a router resource


kayrus:
### Affected Resource(s)
Please list the resources as a list, for example:
- openstack_networking_router_v2

### Actual Behavior

* impossible to unset the `external_fixed_ip`, since the `external_network_id` field has to be also unset, but it is computed.
* impossible to change or set `enable_snat` to false during the router creation, since it requires the **ext-gw-mode** extension to be enabled.
* impossible to set only the `ip_address`  without specifying the `subnet_id`
* impossible to change the `external_fixed_ip` on older neutron versions (upstream neutron already has this fix: https://bugs.launchpad.net/neutron/+bug/1812118)

### Important Factoids

see #449 and #628

PR #1915, opened 2025-06-27
make subnet_id in openstack_networking_router_v2 optional

Hoeze:
I want to fix the issue that it is impossible to set only the ip_address without specifying the subnet_id: #922 .

My solution just sets the `subnet_id` to be `computed`.
Inspired by https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/628.

Unfortunately, I do not know what "fixed_ip" I could use in the tests.
Also, I have no idea of how to test this locally with my terraform installation, because terraform strictly refuses to read my self-built binary with "Error: Required plugins are not installed" and some checksum warning.


Hoeze:
Amazing, thanks a lot for picking this up @kayrus!

PR #1925, opened 2025-07-10
Handle when tenant/project already has access to flavor

hloeung:
Fixes https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1811

### DIFF
--- a/openstack/resource_openstack_compute_flavor_access_v2.go
+++ b/openstack/resource_openstack_compute_flavor_access_v2.go
@@ -62,7 +62,11 @@ func resourceComputeFlavorAccessV2Create(ctx context.Context, d *schema.Resource
 	log.Printf("[DEBUG] Flavor Access Options: %#v", accessOpts)
 
 	if _, err := flavors.AddAccess(ctx, computeClient, flavorID, accessOpts).Extract(); err != nil {
-		return diag.Errorf("Error adding access to tenant %s for flavor %s: %s", tenantID, flavorID, err)
+		if !gophercloud.ResponseCodeIs(err, http.StatusConflict) {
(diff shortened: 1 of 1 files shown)


openstack/resource_openstack_compute_flavor_access_v2.go
```diff
@@ -62,12 +62,16 @@ func resourceComputeFlavorAccessV2Create(ctx context.Context, d *schema.Resource
...
+		}
+	} else {
+		id := fmt.Sprintf("%s/%s", flavorID, tenantID)
+		d.SetId(id)
 	}
 
-	id := fmt.Sprintf("%s/%s", flavorID, tenantID)
-	d.SetId(id)
-
 	return resourceComputeFlavorAccessV2Read(ctx, d, meta)
```
kayrus, line 75:
```suggestion
		if !gophercloud.ResponseCodeIs(err, http.StatusConflict) {
			return diag.Errorf("Error adding access to tenant %s for flavor %s: %s", tenantID, flavorID, err)
		}
		log.Printf("[DEBUG] Tenant %s already has access to flavor %s: %s", tenantID, flavorID, err)
	}

	id := fmt.Sprintf("%s/%s", flavorID, tenantID)
	d.SetId(id)

	return resourceComputeFlavorAccessV2Read(ctx, d, meta)
```

kayrus (commented):
@hloeung thanks for the PR. Please try to avoid if/else conditions when possible.

openstack/resource_openstack_compute_flavor_access_v2.go
hloeung: Thanks for the review, was mostly copying from https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1919

I've pushed up a fix using the suggestions you have made.

PR #1969, opened 2025-10-01
Add fixed_ip block support in networking_port_v2 read/import

### LINKED ISSUE #1959: openstack_networking_port_v2 doesn't import fixed_ip
### Terraform Version
Terraform v1.13.3
on linux_amd64
+ provider registry.terraform.io/terraform-provider-openstack/openstack v3.3.2

### Affected Resource(s)

### Expected Behavior
importing fixed_ip blocks

### Actual Behavior
no fixed_ip blocks imported

### Steps to Reproduce
just run the import example defined in https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/networking_port_v2

Koodt:
Set fixed_ip in resourceNetworkingPortV2Read
Remove ImportStateVerifyIgnore for fixed_ip in import test
Mark subnet_id and ip_address fields as computed: true

Close #1959

### DIFF
--- a/openstack/import_openstack_networking_port_v2_test.go
+++ b/openstack/import_openstack_networking_port_v2_test.go
@@ -25,9 +25,6 @@ func TestAccNetworkingV2Port_importBasic(t *testing.T) {
 				ResourceName:      resourceName,
 				ImportState:       true,
 				ImportStateVerify: true,
-				ImportStateVerifyIgnore: []string{
-					"fixed_ip",
-				},
 			},
 		},
 	})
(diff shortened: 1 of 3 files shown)


Koodt:
That’s a fair point. In that case, maybe we should consider deprecating the all_fixed_ips field in future releases. Since ip_address and subnet_id are tightly coupled and both belong to a single fixed_ip, we could introduce a new field now that explicitly represents this structure.

For naming, something like fixed_ips (without the all_ prefix) would be clearer and closer to how Neutron/OpenStack itself models it.

kayrus:
`all_fixed_ips` is used as a computed field. `fixed_ips` may be used as a filter for the data source and as an argument for the resource.

Koodt:
I agree with this approach and support it. In the future, all_fixed_ips can be removed because fixed_ips will be able to replace its functionality.
Do you want me to move the proposed changes into fixed_ips now?

PR #1970, opened 2025-10-03
Add data source for BlockStorage Volume Type

Koodt:
Add data_source_openstack_blockstorage_volume_type_v3 Add data_source_openstack_blockstorage_volume_type_v3_test Add docs

Close #1963

### DIFF
--- a/openstack/data_source_openstack_blockstorage_volume_type_v3.go
+++ b/openstack/data_source_openstack_blockstorage_volume_type_v3.go
@@ -1,0 +1,129 @@
+package openstack
+
+import (
+	"context"
+
+	"github.com/gophercloud/gophercloud/v2/openstack/blockstorage/v3/volumetypes"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceBlockStorageVolumeTypeV3() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceBlockStorageVolumeTypeV3Read,
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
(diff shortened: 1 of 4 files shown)


openstack/data_source_openstack_blockstorage_volume_type_v3.go
```diff
@@ -0,0 +1,153 @@
...
+
+func dataSourceBlockStorageVolumeTypeV3Read(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	config := meta.(*Config)
+
+	client, err := config.BlockStorageV3Client(ctx, GetRegion(d, config))
+	if err != nil {
+		return diag.Errorf("Error creating OpenStack block storage client: %s", err)
+	}
+
+	listOpts := volumetypes.ListOpts{}
```
kayrus, line 67: I was able to list volume types by name using curl. Therefore it seems that name filtering is hidden somewhere in the cinder source code and it's worth adding it into the gophercloud.

openstack/data_source_openstack_blockstorage_volume_type_v3.go
```diff
@@ -0,0 +1,153 @@
...
+		for _, vt := range allVolumeTypes {
+			if vt.Name == name.(string) {
+				filtered = append(filtered, vt)
+			}
+		}
+
+		allVolumeTypes = filtered
+	}
+
+	if filter, ok := d.GetOk("extra_specs"); ok {
```
kayrus, line 103: it is also possible to list volume types by extra specs, e.g. for `storage_protocol=nfs` you use this filter in curl: `types?extra_specs=%7B%27storage_protocol%27%3A+%27nfs%27%7D`

kayrus (commented):
would be nice to add extra filters support in gophercloud to avoid redundant client-side filtering.

openstack/data_source_openstack_blockstorage_volume_type_v3.go
Koodt: I’m adding support for this to gophercloud as well

kayrus: the corresponding change was merged, you can now implement the proper API request filter

Issue #1818, opened 2024-11-11
Add support for Mistral workflow resources and datasources


anthonyroussel:
## Global

* [ ] Add Github Action workflow for Mistral (#1813)

## Resources

* [ ] openstack_workflow_workflow_v2
* [ ] openstack_workflow_cron_trigger_v2
* [ ] openstack_workflow_execution_v2

## Datasources

* [ ] openstack_workflow_workflow_v2 (#1813)
* [ ] openstack_workflow_cron_trigger_v2
* [ ] openstack_workflow_execution_v2

PR #1996, opened 2025-12-14
[workflow] add openstack_workflow_cron_trigger_v2 resource

anthonyroussel:
This introduce the `openstack_workflow_cron_trigger_v2` resource, which allows users to manage Mistral cron triggers in OpenStack.

See issue #1818

### DIFF
--- a/openstack/resource_openstack_workflow_cron_trigger_v2.go
+++ b/openstack/resource_openstack_workflow_cron_trigger_v2.go
@@ -1,0 +1,150 @@
+package openstack
+
+import (
+	"context"
+	"time"
+
+	"github.com/gophercloud/gophercloud/v2/openstack/workflow/v2/crontriggers"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceWorkflowCronTriggerV2() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceWorkflowCronTriggerV2Create,
+		ReadContext:   resourceWorkflowCronTriggerV2Read,
+		DeleteContext: resourceWorkflowCronTriggerV2Delete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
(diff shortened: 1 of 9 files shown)


kayrus:
@anthonyroussel could you please rebase and fix the tests?

anthonyroussel:
@kayrus Hello, I've rebased my PR, fixed the Mistral tests, and they are now passing.

openstack/resource_openstack_workflow_cron_trigger_v2.go
```diff
@@ -0,0 +1,150 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+
+			"name": {
```
kayrus, line 30: I see that all arguments have `ForceNew: true`. I'm aware that gophercloud doesn't have `UpdateOpts`, but according to API ref, it should be supported https://docs.openstack.org/mistral/latest/user/rest_api_v2.html#post--v2-cron_triggers

Would you like to test POST method and if it works, introduce `UpdateOpts` in gophercloud and update the PR? Or would you like to do this in a separate PR?

anthonyroussel: @kayrus Thank you for the review!

Unfortunately, updating an existing cron trigger is not supported by the Mistral API.

The API only exposes `POST /v2/cron_triggers`, which is used to **create new cron triggers**. There is no `PUT /v2/cron_triggers/:id` endpoint (or equivalent) to update an existing one, as confirmed by both the documentation and the code.

To be sure, I also tested updating a cron trigger with the `POST /v2/cron_triggers` endpoint against an OpenStack Mistral server: when a cron trigger with the same name already exists, the API returns an HTTP 409 Conflict instead of performing an update:

```shell
$ curl -g -i -X POST https://[mistral-api]/v2/cron_triggers -H "X-Auth-Token: [token]" -H "content-type: application/json" -d '{"name": "my_cron_trigger", "pattern": "0 2 * * *", "workflow_name": "test", "workflow_input": "{\"message\": \"test\"}"}'
HTTP/2 409 
content-type: application/json

{
  "faultcode": "Client",
  "faultstring": "Duplicate entry for cron trigger ['name', 'project_id']: test, None",
  "debuginfo": null
}
```

Sources:
* [Mistral API Cron Triggers v2 controller code](https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/cron_trigger.py)
* [Mistral API Cron Trigger v2 API](https://docs.openstack.org/mistral/latest/user/rest_api_v2.html#post--v2-cron_triggers)

kayrus: > There is no `PUT /v2/cron_triggers/:id` endpoint

you're right, thanks for clarifying.

kayrus (approved):
LGTM, thanks!

PR #2008, opened 2026-01-14
Improve compute resource  micro version handling

### LINKED ISSUE #2007: openstack_compute_instance_v2: block_device_mapping.volume_type conflict with hypervisor_hostname
### Affected Resource(s)

```hcl
resource "openstack_compute_instance_v2" "basic" {
  name            = "basic"
  image_id        = "ad091b52-742f-469e-8f3c-fd81cadf0743"
  flavor_id       = "3"
  key_pair        = "my_key_pair_name"
  security_groups = ["default"]

  block_device {
    boot_index            = 0
    delete_on_termination = true
    destination_type      = "local"
    source_type           = "image"
    uuid                       =  "<image-id>"
    volume_type         = "<volume-type>
  }

 hypervisor_hostname = "<hypervisor-hostname>"
}
```

### Expected Behavior
What should have happened?

Create an instance with the desired block-device-mapping and on the desired hypervisor-host.

### Actual Behavior
What actually happened?

The microversion is first set to the correct one for the `hypervisor_hostname` [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_compute_instance_v2.go#L513), before then being downgraded for `volume_type` [here](https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/openstack/resource_openstack_compute_instance_v2.go#L556)

### Steps to Reproduce
1. `terraform apply`

toanju:
The current micro version handling can downgrade the micro version to a
lower than required version. This addresses the issue in
`resource_openstack_compute_instance_v2`

fixes #2007

### DIFF
--- a/openstack/resource_openstack_compute_instance_v2.go
+++ b/openstack/resource_openstack_compute_instance_v2.go
@@ -483,8 +483,9 @@ func resourceComputeInstanceV2Create(ctx context.Context, d *schema.ResourceData
 
 	if networkMode := d.Get("network_mode").(string); networkMode == "auto" || networkMode == "none" {
 		// Use special string for network option
-		computeClient.Microversion = computeV2InstanceCreateServerWithNetworkModeMicroversion
 		networks = networkMode
+
+		bumpClientMicroversion(computeClient, computeV2InstanceCreateServerWithNetworkModeMicroversion)
 		log.Printf("[DEBUG] Create with network options %s", networks)
 	} else {
 		log.Printf("[DEBUG] Create with specified network options")
@@ -504,13 +505,14 @@ func resourceComputeInstanceV2Create(ctx context.Context, d *schema.ResourceData
 	// Retrieve tags and set microversion if they're provided.
 	instanceTags := computeV2InstanceTags(d)
 	if len(instanceTags) > 0 {
-		computeClient.Microversion = computeV2InstanceCreateServerWithTagsMicroversion
+		bumpClientMicroversion(computeClient, computeV2InstanceCreateServerWithTagsMicroversion)
 	}
 
 	var hypervisorHostname string
 	if v, ok := getOkExists(d, "hypervisor_hostname"); ok {
 		hypervisorHostname = v.(string)
-		computeClient.Microversion = computeV2InstanceCreateServerWithHypervisorHostnameMicroversion
+
+		bumpClientMicroversion(computeClient, computeV2InstanceCreateServerWithHypervisorHostnameMicroversion)
 	}
 
 	if v, ok := getOkExists(d, "availability_zone"); ok {
@@ -545,7 +547,7 @@ func resourceComputeInstanceV2Create(ctx context.Context, d *schema.ResourceData
 		// If so, set the client's microversion appropriately.
 		for _, bd := range d.Get("block_device").([]any) {
 			if bd.(map[string]any)["multiattach"].(bool) {
-				computeClient.Microversion = computeV2InstanceBlockDeviceMultiattachMicroversion
+				bumpClientMicroversion(computeClient, computeV2InstanceBlockDeviceMultiattachMicroversion)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1753,3 +1757,18 @@ func suppressPowerStateDiffs(_, old, _ string, _ *schema.ResourceData) bool {
 
 	return false
 }
+
+// update the computeClient microversion properly by checking if it already was set.
+// If it was set only update it when the currently set microversion is less than the required one.
+func setComputeClientMicroversion(computeClient *gophercloud.ServiceClient, requiredMicroversion string) {
```
kayrus: there is a similar func in https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/9c43c7e24fb98bc3226fd311eaa18a3c1df35276/openstack/util.go#L227-L229

Also gophercloud has a [ParseMicroversion](https://github.com/gophercloud/gophercloud/blob/e4836cad60303d9a69a85fa93f5be1f612761554/openstack/utils/discovery.go#L343-L357) func. Maybe we can reuse it?

toanju: updated `compatibleMicroversion` looks sufficient for this. Thanks for the pointer

openstack/resource_openstack_compute_instance_v2.go
```diff
@@ -1753,3 +1756,16 @@ func suppressPowerStateDiffs(_, old, _ string, _ *schema.ResourceData) bool {
...
+	if computeClient.Microversion == "" {
+		computeClient.Microversion = requiredMicroversion
+	} else {
+		compatible, _ := compatibleMicroversion("max", requiredMicroversion, computeClient.Microversion)
+		if compatible {
+			computeClient.Microversion = requiredMicroversion
+		}
+	}
```
kayrus:
```suggestion
		computeClient.Microversion = requiredMicroversion
		return
	}

	compatible, _ := compatibleMicroversion("max", requiredMicroversion, computeClient.Microversion)
	if compatible {
		computeClient.Microversion = requiredMicroversion
	}
```

openstack/resource_openstack_compute_instance_v2.go
kayrus: this func can be moved to utils.go

openstack/resource_openstack_compute_instance_v2.go
toanju: updated, added one more newline to make the linter happy

openstack/resource_openstack_compute_instance_v2.go
toanju: moved, thanks

openstack/util.go
```diff
@@ -367,3 +367,18 @@ func stringSliceToSet(in []string) *schema.Set {
 
 	return schema.NewSet(schema.HashString, result)
 }
+
+// update the computeClient microversion properly by checking if it already was set.
+// If it was set only update it when the currently set microversion is less than the required one.
+func setComputeClientMicroversion(computeClient *gophercloud.ServiceClient, requiredMicroversion string) {
```
kayrus: I think it would make more sense to rename this func to `bumpClientMicroversion`, and to rename the `computeClient` variable to just `client`.

Looking ahead, we may want to reuse this function in other resources. However, I noticed that in other parts of the codebase there's logic to handle microversion downgrades during fallback. Because of that, I'm not sure whether introducing a new function specifically for downgrading the version is necessary. What do you think?

toanju: Updated the naming.

Regarding the fallback handling I'd be generally try the correct micro version first and then fall back to `""`. As described in #2007 we set the micro version twice but overwrite it with the wrong as we downgrade it. Hence I somewhat would see this function as necessary.

kayrus (approved):
LGTM, thanks!

openstack/util.go
```diff
@@ -367,3 +367,18 @@ func stringSliceToSet(in []string) *schema.Set {
 
 	return schema.NewSet(schema.HashString, result)
 }
+
+// update the computeClient microversion properly by checking if it already was set.
```
kayrus:
```suggestion
// Update the client microversion properly by checking if it already was set.
```

kayrus (approved):

kayrus:
Let’s leave this PR as is for now. We'll address any other cases as they come up.

PR #2013, opened 2026-01-19
tests: fix octavia acceptance tests

kayrus:
This is an attempt to fix octavia functional tests


kayrus:
well, at least other branches are green now. will cover barbican issue in a new PR

Issue #1188, opened 2021-02-27
Remove deprecated GetOkExists


ozerovandrei:
Usage of `GetOkExists` is discouraged in SDK V2 due to undefined behaviors and may beremoved in a future version of the SDK.

https://github.com/hashicorp/terraform-plugin-sdk/blob/v2.4.4/helper/schema/resource_data.go#L106

SDK V2 will be merged in https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1139

kayrus:
An alternative is to use `GetRawConfig`

PR #2051, opened 2026-07-07
Fix: use getOkExists for zero-value fields in openstack_compute_flavor_v2 data source

### LINKED ISSUE #2050: Bug: `disk = 0` (and `ram = 0`, `vcpus = 0`, `swap = 0`) silently ignored in `openstack_compute_flavor_v2` data source
> **Disclaimer:** This bug report was created with the assistance of AI tools. The root cause analysis was performed by an AI assistant reviewing the provider source code; the findings were then confirmed by reproducing the issue on a real OpenStack installation. I have reviewed the proposed changes and verified the fix locally.

---

# Bug: `disk = 0` (and `ram = 0`, `vcpus = 0`, `swap = 0`) silently ignored in `openstack_compute_flavor_v2` data source

## Summary

The `openstack_compute_flavor_v2` data source uses `d.GetOk()` to check whether `disk`, `ram`, `vcpus`, and `swap` were set in the configuration. Because `GetOk()` treats the zero-value (`0`) as "not set", specifying `disk = 0` (a valid and common value meaning "no ephemeral disk") causes the filter to be **silently skipped**. This leads to unexpected "more than one result" errors when multiple flavors match the remaining criteria.

## Terraform Version

```
Terraform (or OpenTofu) 1.x
terraform-provider-openstack v3.4.0
```

## Affected Resource(s)

- `data "openstack_compute_flavor_v2"`

## Terraform Configuration Files

```hcl
data "openstack_compute_flavor_v2" "small" {
  vcpus = 2
  ram   = 8192
  disk  = 0
}

output "flavor_id"   { value = data.openstack_compute_flavor_v2.small.id }
output "flavor_name"  { value = data.openstack_compute_flavor_v2.small.name }
```

## Expected Behavior

The data source should filter for flavors with **exactly** 0 GB disk, 2 vCPUs, and 8192 MB RAM. If only one flavor matches, it should be returned.

## Actual Behavior

The `disk = 0` filter is silently ignored because `d.GetOk("disk")` returns `(0, false)` for the zero-value. The query effectively becomes "2 vCPUs + 8192 MB RAM, any disk", which matches multiple flavors and produces:

```
Error: Your query returned more than one result. Please try a more specific search criteria
```

## Root Cause

In `openstack/data_source_openstack_compute_flavor_v2.go`, the filtering loop uses `d.GetOk()`:

```go
// d.GetOk is used because 0 might be a valid choice.
if v, ok := d.GetOk("disk"); ok {
    if flavor.Disk != v.(int) {
        continue
    }
}
```

The comment claims "0 might be a valid choice", but `d.GetOk()` returns `false` for zero-values, so the filter is skipped. The same bug affects `ram`, `vcpus`, and `swap`.

## Suggested Fix

Use `d.GetOkExists()` instead of `d.GetOk()` for fields where `0` is a valid value. The codebase already uses a `getOkExists()` helper for `is_public` — the same pattern should be applied to `disk`, `ram`, `vcpus`, and `swap`.

## Workaround

Use `flavor_id` or `name` to uniquely identify the flavor, bypassing the spec-based filter.

## References

- Related issue (different problem, same data source): #1081
- `d.GetOk` vs `d.GetOkExists` documentation: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.GetOkExists

benedikt-haug:
## Problem

`d.GetOk()` returns `false` for zero-values (0, 0.0), causing `disk = 0`, `ram = 0`, `vcpus = 0`, and `swap = 0` filters to be silently skipped in the `openstack_compute_flavor_v2` data source. The existing comment says "0 might be a valid choice" — but `GetOk` treats 0 as "unset".

## Fix

Replace `d.GetOk()` with `d.GetOkExists()` for `ram`, `vcpus`, `disk`, `swap`, and `rx_tx_factor`. `GetOkExists` distinguishes between "field not set" and "field explicitly set to zero".

## Verification

Built the provider locally and tested against a real OpenStack installation. Verified that `disk = 0` is now correctly honored as a filter — only flavors with `disk = 0` match, whereas before the filter was ignored and any flavor matching the other criteria was returned.

> **Disclaimer:** This bug report and fix were created with the assistance of AI tools. The root cause analysis was performed by an AI assistant reviewing the provider source code. The fix was reviewed and verified by me on a real OpenStack installation.

Fixes #2050

### DIFF
--- a/openstack/data_source_openstack_compute_flavor_v2.go
+++ b/openstack/data_source_openstack_compute_flavor_v2.go
@@ -191,32 +191,32 @@ func dataSourceComputeFlavorV2Read(ctx context.Context, d *schema.ResourceData, 
 				}
 			}
 
-			// d.GetOk is used because 0 might be a valid choice.
-			if v, ok := d.GetOk("ram"); ok {
+			// d.GetOkExists is used because 0 might be a valid choice.
+			if v, ok := d.GetOkExists("ram"); ok {
 				if flavor.RAM != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOk("vcpus"); ok {
+			if v, ok := d.GetOkExists("vcpus"); ok {
 				if flavor.VCPUs != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOk("disk"); ok {
+			if v, ok := d.GetOkExists("disk"); ok {
 				if flavor.Disk != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOk("swap"); ok {
+			if v, ok := d.GetOkExists("swap"); ok {
 				if flavor.Swap != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOk("rx_tx_factor"); ok {
+			if v, ok := d.GetOkExists("rx_tx_factor"); ok {
 				if flavor.RxTxFactor != v.(float64) {
 					continue
 				}
--- a/openstack/data_source_openstack_compute_flavor_v2.go
+++ b/openstack/data_source_openstack_compute_flavor_v2.go
@@ -191,32 +191,32 @@ func dataSourceComputeFlavorV2Read(ctx context.Context, d *schema.ResourceData, 
 				}
 			}
 
-			// d.GetOkExists is used because 0 might be a valid choice.
-			if v, ok := d.GetOkExists("ram"); ok {
+			// d.GetOk is used because 0 might be a valid choice.
+			if v, ok := getOkExists(d, "ram"); ok {
 				if flavor.RAM != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOkExists("vcpus"); ok {
+			if v, ok := getOkExists(d, "vcpus"); ok {
 				if flavor.VCPUs != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOkExists("disk"); ok {
+			if v, ok := getOkExists(d, "disk"); ok {
 				if flavor.Disk != v.(int) {
 					continue
 				}
 			}
 
-			if v, ok := d.GetOkExists("swap"); ok {
+			if v, ok := getOkExists(d, "swap"); ok {
 				if flavor.Swap != v.(int) {
 					continue
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


openstack/data_source_openstack_compute_flavor_v2.go
```diff
@@ -191,32 +191,32 @@ func dataSourceComputeFlavorV2Read(ctx context.Context, d *schema.ResourceData,
 				}
 			}
 
-			// d.GetOk is used because 0 might be a valid choice.
-			if v, ok := d.GetOk("ram"); ok {
+			// d.GetOkExists is used because 0 might be a valid choice.
+			if v, ok := d.GetOkExists("ram"); ok {
```
kayrus: `GetOkExists` is deprecated, see #1188
We use `getOkExists(d, key)`, see #1759

benedikt-haug: Thanks @kayrus! Fixed in 419ba80b — switched to `getOkExists(d, key)` helper from util.go, consistent with the existing `is_public` usage in the same file.

kayrus: could you add a functional test that covers this case? e.g. create a dummy image without a payload that has 0 for corresponding values?

benedikt-haug: sadly don't have a dev openstack setup to run the tests against, so couldn't run the tests myself. Is this draft close to the testcase you thought off?

openstack/data_source_openstack_compute_flavor_v2_test.go
```diff
@@ -247,3 +247,49 @@ func testAccComputeV2FlavorDataSourceFlavorID(flavorName string) string {
...
+resource "openstack_compute_flavor_v2" "flavor_1" {
+  name      = "%s"
+  ram       = 512
+  vcpus     = 1
+  disk      = 0
+  is_public = true
+}
+
+data "openstack_compute_flavor_v2" "flavor_1" {
+  vcpus = 1
```
kayrus: this hack creates a dependency and data source will be executed right after the flavor is created
```suggestion
  vcpus = openstack_compute_flavor_v2.flavor_1.vcpus
```

kayrus (approved):

kayrus:
@benedikt-haug approved. thank you.

PR #2057, opened 2026-07-21
Blockstorage: preserve volume backup ID in state

Sebor:
Currently, backup_id is not stored in the Terraform state, causing every terraform apply to attempt to recreate the volume.

This pull request prevents that from happening and eliminates the need to use "ignore_changes = [backup_id]"


Sebor:
Hi @kayrus @nikParasyr,
Could you review my PR please?

PR #2062, opened 2026-08-04
fix: set ID of DNS zone or recordset immediately to update the state

### LINKED ISSUE #1111: dns recordset/zone created in ERROR state creates invalid terraform state
### Terraform Version
v0.14.0

### Affected Resource(s)
- openstack_dns_recordset_v2
- openstack_dns_zone_v2

### terraform apply output snips
https://gist.github.com/Roukoswarf/b4c97f665e0d8db2486a262ae9121be4

### Expected Behavior
State should be aware that the record was created, and is in pending state, or state should be pulled from openstack before attempting recreation, so that this can fix itself when the resource becomes active on subsequent runs.

### Actual Behavior
First run fails, due to ERROR state check on zone, which in openstack represents a replication delay.
Second run fails, even if the zone is now ACTIVE, as state does not reflect that it was created, and receives 409 conflict when creating an already-existing record.

### Steps to Reproduce
1. `terraform apply` - with openstack setting the zone into ERROR
1. `terraform apply` - with the zone being ACTIVE (propagation delay resolved)

### Important Factoids
Some of our dns backends have replication delay issues, which may cause ERROR states which eventually resolve themselves.

frittentheke:
If we wait for a zone or recordset to become `ACTIVE` and time out, the state is never updated and a dangling record exists in the API. This record then causes 409 Conflict on subsequent terraform runs.

Fixes: #1111


frittentheke:
@kayrus @TBExtent could you PTAL at this one here?

PR #2069, opened 2026-08-17
resource/openstack_identity_user_v3: add write-only password support (password_wo / password_wo_version)

charliemaiors:
## Summary

This PR adds support for Terraform's **write-only attributes** to the `openstack_identity_user_v3` resource, allowing the user's password to be provided via a `password_wo` attribute that is never persisted to state, paired with a `password_wo_version` counter to trigger updates.

## Motivation

The existing `password` attribute is a regular (non write-only) `Sensitive` field, which means its value is stored in the Terraform state file, even though it is marked sensitive. This is a known concern for secret management: anyone with access to the state file can retrieve the plaintext password.

Terraform 1.11+ write-only attributes (`ephemeral`/`WriteOnly` schema fields) solve this by never writing the value to state. Since write-only attributes have no persisted value to diff against, a companion "version" attribute (`password_wo_version`) is used to signal that the value has changed and an update should be triggered — this is the same pattern already adopted by other providers/resources for write-only secrets.

## Changes

- **`openstack/resource_openstack_identity_user_v3.go`**
  - Added two new optional schema attributes:
    - `password_wo` (`TypeString`, `Sensitive`, `WriteOnly`) — the write-only password value.
    - `password_wo_version` (`TypeInt`) — a version counter used to detect password changes since `password_wo` itself is never stored in state.
  - **Create**:
    - Returns an error if both `password_wo` and `password` are set at the same time, to avoid ambiguous behavior.
    - If `password_wo` is set, `password_wo_version` is now required; the user is created using `password_wo` instead of `password`.
  - **Update**:
    - Returns an error if both `password_wo_version` and `password` have changed in the same apply, to avoid ambiguous behavior.
    - If `password_wo_version` has changed, `password_wo` is required and its value is used to update the user's password.
    - The existing `password` change-detection logic is left untouched for backward compatibility.

- **`openstack/resource_openstack_identity_user_v3_test.go`**
  - Added an additional acceptance test step (`testAccIdentityV3UserWriteOnly`) that creates a user using `password_wo` / `password_wo_version = 1` instead of `password`, verifying the resource is created correctly and that `password_wo_version` is correctly tracked in state.

## Backward compatibility

- Fully backward compatible: `password` continues to work exactly as before for users who don't opt into the write-only pattern.
- `password_wo` / `password_wo_version` are both optional and only take effect when explicitly set.

## Notes / Open points for reviewers

- Documentation (`docs/resources/identity_user_v3.md`) has not been updated yet in this PR — happy to add the `password_wo` / `password_wo_version` attribute descriptions and a usage example if the approach is accepted.
- Requires Terraform >= 1.11 (or OpenTofu with write-only attribute support) to actually benefit from the `WriteOnly` behavior; on older versions the attribute will behave as a normal optional field but won't be persisted differently — worth confirming the provider's minimum supported Terraform version is compatible.
- Open to feedback on whether the "must set both attributes together" validation should instead be handled via `ConflictsWith` / `RequiredWith` schema constraints rather than manual checks in the CRUD functions, for more idiomatic SDKv2 validation.

## Testing

- Ran `make lint` (golangci-lint, including `wsl`/`whitespace` checks) locally.
- Added acceptance test step `testAccIdentityV3UserWriteOnly` (requires `TF_ACC=1` and a live OpenStack environment to execute).

### DIFF
--- a/openstack/resource_openstack_identity_application_credential_v3.go
+++ b/openstack/resource_openstack_identity_application_credential_v3.go
@@ -62,12 +62,14 @@ func resourceIdentityApplicationCredentialV3() *schema.Resource {
 				Type:      schema.TypeString,
 				Optional:  true,
 				WriteOnly: true,
+				ConflictsWith: []string{"secret"},
 			},
 
 			"secret_wo_version": {
 				Type:     schema.TypeInt,
 				Optional: true,
 				ForceNew: true,
+				ConflictsWith: []string{"secret"},
 			},
 
 			"project_id": {
--- a/openstack/resource_openstack_identity_application_credential_v3.go
+++ b/openstack/resource_openstack_identity_application_credential_v3.go
@@ -59,16 +59,16 @@ func resourceIdentityApplicationCredentialV3() *schema.Resource {
 			},
 
 			"secret_wo": {
-				Type:      schema.TypeString,
-				Optional:  true,
-				WriteOnly: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				WriteOnly:     true,
 				ConflictsWith: []string{"secret"},
 			},
 
 			"secret_wo_version": {
-				Type:     schema.TypeInt,
-				Optional: true,
-				ForceNew: true,
+				Type:          schema.TypeInt,
+				Optional:      true,
+				ForceNew:      true,
 				ConflictsWith: []string{"secret"},
 			},
 
--- a/openstack/resource_openstack_identity_application_credential_v3.go
+++ b/openstack/resource_openstack_identity_application_credential_v3.go
@@ -58,6 +58,18 @@ func resourceIdentityApplicationCredentialV3() *schema.Resource {
 				ForceNew:  true,
 			},
 
+			"secret_wo": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				WriteOnly: true,
+			},
+
+			"secret_wo_version": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+
 			"project_id": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -146,7 +158,13 @@ func resourceIdentityApplicationCredentialV3Create(ctx context.Context, d *schem
 
 	log.Printf("[DEBUG] openstack_identity_application_credential_v3 create options: %#v", createOpts)
 
-	createOpts.Secret = d.Get("secret").(string)
+	if d.Get("secret_wo") != nil && d.Get("secret") != nil {
+		return diag.Errorf("Either secret or secret_wo could be specified, specifying both could lead to an inconsistent view")
+	} else if d.Get("secret_wo") != nil {
+		createOpts.Secret = d.Get("secret_wo").(string)
+	} else {
+		createOpts.Secret = d.Get("secret").(string)
+	}
 
 	applicationCredential, err := applicationcredentials.Create(ctx, identityClient, tokenInfo.userID, createOpts).Extract()
 	if err != nil {
--- a/openstack/resource_openstack_identity_application_credential_v3.go
+++ b/openstack/resource_openstack_identity_application_credential_v3.go
@@ -160,14 +160,6 @@ func resourceIdentityApplicationCredentialV3Create(ctx context.Context, d *schem
 
 	log.Printf("[DEBUG] openstack_identity_application_credential_v3 create options: %#v", createOpts)
 
-	if d.Get("secret_wo") != nil && d.Get("secret") != nil {
-		return diag.Errorf("Either secret or secret_wo could be specified, specifying both could lead to an inconsistent view")
-	} else if d.Get("secret_wo") != nil {
-		createOpts.Secret = d.Get("secret_wo").(string)
-	} else {
-		createOpts.Secret = d.Get("secret").(string)
-	}
-
 	applicationCredential, err := applicationcredentials.Create(ctx, identityClient, tokenInfo.userID, createOpts).Extract()
 	if err != nil {
 		if gophercloud.ResponseCodeIs(err, http.StatusNotFound) {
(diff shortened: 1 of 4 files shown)


thundertaker:
Could you please also add wo support for "secret" field at openstack_identity_application_credential_v3 resource?

charliemaiors:
Yep, implementing it right now.

charliemaiors:
Done

charliemaiors:
Submitted a fix for the tests, could you please approve?

charliemaiors:
Another fix, could you please run the tests?

charliemaiors:
I didn't quite understand the latest errors, I've made a "fix" could you try to run again?

charliemaiors:
Ok understood the errors, could you please run it again?

charliemaiors:
Guys any update?

kayrus:
@charliemaiors thanks for update. I haven't seen docs update with examples on how to use new `password_wo` along with `password_wo_version`

docs/resources/identity_application_credential_v3.md
```diff
@@ -36,6 +36,27 @@ resource "openstack_identity_application_credential_v3" "swift" {
...
+ephemeral "random_password" "secret" {
+  length           = 16
+  special          = false
+}
+
+resource "openstack_identity_application_credential_v3" "swift" {
+  name        = "swift"
+  description = "Swift technical application credential"
+  secret_wo      = "${ephemeral.random_password.secret.result}$"
+  secret_wo_version = 1
```
kayrus, line 55: Please add new arguments below the `Argument Reference` in docs

kayrus (commented):
Does it make sense to specify `Sensitive: true` along with the `WriteOnly: true`?

charliemaiors:
IMHO make sense to have `Sensitive: true` with `WriteOnly: true`; in both cases, it will end up in the Terraform state. If you put it in the outputs with the sensitive value, you have to configure the backend in order to visualize it.
I will fix the missing part in the documentation and add Sensitive to the arguments.

openstack/resource_openstack_identity_application_credential_v3.go
kayrus: add a condition here, otherwise `secret` end up in state anyway

docs/resources/identity_application_credential_v3.md
```diff
@@ -36,6 +36,27 @@ resource "openstack_identity_application_credential_v3" "swift" {
...
 ```
 
+
+### Predefined secret with secret_wo
+
+Application credential below will have only one `swiftoperator` role using `secret_wo`
+
+```hcl
+
+ephemeral "random_password" "secret" {
```
kayrus, line 46: How would I know the password if it's reflected in neither the state nor the plan?

charliemaiors: You need to store it in a secret manager and then is up to the user to retrieve it.

openstack/resource_openstack_identity_application_credential_v3.go
charliemaiors: Secret will be written in the terraform state, secret_wo no.

kayrus: yes, that is why you need to add a condition to prevent `secret_wo` to end up in the `secret` state

charliemaiors: Done

openstack/resource_openstack_identity_application_credential_v3_test.go
```diff
@@ -57,7 +57,26 @@ func TestAccIdentityV3ApplicationCredential_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						"openstack_identity_application_credential_v3.app_cred_1", "unrestricted", "true"),
 					resource.TestCheckResourceAttr(
-						"openstack_identity_application_credential_v3.app_cred_1", "secret", "foo"),
+						"openstack_identity_application_credential_v3.app_cred_1", "secret_wo_version", "1"),
```
kayrus, line 63: please add a check that the `secret` is empty

openstack/resource_openstack_identity_user_v3_test.go
```diff
@@ -59,6 +61,20 @@ func TestAccIdentityV3User_basic(t *testing.T) {
...
+			{
+				Config: testAccIdentityV3UserWriteOnly(writeOnlyProjectName, writeOnlyUsername),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckIdentityV3UserExists(t.Context(), "openstack_identity_user_v3.user_1", &user),
+					testAccCheckIdentityV3ProjectExists(t.Context(), "openstack_identity_project_v3.project_1", &project),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_user_v3.user_1", "name", &user.Name),
+					resource.TestCheckResourceAttrPtr(
+						"openstack_identity_user_v3.user_1", "description", &user.Description),
+					resource.TestCheckNoResourceAttr("openstack_identity_user_v3.user_1", "password_wo"),
```
kayrus: please add a check that the `password` is empty

kayrus (commented):
looks better, I guess after the final notes we can merge

openstack/resource_openstack_identity_user_v3_test.go
charliemaiors: Adding

charliemaiors: Done

openstack/resource_openstack_identity_application_credential_v3_test.go
charliemaiors: Done

charliemaiors:
Tests failed because of the Terraform setup. What could I do?

kayrus:
> resource_openstack_identity_user_v3_test.go:27: Step 2/3 error: Check failed: Check 1/7 error: Not found: openstack_identity_user_v3.user_1

please fix this test

charliemaiors:
Problem found, hopefully fixed

kayrus:
BTW, you can to run functional tests directly in your local env or any other openstack cloud. see https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/main/docs/index.md#testing-and-development

charliemaiors:
I don't have access to a local env or any other Openstack cluster unfortunately, I can change the workflows to run it on my account and run the functional tests on a commit on the main branch.

charliemaiors:
Fixed also a missing hasChanged in resource update

PR #2070, opened 2026-08-21
build(deps): bump gophercloud to v2.14.0

TBExtent:
Gophercloud v2.14.0 adds support for get and update operations on Manila share types (https://github.com/gophercloud/gophercloud/pull/3902). This is required to add support for managing Manila share types to the Terraform provider.


kayrus:
fixed in #2071

